PR #38, opened 2015-11-01
Improved project structure, replaced node-unit with tape

tetsuo:
Not sure if these can be merged easily :) But here's my take for some issues I've encountered.

## export individual files

Plain js dependents had no access to individual files (other than the webpack bundle), which is kind of a issue if you need to use `instanceof` and sort.

I needed this to be exact:

```js
var ObservableObject = require('mobservable/lib/observableobject').ObservableObject
var m = require('mobservable')

var a = m.observable({})

a.$mobservable instanceof ObservableObject
```

Exporting transpiled files instead of the bundle also allows node (or browserify) to hand over module resolution task.

## browserify

Building dist separately using browserify (tsify) and extracting sourcemaps- both for minified and unminified versions.

## tape migration

tape is being used in this fork, tape-run to run them on a headless electron. (`make test-browser`) Babel test is being browserified upfront. Typescript test is compiled into a tmp dir. (But you can also run it directly with `node test/typescript/x.js`)

## some cleanup

Removed `publish.js`, `.vscode` and `test/debugger`

Maybe `coveralls` should be included in `.travis.yml` as a script.
...

So far tests are running fine. (Except for one in browser tests, `transactions 4`)

So, what do you think? :)


mweststrate:
Hi Onur,

Just glanced at it briefly, but this seems to be great stuff! The tape
migration was actually on my wish list for a while :) Not sure yet about
browserify and webpack because so far I always had issues with browserify,
but that might be just me ;-). Anyway thanks a lot! Probably I will not
deep dive into it or merge for the next few days because I just arrived at
Reactive2015 and will be away until early next week. (I will give a talk on
mobservable on Wednesday, so for important suggestions just fire away ;-)).

Cya!

Michel

tetsuo:
@mweststrate Hey!

Ah, great, good luck at Reactive2015 :)

This is more or less a quick prototype indeed. Just wanted to hear your opinion.

I've started playing with mobservable on plain js, and quickly found myself doing lots of type comparisons. Often times these are very low-level and not that future-proof:

```js
function isObservableArray (s) {
  return m.isObservable(s) && (s.constructor.name === 'ObservableArray')
}

function isObservableObject (s) {
  return m.isObservable(s) && (s.$mobservable.constructor.name === 'ObservableObject')
}
...
``` 

Now I'm dabbling with TypeScript since yesterday, I've realized using mobservable in TS is a whole different story, the real deal- so to speak.

Well, great to hear you are open to these :) I wasn't so sure if the tape migration and sort worth the effort. 

Have super fun!

mweststrate:
Hi Onur,

I merged a lot of your changes into https://github.com/mweststrate/mobservable/commits/pull/38/tape-misc. I used all your tape migrations and `src/` and `lib/` folder improvements. `dist` provide now the one file umd bundle (including minificating & source maps). `lib/` provides the transpiled files as suggested.

I stuck to webpack instead of browserify. I didn't look into running the browser tests as I had some trouble getting it running on my machine. Maybe i'll look later into that. 

If you have any further suggestions, just let me know. Otherwise I'll merge it into master in the next few days.

Thanks a lot for all the additions and suggestions!

tetsuo:
@mweststrate great work  :+1: I'm looking into browser tests right now.

tetsuo:
`npm run test-browser` is working fine if I'm not missing something, everything passes :100:

mweststrate:
@tetsuo jup fixed it yesterday evening. Probably was a local hiccup on my npm cache or something. The changes of this PR are now part of release 1.0.5. So thanks a lot!

The previously failing test related to the use of `console.table`, which has been fixed.

tetsuo:
Nice :) Looks like babel tests are remained as `console.log`s (sending a pr in a moment) I'm seeing you've decided not to include on-the-fly compilation for this one and typescript tests.

mweststrate:
Yeah I didn't see directly the added value of on-the-fly compilation so figured the additional complexity wasn't really needed?

tetsuo:
Yep indeed, adds unnecessary complexity. I was testing browserify build with `vm`, then it became babel test (◕‿◕✿)

PR #42, opened 2015-11-11
remove dup array perf tests

tetsuo:
Not sure if this was intentional or not?


mweststrate:
Hej Onur,

This was originally intentional but no longer needed. I saw that there is still an error in the invocation, so i'll fix this one on a local branch.

PR #52, opened 2015-11-13
Minor tweak to the index imports

dzearing:
This will help systemjs bundler, which isn't totally intelligent about commonjs import implicit resolution, find the index.js.


mweststrate:
Thanks! Released it as 1.0.6

PR #66, opened 2015-12-06
Fixed transaction inside transaction through a defensive copy. Fixes #65

### LINKED ISSUE #65: Transactions: cannot read property 'markReady' of undefined
I think the logic [here](https://github.com/mweststrate/mobservable/blob/master/src/dnode.ts#L49) is to catch some updating changes mid-update.

However, It looks like `markReady` can alter the changedValues array values which is causing the 'cannot read property 'markReady' of undefined' error.

Should the solution be:

```
           var length_1 = changedValues.length;
            for (var i = 0; i < length_1; i++)
                !changedValues[i] || changedValues[i].markReady(true);
            changedValues.splice(0, length_1);
            if (changedValues.length)
                throw new Error("[mobservable] Illegal State, please file a bug report");
```

or 

```
            var values = changedValues.splice(0);
            for (var i = 0, len = values.length; i < len; i++)
                values[i].markReady(true);
            if (changedValues.length)
                throw new Error("[mobservable] Illegal State, please file a bug report");
```

or do you need me to try to figure out why the issue is happening in the first place? (I've got a transaction that does quite a variety of actions so perhaps values could be removed mid update, but there are done being added so it is not the `Illegal State`).

kmalakoff:
See https://github.com/mweststrate/mobservable/issues/65


mweststrate:
Added tests to verify the bug and solution, and published as 1.1.4. 

Thanks!

kmalakoff:
Thank you! 

(sorry, I should have submitted a test too...next time!)

PR #70, opened 2015-12-10
Feature/transform

kmalakoff:
Added tests: https://github.com/mweststrate/mobservable/issues/67#issuecomment-163342323

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,54 +1,55 @@
 {
-    "name": "mobservable",
-    "version": "1.1.3",
-    "description": "Observable data. Reactive functions. Simple code.",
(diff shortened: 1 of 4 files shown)


mweststrate:
Merged, thanks a lot!

kmalakoff:
@mweststrate my pleasure. 

I'm not sure if the version in the tests has asFlat, asReference, etc correct, though. I started to change it, but the tests broke and I'm going to start moving my code over to using transformations before I look into this and also need to get back to feature development. 

If you want to update the tests to use the right method until I get around to it (it could be 4+ days), be my guest...

PR #80, opened 2015-12-27
Update observer-component.md

jeblister:
add a missing observable function :
```
var timerData = observable({
    secondsPassed: 0
});
```


mweststrate:
Woops! Thanks for noticing!

PR #81, opened 2015-12-27
Update modifiers.md according to description

jeblister:
fix a doubler function methode.

Add console log for client side code.


mweststrate:
Tnx!

PR #91, opened 2016-01-19
Feature/2 cycles

kmalakoff:
Regarding https://github.com/mweststrate/mobservable/issues/86, I've added a couple of tests from my experiences with cycles and the doc outline.

Everything passed on V2 and I've confirmed two of the tests fail on V1...awesome!


mweststrate:
Tnx!

PR #94, opened 2016-01-23
Fix typo - introducton -> introduction

bebraw:


nightwolfz:
Maybe change tojson to toJSON at the same time?

mweststrate:
@bebraw tnx!
@nightwolfz which `tojson`?

PR #107, opened 2016-01-31
Support frozen objects

andykog:
This code:
```js
var arr = mobservable.observable([]);
var person = Object.freeze({ name: 'bob' })
arr.push(person)
```
throws error `Cannot define property:$mobservable, object is not extensible.`

This also prevents from pushing react elements into mobservable array:
```
buttons.push(<button title="test" />);
```


mweststrate:
Thanks for the addition!

mweststrate:
Released as part of version 1.2.4

PR #111, opened 2016-02-09
skip properties which are not writable or not configurable

### LINKED ISSUE #110: ES5 property descriptor support is somewhat lacking
mobservable tries to redefine a property marked as configurable: false into setter/getter
```javascript
var te = {}
Object.defineProperty(te, 'key', {
  enumerable: true,
  configurable: false,
  writable: false,
  value: 'static'
})

const a =  mobservable.observable(te)
// last line throws
/home/capaj/git_projects/mobservable-db/node_modules/mobservable/lib/observableobject.js:53
        Object.defineProperty(this.target, propName, {
               ^

TypeError: Cannot redefine property: key
    at Function.defineProperty (native)
    at ObservableObject.defineReactiveProperty (/home/capaj/git_projects/mobservable-db/node_modules/mobservable/lib/observableobject.js:53:16)
    at ObservableObject.set (/home/capaj/git_projects/mobservable-db/node_modules/mobservable/lib/observableobject.js:38:18)
    at extendObservableHelper (/home/capaj/git_projects/mobservable-db/node_modules/mobservable/lib/core.js:348:18)
    at makeChildObservable (/home/capaj/git_projects/mobservable-db/node_modules/mobservable/lib/core.js:440:16)
    at Object.observable (/home/capaj/git_projects/mobservable-db/node_modules/mobservable/lib/core.js:43:20)
    at Object.<anonymous> (/home/capaj/git_projects/mobservable-db/aa.js:11:24)
    at Module._compile (module.js:398:26)
    at Object.Module._extensions..js (module.js:405:10)
    at Module.load (module.js:344:32)

```

I think when a property is not configurable, mobservable should just skip it.

Also when property is `configurable: true` but marked as `writable: false`, mobservable should also skip it and do not try to make it into a getter/setter.

This should be easily done with like 2 if statements(+ 2 simple tests)-should I bother with a PR, or can you do it yourself @mweststrate ?

capaj:
it doesn't make sense to observe those properties,
sorry about all the whitespace changes, tried to get rid of them but my atom always makes them on save.
fixes #110


capaj:
just realised-we still need a mobservable getter on a non-writable property. Skipping is only allright for `configurable: false`. I will try to get a proper fix fleshed out in the evening.

mweststrate:
Much appreciated! Maybe it could even throw if the decorator is added to a
non-configurable property?

capaj:
@mweststrate haven't thought about that one. I mostly use it on objects. But yeah, in case you use it on single property, it makes total sense to fail fast and throw.

mweststrate:
Finished this on the `feature/2` branch: when using `extendObservable` or `@observable` on a non-configurable / non-writable property it will throw. But when using `observable(object)`, it will skip non-configurable / non-writable properties.

mweststrate:
Apologies btw, somehow your commits are not visible in the history, I guess I merged in an incorrect manner.

capaj:
@mweststrate awesome! so this will get released with mobservable 2.0.0 right?

mweststrate:
Indeed :)

Op zo 21 feb. 2016 om 23:43 schreef Jiri Spac <notifications@github.com>:

PR #113, opened 2016-02-14
Preupdate events

capaj:
I am experimenting with mobservable and I would love to be able to do validation inside my `observe` method callback. Currently it only emits event after the value has been set/updated. This was according to Object.observe strawman proposal so I can understand why it has been done like that. Problem is that proposal is no longer relevant and so I think it would be okay to go on and implement more than what it proposed.
Currently I have a code like this:

```javascript
toBeObserved._disposer = mobservable.observe(toBeObserved, (change) => {
      const id = toBeObserved.id
      debug(`${name} ${id} change `, change)
      if (change.type === 'preupdate') {
        if (change.newValue !== change.oldValue) {
          const coerced = joi.attempt(change.newValue, schema[change.name])
          if (coerced !== change.newValue) {
            throw new TypeError(`Expected ${schema[change.name]._type} but assigned ${joi.compile(change.newValue)._type} on property ${change.name} on object ${id}`)
          }
        }
      } else {
        putDoc(toBeObserved)
      }
    })
```

Which needs notifications before the value is set. Hence why I added `preupdate`. This PR would make it possible to do this with mobservable.

At this moment I am not sure if other dependent packages would need any change, but I presume they would, because you probably always assume to get an event only after a change has occurred. @mweststrate let me know if this would be too big to do. I can support my own branch if this change is not feasible, but sooner I know the better.


mweststrate:
Hi @capaj 

Tnx for the contribution! I'm gonna think this through after wrapping up 2.0 :).

capaj:
@mweststrate thanks, I think it would be a nice addition, but it probably requires a bit more work-for example to have this preupdate for arrays and maps as well.

mweststrate:
@capaj Might this feature request obsolete this PR? https://github.com/mobxjs/mobx/issues/142

capaj:
@mweststrate yes, but I wonder what would the custom setters/getters API look like. I imagine implementing custom setters/getters would be much bigger in scope, than just these `preupdate` events. I would be satisfied with any of those two.

capaj:
closed in favour of #182

PR #139, opened 2016-02-29
Fix 404 in link to API documentation from readme

bwhitty:
https://mobxjs.github.io/ 404s and does not actually link to the API documentation, so here's a fix!


mweststrate:
Wow, stupid mistake. thanks for spotting!

PR #171, opened 2016-03-24
Add a a section on stateless components and HMR in `mobx-react`

cafreeman:
Seems like Hot Module Reloading is beginning to be a bit of a sticking point when people are first learning `mobx-react`, especially when it comes to stateless components, so I thought a section in the best practices covering this would help new users.


cafreeman:
@mweststrate not sure if I need to tag you specifically in this or if it filters through the `MobX` organization automatically, so...tagging you just in case :)

mweststrate:
Looks great :) Thanks a lot!

PR #182, opened 2016-04-11
preupdate event added

capaj:
for earlier discussion, see https://github.com/mobxjs/mobx/pull/113

### DIFF
--- a/src/types/observableobject.ts
+++ b/src/types/observableobject.ts
@@ -115,7 +115,7 @@ function defineObservableProperty(adm: IObservableObjectAdministration, propName
 }
 
 /**
-	* Observes this object. Triggers for the events 'add', 'update' and 'delete'.
+	* Observes this object. Triggers for the events 'add', 'update', 'preupdate' and 'delete'.
 	* See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/observe
 	* for callback details
 	*/
--- a/src/types/observableobject.ts
+++ b/src/types/observableobject.ts
@@ -82,9 +82,18 @@ function defineObservableProperty(adm: IObservableObjectAdministration, propName
 			return observable.get();
 		},
 		set: isComputed
-			? throwingComputedValueSetter 
+			? throwingComputedValueSetter
 			: function(newValue) {
 				const oldValue = (observable as any).value;
+				if (adm.events) {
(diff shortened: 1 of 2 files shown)


capaj:
the project I need this for is https://github.com/capaj/mobxdb, I use preupdate events here exactly:
https://github.com/capaj/mobxdb/blob/master/lib/model.js#L84

mweststrate:
Hey thanks! Gonna look into this after next week, I left the other PR lingering around for way too long. Did you notice that these events won't respect transactions? But that is probably what you need anyway.

capaj:
@mweststrate yes, these events wouldn't respect transactions. It would be better if they did. I will have to look into how transactions are managed. No rush, in mobxdb I install mobx from my fork, which is ok for now.

mweststrate:
@capaj 

Thought about this for a while, and it might be more generally solvable: What do you think of this: we introduce a new concept, `intercept`, which is kind of the inverse of `observe` and works a bit like proxies. When you add an intercept handler to an object it will receive the splice event and that handler should return the same or a new splice, or no splice at all if the change is to be rejected.

That mechanism could then probably also used for general 'middleware' like logging actions as done in #193, creating json patch streams etc.

capaj:
@mweststrate I am all for approaching this in more systemic way. I probably could have just opened an issue around this, but I like that PR clearly demonstrates what I need.
Intercept sounds good. What exactly do you mean by `splice`? Like an array? That would work.
I guess it has to be an array because of transactions. Otherwise you would only receive/handle one event at a time.
Feel free to implement it. I have my hands full for the next two weeks.If you don't, I'll try to make another PR.

mweststrate:
@capaj could you verify that this is on the right track:

`intercept(object, handler: change => change): disposer` should work for object similarly (will add tests for that as well). The change object that you can intercept in the handler has the following signature:

```typescript
export interface IObjectWillChange {
	object: any;
	type: "update" | "add";
	name: string;
	newValue: any;
}
```

In the handler you can:
1. return nothing to cancel the change
2. throw some error
3. return a change object to proceed. Optionally you can change the `newValue` to some other value. 

Multiple handler can be changed, they will be executed in the order they are attached.

The following test demonstrates the above possibilities: 
https://github.com/mobxjs/mobx/commit/e813700b445ec03f52e6078af8263494ba89d136#diff-6f1fd8cea5f832c81d52ecd7b5c878fdR5

PR #192, opened 2016-04-14
Fix @observable decorator (fixes #191)

### LINKED ISSUE #191: Shared elements when using Store classes with @observable
I seem to be encountering some weird behavior when working with observables in classes (or there's a crucial detail about property initializers that I'm failing to grasp).

Specifically, certain properties are being shared across all instances of a class (while others remain localized), so that every time a new instance is created, all the other instances update to reflect the new value.

For example:

```js
class Field {
  @observable fieldName;
  @observable bound = {
    field: null,
    lowerBound: 0,
    upperBound: null,
  }

  constructor(fieldName) {
    this.fieldName = fieldName;
    // All instances of `Field` will have the same bound.field for some reason
    this.bound.field = fieldName;
  }
}
```

In this case, there is a component that uses `fieldName` in the header, and `bound` in the body of a panel. Every time you add a new `Field` to the store, the new component renders with the correct `fieldName` in the header, but not in the body. For some reason, `this.bound.field` is shared across all instances, but `this.fieldName` is not (notice how the header reflects individual field names, but all 3 components now say "field 3" in the body, since field 3 was the most recent value.)

image

On the other hand, this works:
```js
class Field {
  @observable fieldName;

  constructor(fieldName) {
    this.fieldName = fieldName;
    // In this case, each instance will have its own unique `bound` property
    extendObservable(this, {
      bound: {
        field: fieldName,
        lowerBound: 0,
        upperBound: null,
      },
    });
}
```

Using `extendObservable` to add the entire property to the instance in the constructor does successfully localize the entire `bound` property, so that adding a new `Field` to the store correctly adds a fully encapsulated instance.

image

An especially weird side note is that even though `this.bound.field` is shared across all instances, the other two properties of `this.bound` are not. This means that I can change the other fields of `bound` and those changes are reflected solely in the component they belong to, even though the `field` property is still being shared across all three.

image

For reference, I'm using Babel 6 with the following plugins:

  - "babel-plugin-array-includes": "^2.0.3",
  - "babel-plugin-transform-decorators-legacy": "^1.3.4",
  - "babel-plugin-transform-object-assign": "^6.5.0",
  - "babel-preset-es2015": "^6.6.0",
  - "babel-preset-react": "^6.5.0",
  - "babel-preset-react-hmre": "^1.1.1",
  - "babel-preset-stage-1": "^6.5.0",

In addition, I'm using these packages:

  - "mobx": "^2.0.4",
  - "mobx-react": "^3.0.3",
  - "react": "^0.14.8",
  - "react-dom": "^0.14.8"

andykog:
The issue is described in fiddle: https://jsfiddle.net/Sl1v3r/ogkctduh/

### DIFF
--- a/src/api/observabledecorator.ts
+++ b/src/api/observabledecorator.ts
@@ -30,21 +30,21 @@ export function observableDecorator(target: Object, key: string, baseDescriptor:
 		return computed.apply(null, arguments);
 	}
 	const descriptor: PropertyDescriptor = {};
-	let baseValue = undefined;
-	if (baseDescriptor) {
-		if (baseDescriptor.hasOwnProperty("value"))
-			baseValue = baseDescriptor.value;
-		else if ((<any>baseDescriptor).initializer) { // For babel
-			baseValue = (<any>baseDescriptor).initializer();
-			if (typeof baseValue === "function")
-				baseValue = asReference(baseValue);
-		}
-	}
 
 	invariant(typeof target === "object", `The @observable decorator can only be used on objects`, key);
 	descriptor.configurable = true;
 	descriptor.enumerable = true;
 	descriptor.get = function() {
+		let baseValue = undefined;
+		if (baseDescriptor) {
+			if (baseDescriptor.hasOwnProperty("value"))
+				baseValue = baseDescriptor.value;
+			else if ((<any>baseDescriptor).initializer) { // For babel
+				baseValue = (<any>baseDescriptor).initializer();
+				if (typeof baseValue === "function")
+					baseValue = asReference(baseValue);
+			}
+		}
 		// the getter might create a reactive property lazily, so this might even happen during a view.
 		allowStateChanges(true, () => {
 			setObservableObjectProperty(asObservableObject(this, undefined, ValueMode.Recursive), key, baseValue);


jeffijoe:
Nice! Mind explaining how the fix works? :)

andykog:
@jeffijoe, decorators only run once when the class is defined, but as we need unique initial value per instance, I moved code, calling `initializer()` in `get()` function, so it is executed on the first call to any instance's value (and immediately being overridden with instance getter).

andykog:
@mweststrate, need your approve :-)

mweststrate:
@andykog thanks! Yes I think this is the solution indeed. would you mind adding tests to the babel and typescript test files (test/babel/babel-tests) and (test/typescript-tests.ts) that verify this behavior? I'm especially curious whether #191 also occurs when using typescript in the current version. 

I think the typescript compiler lifts all field initializations to the constructor function itself and in that case (typescript) this pattern might not be need. 

Those competing decorators implementations are quite a sad story :disappointed:

mweststrate:
Just verified, typescript lifts the initializers indeed. I'll merge and release, thanks a lot!

mweststrate:
Btw, added tests already, see above commit. Verified indeed that #191 fails _before_ 2.1.3 when using babel but succeeds when using typescript. Thanks for finding this one.

mweststrate:
Btw, your change proved that this code could even further be simplified: https://github.com/mobxjs/mobx/commit/de63c5520e6647d13b56905b1522b5c30fd2e3a0, it relied on a transformation that doesn't exist anymore.

PR #200, opened 2016-04-19
corrected spelling for automatically

tonytainguyen:
changed to correct spelling for automatically.


mweststrate:
Actually I intended atomically here (is that really a word), but to avoid confusion I should mention both indeed :)

tonytainguyen:
Oh really? I actually did look it up before I did the pull request. But I have to admit, as I was reading it the first word I said in my head was automatically and then re-read it again to realize it was 'atomically'. Haha.

Issue #160, opened 2016-03-14
TypeError: undefined is not a constructor (evaluating 'newItems.map(adm.makeChildReactive)')


jeffgran:
I can't figure out this error, but it seems like it may be a problem with mobx. But it might not.

The top of the stack trace has `spliceWithArray` and `replace` (from `observablearray.ts`), and then my code, which is calling `collection.push(myobservableobject)` where `collection` is an empty observable array.

One point of confusion is why is my code calling `push` but then the stack trace shows that I called to `replace`?

I will see if I can put together a minimal case to reproduce.

jeffgran:
can't figure out how to reproduce this reliably. However, I found a workaround: use chrome for my automated tests instead of PhantomJS. It seems to be something that only happens under PhantomJS.

Closing since I am unable to provide reproducible case.

mweststrate:
Ok, tnx for sorting out!

bstst:
This happens to me on iOS 7: `"Mozilla/5.0 (iPhone; CPU iPhone OS 7_1_2 like Mac OS X) AppleWebKit/537.51.2 (KHTML, like Gecko) Version/7.0 Mobile/11D257 Safari/9537.53"`.

Still trying to pin-point the culprit though.

bstst:
Looking in to it a bit more. For some strange reason on iOS7 Safari (the older iPhone I mentioned above), when I have an observable array, e.g.:
`@observable items = [];`

Then elsewhere in the componentWillMount method (doesn't really matter if `item` is a simple or observable object):
`this.items.push(item);`

The stack moves to `ObservableArray.prototype.replace = function (newItems) {`, not `ObservableArray.prototype.push` as it does on new browsers.

From there on the stack breaks, obviously, as newItems is not an array but the passed in object.
Any ideas?

bstst:
Was able to make it work by making the order of non-enumerable properties of the object to correspond to the order of the prototype methods:

```
makeNonEnumerable(ObservableArray.prototype, [
    "constructor",
    "observe",
    "clear",
    "replace",
    "toJSON",
    "peek",
    "find",
    "splice",
    "push",
    "pop",
    "shift",
    "unshift",
    "reverse",
    "sort",
    "toString",
    "toLocaleString",
    "remove",
    "split"
]);
```

Is that something that old safari might be doing with the prototype? By the looks of it the methods are getting mapped incorrectly.

bstst:
[...]

PR #201, opened 2016-04-19
fix makeNonEnumerable bug in iOS 7.x 

luosong:
makeNonEnumerable ObservableArray.prototype have weird result in iOS 7.x.
**toString** changed to **toLocalString**, **toLocalString** changed to **remove**, but **remove** changed **toString** !!!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "chanjet-mobx",
+  "name": "mobx",
   "version": "2.1.3",
   "description": "Simple, scalable state management.",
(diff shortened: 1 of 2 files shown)


mweststrate:
Thanks for the PR! This sounds awkwardly similar to issue #160 (see the bottom comments). Did you base this on 2.1.2 or higher? If the issue still occurred in that version for you I'll merge it.

mweststrate:
Never mind, it is based on that one, I'll released it as 2.1.4. Thanks again!

PR #221, opened 2016-04-27
Smalldots MobX Store

hnordt:
Add Smalldots MobX Store to related projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ And finally kudo's for all the people that believed in, tried and validated MobX
 * [rx-mobx](https://github.com/chicoxyzzy/rx-mobx) Convert Mobx observables to RxJS and vice versa
 * [mobx-store](https://github.com/AriaFallah/mobx-store) A lowdb inspired data store with declarative querying, observable state, and easy undo/redo.
 * [reaxor](https://github.com/KadoBOT/reaxor) Boilerplate for better state management, styling, testing and cleaner code
+* [Smalldots MobX Store](https://github.com/smalldots/mobx-store) Store API for MobX
 
 _Feel free to create a PR to add your own!_
 


AriaFallah:
😱  another mobx-store

hnordt:
Yeah, I'm opensourcing the approach I'm using on my production apps :D

AriaFallah:
@hnordt I plan to eventually do the same with https://github.com/AriaFallah/mobx-data, but I haven't gotten to that stage yet. I'd like it to be like https://github.com/emberjs/data, but much simpler.

hnordt:
I don't know Ember Data but it's a great idea.

Personally I don't like to create big libraries, I prefer tiny modules.

For example, I don't think Smalldots MobX Store needs to worry about storage. It provides two methods (`hydrate` and `toJSON`) that enables you to implement the way you like it.

AriaFallah:
Agreed that's actually why I made my own mobx-store first. mobx-data would just be responsible for all the boilerplate related to fetching from a REST API.

mweststrate:
Cool to see that it is so easy to add standardized store approaches on top of MobX. Keep up the good work :)

hnordt:
@AriaFallah in my apps I just create an `api.js` file and export some functions to be used with my `appStore`, for example: `appStore.setAsync('users', api.getUsers())`.

Looking forward to see your approach!

```
// api.js

import appStore from './appStore'

export function request({
  method = 'GET',
  path,
  headers,
  body,
  ...other
}) {
  const { token } = appStore.get('auth') || {}
  return fetch(`https://dev.4dhealthware.com/api${path}/`, {
    ...other,
    method,
    headers: {
      ...headers,
      'Accept': 'application/json',
      'Content-Type': 'application/json',
      'Authorization': token ? `Token ${token}` : null
    },
    body: JSON.stringify(body)
  }).then(response => {
    return response.ok ? response.json() : Promise.reject(response)
  })
}

export function login(email, password) {
  return request({
    method: 'POST',
    path: '/auth/login',
    body: { email, password }
  }).then(({ key: token }) => ({ token }))
}

export function getUser() {
  return request({ path: '/auth/user' })
}
```

MiracleBlue:
I do love this idea @AriaFallah, I've already started on my own implementation for work (based on Ember Data), perhaps we can join forces!

AriaFallah:
@MiracleBlue perhaps! But I think you should go on without me for the time being.

Right now I'm not working with APIs at all because the localstorage features of [my mobx store](https://github.com/AriaFallah/mobx-store) have made it so easy to prototype without an API. I just write and read all my data from there. A lot of my pet projects have died because I was too lazy to write an API, but now the one I'm currently working on is still alive and well.

I completely went off topic, but I just think this is really cool:

MiracleBlue:
Good to hear :) well hopefully work will allow me to open source the one I'm working on now.

PR #226, opened 2016-04-29
Fix typo in README.md

mattdistefano:
> All components render smartly but are defined in a dumb manner. MobX will **simple** make sure the components are always re-rendered whenever needed, but also no more than that.

Now "simply".


mweststrate:
Tnx :)

PR #233, opened 2016-05-03
Clarifying normalization

mattdistefano:
PR just fixes a typo. Related, though - would this benefit from a caveat along the lines of "unlike many Flux architectures, with MobX there is no need to normalize your data _for use in your store_"? I suppose it's obvious in context, but the existing wording could maybe be misinterpreted as meaning mobx has some bearing on, say, your database normalization.

### DIFF
--- a/docs/best/store.md
+++ b/docs/best/store.md
@@ -111,7 +111,7 @@ There is no need to treat your client-side application state as some kind of dat
 Real references, cyclic data structures and instance methods are powerful concepts in JavaScript.
 Domain objects are allowed to refer directly to domain objects from other stores.
 Remember: we want to keep our actions and views as simple as possible and needing to manage references and doing garbage collection yourself might be a step backward.
-Unlike many Flux architectures, with MobX there is no need to denormalize your data, and this makes it a lot simpler to build the _essentially_ complex parts of your application:
+Unlike many Flux architectures, with MobX there is no need to normalize your data, and this makes it a lot simpler to build the _essentially_ complex parts of your application:
 your business rules, actions and user interface.
 
 Domain objects can delegate all their logic to the store they belong to if that suits your application well.


mweststrate:
I agree that it is a bit confusing term, but people familiar with flux well recognize it, so I guess it's okay. Thanks for the PR.

mattdistefano:
Yeah I guess my argument in favor of some additional context around the term is just that not everyone reading the docs will be fluent in Flux concepts. But given the ecosystem around the project it's probably not crazy to assume some baseline familiarity either.

oyeanuj:
@mweststrate Not sure if this is the right issue, but I was a little unclear about that line in the README.

> With MobX you don't need to normalize your data.

FWIW, I am coming from the Redux with a fair bit of experience, and read the above line, and immediately wondered 'why'? Can you elaborate on it? I imagine the data stays nested but how exactly does one deal with data with nested entities?

I am sure there are other n00bs like me out there :)

Thank you!

mweststrate:
@oyeanuj there is no short answer here I think, but the basic idea is that because data is mutable, you can have referential transparency, making it safe to refer from one object to another. In redux it would be unsafe to do `message.author = someUserObject` because on any future update on the `someUserObject` the state will contain a fresh user object, but your `message.author` would still be referring to the old one. With Mobx you don't need to replace the complete state, but just update the existing `someUserObject`, as long as the object is just updated, the `message.author` will never be stale (referential transparency; changes are visible through a reference like `message.author.name`), hence you don't need the normalization

oyeanuj:
@mweststrate thank you for that helpful explanation, exactly what I was looking for. 

Just one follow-up - how does it work when we have the same entity being received in different API calls.

In your above example, 
```javascript
message.author = someUserObject
```

We have a couple of different API calls which deal with user entities - lets say, an API call which gets a list of authors, and an API which gets a list of commenters - both receiving a list of user entities.

Within the normalized world, you try and normalize the data and update the user entities list with the updated user entity from the API, so `message.author = someUserId` and that `userId` will always have the latest. How does the same translate in MobX? Does it become .. 

```javascript
message.author = someUserObjectId
post.commenter = someUserObjectId

//And then a list of users somewhere 
users[someUserObjectId] = someUserObject
```


In other words, do we try to ensure that list of commenters and list of authors all just contain a reference to the `someUserObject` which is where the latest state of the object is? Or is there a more 'mobX' way of dealing with this?

PR #243, opened 2016-05-10
Fix component render call

cbenz:
It seems this line has been copy-pasted with an extra unnecessary prop?

### DIFF
--- a/docs/refguide/observer-component.md
+++ b/docs/refguide/observer-component.md
@@ -70,7 +70,7 @@ import {observable} from "mobx"
 	}
 })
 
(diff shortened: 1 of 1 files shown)


capaj:
@cbenz no, I don't think so. See the component implementation here: https://github.com/cbenz/mobx/blob/43862b02939ef56016edfe99327160a141b6f4b8/docs/refguide/observer-component.md#stateless-function-components

mweststrate:
Note that in the render function the `timerData` is still accessed through the props. This is kinda of intentional, although the component can indeed access the state directly from the closure, it will become harder to test the component as stand alone unit if you don't use props to wire the component up.

PR #255, opened 2016-05-20
Update extras.md

mkartchner994:
Looks like the [Miscellaneous](http://mobxjs.github.io/mobx/refguide/extras.html) page in the GitBook does not render the JS block correctly without the closing ```


mattruby:
Thanks!  Nice catch.

PR #259, opened 2016-05-24
Fixed bug.

chenxsan:
When double clicked, the prompt showed, if i canceled it, the value was cleared. I thought this might be a bug, I would expect the value to be kept.


mweststrate:
That makes a lot of sense indeed, thanks for spotting and fixing!

PR #265, opened 2016-05-26
Fix typo in error message.

mtyaka:
"... calculating **the** your computed value ..." -> "... calculating your computed value ..."


mweststrate:
Thanks!

Issue #293, opened 2016-06-02
Naming autoruns


jamiewinder:
I know you can name an autorun by naming the function parameter:

```javascript
const disposer = autorun(function iAmANamedAutorun() {
    // ...
});
```

Which is great for debugging, but it can be a bit of a pain to fully utilize. For example, if I want to do something like this:

```javascript
const fn = () => {
     // ...
}
fn.name = 'update autorun for item#' + item.id;
const disposer = autorun(fn);
```
It's a little convoluted. It'd be nice if I could do:
```javascript
const disposer = autorun('update autorun for item#' + item.id, () => {
});
```

This'd also align nicely with how an `[@]action` can be named. I'm happy to try adding this if there is any interest?

mweststrate:
Sounds like a very good idea! Feel free to go ahead, I don't think it will introduce overload issues

PR #294, opened 2016-06-02
Named autorun function

jamiewinder:
Implements optionally named autoruns suggested here:
https://github.com/mobxjs/mobx/issues/293

### DIFF
--- a/src/api/autorun.ts
+++ b/src/api/autorun.ts
@@ -127,6 +127,7 @@ export function autorunAsync(arg1: any, arg2: any, arg3?: any, arg4?: any) {
 
 	if (scope)
 		func = func.bind(scope);
+
 	let isScheduled = false;
 
 	const r = new Reaction(name, () => {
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for the PR and looking good! Some notes / room for improvements (don't feel obliged to do so, but help is appreciated nonetheless :))

* [ ] add unit tests
* [ ] update docs
* [ ] add to release notes
* [ ] introduce optional name param to `autorunAsync`, `when` and `reaction` as well?

jamiewinder:
Thanks for the feedback so far!

I've now added the same parameter to `autorunAsync`, `when`, and `reaction`. I've hopefully kept the same logic and defaults (I notice `autorunAsync` has delay default to 1ms, and reaction defaults to 0ms - intentional?). 

I haven't added the parameter to `autorunUntil` - presumably if it's deprecated, it isn't necessary?

I don't mind updating the documentation once you're happy with the code changes. Unit tests aren't my strong point however, so I might not be the best person to write those!

jamiewinder:
Just to add - since the `name` is now determined in the `when` function rather than `autorun`, we have the opportunity to default to `When@###` rather than `Autorun@###` as it will currently. This appears to fit with what the defaults are in `autorunAsync` (AutorunAsync@###) and `reaction` (Reaction@###).

mweststrate:
Merged the changes and will be part of 2.3. Thanks!

jamiewinder:
Great! Not sure if my last comment was noticed (about defaulting the name of a `when` to When@ rather than Autorun@) but it can be updated if needed. Happy to help, and many thanks for the library (and accepting my first every GitHub PR!)

mweststrate:
@jamiewinder thanks! missed that indeed :) Thanks for contributing!

Issue #302, opened 2016-06-03
`intercept()` and `observe()` don't track nested observables


antitoxic:
```js
var appState = observable({
    counter: 2,
    todos: [{
        title: 'Submit mobx issue',
        done: false
    }]
})
intercept(appState, function(change) {
    console.log(change)
    return change;
})
observe(appState, function(change) {
    console.log(change)
    return change;
})
appState.todos[0].title = "Get stuff done";
```

No output. Bug or feature?

AriaFallah:
Intended unfortunately. Although I would like a way to be able to deeply observe as well.

antitoxic:
So, `autorun` tracks only what properties that has been accessed (*just like `observer` I guess*)

Make sense but again I was a bit surprised. I will try to do a pull request for the docs and add a simple clarifying sentence.

antitoxic:
Ah completely messed up previous comment. I meant to add that I got confused with autorun as well.

As for `intercept()` and `observe()` I will also include some clarification and the docs and make a pr. 

I also wish there was a nested option for `()` and `observe()`. Otherwise this forces me to break down my state to stores on every nested object/array.

mweststrate:
Doc PR's are more then welcome!

If you need to deep observe, I recommend using `autorun` or `reaction`. Did that fit? It is also good to note that you can make observe handlers generic; e.g. reuse the same callback over and over again because the causing object is always in `change.object`. (You still need to register them all though, so that is why I usually opt for `autorun` / `reaction`

matthewmueller:
so why this is intentional? is there a limitation in intercepting nested values?  you can also listen on:

```js
appState.todos.intercept(fn)
```

but maybe there's harm in handling both?

AriaFallah:
@matthewmueller 

From what I know, `intercept` basically works like this, and thus can't look any deeper into your object:

```js
spy(function(change) {
  // references match
  if (change.object === myObject) {
    // Do stuff with change info
  }
})
```

I experimented a lot with deep observation, and short of adding a non-enumerable field marking the parent of a given observable (which is in my opinion too much work), it didn't work well.

---

A possible alternative is wrapping all the changes you want to intercept inside of `action`, and then intercepting the action like so

```js
const myObject = { a: 1 }

const modifyObject = action('modify', function() {
  myObject.a = 5
})

spy(function(change) {
  // action which modifies all your objects detected!
  if(change.type === 'action' && change.name === 'modify') {
[...]

PR #304, opened 2016-06-03
Clarifications on what is observed and when

antitoxic:
I really wish `intercept` and `observe` worked on nested observable structures but until then let's get the docs clear.

Related to #302


mweststrate:
:) tnx!

Issue #306, opened 2016-06-06
[ComputedValue] It is not allowed to assign new values to computed properties [React Native, iOS 8]


aksonov:
I'm not changing any computed property. No exception for iOS 9, but only iOS 8. Problem happens during intensive load of messages during React rendering with that computed value (i.e. computed property is constantly changed because observable data is changed)

![simulator screen shot 06 jun 2016 17 25 26](https://cloud.githubusercontent.com/assets/1321329/15827562/e7685c8e-2c0b-11e6-8bb7-cc5b8665086d.png)

mweststrate:
Could you also include all console output?

aksonov:
I don't see any logging within console. If you meant full error stack trace, it is very very long and i don't see how to get it as text, i can't cut & paste from that screen... Here is beginning
![simulator screen shot 07 jun 2016 17 32 54](https://cloud.githubusercontent.com/assets/1321329/15864090/e98d4482-2cd5-11e6-8e25-4432ee257d7b.png)

mweststrate:
Could you check whether MobX 2.3 improves this?

aksonov:
Just found that 2.3 brakes @autobind support (with autobind-decorator module)- my app cannot work at all (i.e. `this` is not visible within @action methods). Could you fix it?

mweststrate:
Could you make a reproduction? cannot 100% agree that it is fixable, (I normalized subtle difference between typescript and babel, so I had to find a subset of patterns which is supported by both). 

Not however that autobind is not needed if field initializers are enabled in your babel rc, eg. the following is already bound to this:

`@action myAction = () =>  { }`

mweststrate:
And babel also has a syntax for autobinding, something like `::field`, but I always forget the name :)

aksonov:
With 2.3 i've got error `this.model` is undefined.

```javascript
@autobind
export default class XmppStore {

  @action onConnect(){
    this.model.connected = true;
    this.model.connecting = false;
  }
}
```

mweststrate:
Would it be acceptable for you to write it as `@action onConnect = () => {...`? I think it might relate to the initialization order of the decorators.

aksonov:
@mweststrate It works under iOS 8 now, thanks!!

aksonov:
It is interesting what the reason was...

mweststrate:
[...]

Issue #211, opened 2016-04-22
Make modifiers api simpler & uniform


mweststrate:
I think the use of modifiers can be made more intuitive. In short, instead of the (imho) weird wrapping functions as modifiers, use a settings object that can be passed to both `extendObservable` and `(@)observable`. 

Some examples:

(ES6) Create an observable reference to a non-observable array:
`@observable prop = asReference([])` -> `@observable({ reference: true }) = []`

(ES6) Create an observable array, but do not make the children observable:
`@observable prop = asFlat([])` -> `@observable({ shallow: true }) = []`

(ES5) Create an observable reference to a non-observable array:
`extendObservable(this, { prop: [] }, { reference: true })`

This means that you will sometimes need two `extendObservable` calls where previously one would do, but I think the api will be cleaner.

So the modifiers would get a slight rename as well: `asStructure` ->  `struct`, `asReference` -> `reference`, `flat` -> `shallow`.

It could be interesting to merge the concepts of `shallow` and `reference`, as I don't think anybody is using shallow objects, and for primitive values shallow and reference behave the same. But in that case it would not be possible anymore to create references to non-observable arrays anymore.

Issues #197 #173 #55 should be addressed by these changes.

What do you guys think? Suggestions are welcome!

AriaFallah:
I'm not so sure about this one. For example in my code for mobx-store I'm using

```js
db.set(key, createData(db, key, value.__data))
```

where createData is

```js
export function createData(obj: Object, key: string, value: Array<any> = []) {
  return {
    obs: value, // The observable array
    __data: asReference(value) // The underlying array
  }
}
```

If you change the API, how would you handle cases like this?

Also in the case that you have something like

```js
observable({
  a: asFlat([]),
  b: asStructure({}),
  c: asReference([])
})
```

I'd need to call extendObservable 3 times, which doesn't seem so intuitive.

As a side note:

The names `asStructure` and `struct` never made that much sense to me. Something with `deep` in it is a bit more intuitive considering it's deeply comparing the observable value to the previous one.

hnordt:
I liked te idea. The current modifiers API is weird, the new proposal are cleaner.

AriaFallah:
@mweststrate Would it be possible to just have both? Keeping them as exports, but also updating the `@observable` decorator?

mweststrate:
[...]

PR #313, opened 2016-06-10
Fix decorators

### LINKED ISSUE #288: Observing a store property seems to work when using Typescript but not when using ES6/Babel
Running the following snippet with ES6/Babel leads to an invariant error:

```javascript
class Store {
  @observable foo = ''
}

const store = new Store();

mobx.observe(store, 'foo', () => {
  console.log('Change observed');
}, true);

// --> Invariant failed: first argument of observe should be an (observable)object or observableMap if a property name is given
```

Running it with Typescript (tried in JSBin) seems to work just fine.

mweststrate:
* decorators are now stored on prototype if possible: gives significant perfomance improvements in CPU and MEM usage (30-50% in benchmarks)
* subtle differences between typescript and babel are now largely gone
* Object.keys(instance) will no longer report observable properties, in contrast to for .. in ...
* more code reuse between decorators
* improved test coverage
* This PR also fixes #288 and makes it a lot easier to make decorators configurable (see also #211)

Benchmark of creating and modifying 100K objects:

language | instantiating (ms) | modifying(ms) | peak mem
---- | --- | ---- | ---
ts (before PR) |  1308 |  7491 | 360MB
ts | 938 | 2222 | 198MB
babel (before PR) | 14 | 5348 | 374MB
babel | 19 | 2046 | 194MB

### DIFF
--- a/src/api/computeddecorator.ts
+++ b/src/api/computeddecorator.ts
@@ -1,36 +1,15 @@
 import {ValueMode, getValueModeFromValue, asStructure} from "../types/modifiers";
 import {IObservableValue} from "./observable";
-import {asObservableObject, setObservableObjectInstanceProperty, defineObservableProperty} from "../types/observableobject";
-import {invariant, assertPropertyConfigurable} from "../utils/utils";
+import {asObservableObject, defineObservableProperty} from "../types/observableobject";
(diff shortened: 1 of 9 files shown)


mweststrate:
This *might* also fix #306

PR #322, opened 2016-06-14
Fix inconsistent title/desc in getting-started doc

spion:
Makes it consistent with the title tag, which says ten. 

Not normally an issue except when shared on social networks. e.g. facebook messenger:

image

Is it five? Or ten? 😀


mweststrate:
Thanks for noticing!

PR #335, opened 2016-06-17
add ts-mobx-react-starter-kit to README

Torwori:
Hey guys,

I made a starter kit, it would be nice if it could be linked here. :smile:


mweststrate:
Nice!

PR #342, opened 2016-06-18
Fixed #341 ObservableArray.find return null instead of undefined

### LINKED ISSUE #341: ObservableArray.find returns null instead of undefined
Hello,
searched in the issues but did not find anything. Is this a bug? [MDN Array find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) entry states:
> The find() method returns a value in the array, if an element in the array satisfies the provided testing function. Otherwise undefined is returned.

But in `observablearray.ts` it returns `null`:
```TypeScript
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find
find(predicate: (item: T, index: number, array: ObservableArray<T>) => boolean, thisArg?, fromIndex = 0): T {
  this.$mobx.atom.reportObserved();
  const items = this.$mobx.values, l = items.length;
  for (let i = fromIndex; i < l; i++)
    if (predicate.call(thisArg, items[i], i, this))
      return items[i];
  return null;
}
```
In my code I am checking on `undefined`.

hellectronic:
[MDN Array find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) entry states:
> The find() method returns a value in the array, if an element in the array satisfies the provided testing function. Otherwise undefined is returned.

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -276,7 +276,7 @@ export class ObservableArray<T> extends StubArray {
 		for (let i = fromIndex; i < l; i++)
 			if (predicate.call(thisArg, items[i], i, this))
 				return items[i];
-		return null;
+		return undefined;
 	}
 
 	/*


mweststrate:
Thanks!

Issue #333, opened 2016-06-16
App broken after upgrading mobx to 2.3.1


otbe:
Hi,

after upgrading mobx to 2.3.1 from 2.2.2 my app is somehow broken. Simple ```autorun```s dont work  and components are not reactive anymore. It has been discussed at gitter and this issue should keep attention on this as long as is needed.

Project is at https://github.com/otbe/dwatch/tree/mobx-2.3.1
Just ```npm i``` and ```npm run test```. One test will fail, because it relies heavily on ```autorun()```.

Thanks!

mweststrate:
Hi @otbe 

I am unfamiliar with Karma, can you elaborate how I can use a debugger in your test suite?

otbe:
Change singleRun to false in karma.conf.js, visit http://localhost:<port> (port will be logged) and click "Debug" in the upper right corner. 
But I don't know if this performs very well because it is an electron/node project. Will try it later today. :)

andykog:
Seems, that @otbe problem is not typescript, targeting ES3, but rather `reflect-metadata` that does not returns descritor, if 4th argument is present, but undefined. https://github.com/rbuckton/ReflectDecorators/blob/master/Reflect.js#L562
#343 can fix this, can't find better solution :-/

otbe:
Hi @andykog,

Im not targeting ES3 with TS, but I think you are right. Its a mismatch between TS and ```reflect-metdata```.
```Reflect.decorate``` returns no descriptor if no descriptor was passed in. TS only defines a property if a descriptor was returned. Somehow wired :P
But who's right?

andykog:
I have the same question, didn't find any specification on Reflect.decorate

otbe:
Opened an issue for this https://github.com/Microsoft/TypeScript/issues/9265

mweststrate:
Did somebody check whether #343 fixes the problem? It might be interesting to incorporate the patch in that cases, regardless whether it is TS or Reflect is wrong?

andykog:
I did :-)

mweststrate:
Ok, I think we can leave a big ugly comment pointing to this issue then and (if / as long as) nothing else breaks incorporate it.

mweststrate:
@andykog I just tested the change, but it didn't fix the issue, did  you install it in `app/` in the project?

andykog:
Sorry, I've tested isolated testcase `should persist changes to persistence`, but seems that another one fails now: `should recover settings from persistence if present`
[...]

PR #343, opened 2016-06-19
Fix decorator in typescript targeting ES3

andykog:
When tsc target is set to ES3 (default), it won't add 3'rd argument for decorator https://github.com/Microsoft/TypeScript/blob/v1.8.10/src/compiler/emitter.ts#L5707
Then it won't use Object.defineProperty https://github.com/Microsoft/TypeScript/blob/v1.8.10/src/compiler/emitter.ts#L364

I don't know, whether we should care about this setup, because mobx itself is not usable under ES3 environment, but this fixes problem on updating from mobx 2.2.2

### DIFF
--- a/src/utils/decorators.ts
+++ b/src/utils/decorators.ts
@@ -50,6 +50,8 @@ export function createClassPropertyDecorator(
 			};
 			if (arguments.length < 3) {
 				// Typescript target is ES3, so it won't define property for us
+				// or using Reflect.decorate polyfill, which will return no descriptor
+				// (see https://github.com/mobxjs/mobx/issues/333)
 				Object.defineProperty(target, key, descriptor);
 			}
 			return descriptor;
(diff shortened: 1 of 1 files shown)


andykog:
Fiddle, representing the problem: https://jsfiddle.net/Sl1v3r/r3oLj9fz/

src/utils/decorators.ts
```diff
@@ -48,6 +48,11 @@ export function createClassPropertyDecorator(
 					}
 				}
 			};
+			if (arguments.length < 3) {
+				// Typescript target is ES3, so it won't define property for us
+				Object.defineProperty(target, key, descriptor);
```
mweststrate: Hey @andykog, what will this fix? If target is ES3, then `Object.defineProperty` should not be available in the first place right? I guess people should just not compile for ES3. So maybe we should just throw an exception here?

andykog: I mean some people can forget to specify typescript target, i thought, it would be nice if mobx still works. Fixing jsfiddle is also nice.
And it fixes Reflect.decorate (#333)

mweststrate: Ha sorry yeah read this in the wrong other :) Thanks for sorting this out!
I'll run some more tests but I think this will do.

Op ma 20 jun. 2016 om 16:32 schreef Andy Kogut <notifications@github.com>:

mweststrate:
Thanks again for sorting this out!

PR #348, opened 2016-06-20
Add Symbol.iterator to maps

benjamingr:
support iteration protocol, this lets libraries and `for...of` loops directly consume 
maps like native maps.

Note that a more correct type would be an iterator, something like:

```js
    entries(): Iterator<[string, V]>;
    keys(): Iterator<V>;
    values(): Iterator<V>;
```

### DIFF
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -203,7 +203,11 @@ export class ObservableMap<V> implements IInterceptable<IMapWillChange<V>>, ILis
 	values(): V[] {
 		return this.keys().map(this.get, this);
 	}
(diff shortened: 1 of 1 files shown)


benjamingr:
Old node failing on no symbols. I assumed TypeScript shimmed those. https://github.com/Microsoft/TypeScript/issues/980 

Any idea?

mweststrate:
Closed in favor of https://github.com/mobxjs/mobx/commit/606a1699cbfd52eb5fe6ee9495bbdf2553f74e05. Will be released as part of the next version

PR #359, opened 2016-06-23
Make observable decorator define instance properties

andykog:
Since v2.3.1 `@observable` Typescript decorator does't define instance property, so, for example:
```
class Store {
   @observable cats = 5;
}
var store = new Store();
store.hasOwnProperty('cats'); // returns false
Object.assign({}, store); // returns {}
```

Note, that it didn't define instance property before v2.3.1 in Babel version, only in Typescript one.

@mweststrate, if this is expected behaviour, please close my pr.

### DIFF
--- a/src/api/observabledecorator.ts
+++ b/src/api/observabledecorator.ts
@@ -11,7 +11,7 @@ const decoratorImpl = createClassPropertyDecorator(
 		if (typeof baseValue === "function")
 			baseValue = asReference(baseValue);
 		const adm = asObservableObject(target, undefined, ValueMode.Recursive);
-		defineObservableProperty(adm, name, baseValue, false);
+		defineObservableProperty(adm, name, baseValue, true);
 		allowStateChangesEnd(prevA);
 	},
 	function (name) {


andykog:
I could add tests for that, by the way.

andykog:
This pr fixes #333

otbe:
I can confirm this. With both PRs DWatch is back again :D

mweststrate:
Hmm I think this change undoes the performance gains of #313, I'm gonna check whether there is a better solution...

donaldpipowitch:
I really would prefer to accept this PR. I don't favor the performance gains in this case, but clarity interoperability. Version 2.3 just works not the way I'd expect it ([example](https://jsfiddle.net/wv3yopo0/258/)).

mweststrate:
I agree, the impact of this is bigger than it should be. I am gonna revert
that change this weekend. I think part of the optimization can still be
applied

Op vr 24 jun. 2016 08:39 schreef Donald Pipowitch <notifications@github.com

mweststrate:
Good news! I was able to get almost the same performance characteristics (especially, saving memory) as defined in #313 without loosing property ownership on the instances. Will release this soon as 3.3.3

language | instantiating (ms) | modifying(ms) | peak mem
---- | --- | ---- | ---
ts (v1) |  1308 |  7491 | 360MB
ts (v2) | 938 | 2222 | 198MB
ts (v3) | 1478 | 2697 | 209MB
babel (v1) | 14 | 5348 | 374MB
babel (v2) | 19 | 2046 | 194MB
babel (v3) | 27 | 2449 | 193MB

Babel decorators are still a bit yuki, e.g. they initialize lazily, but that cannot be solved without a better transpiler plugin

PR #365, opened 2016-06-27
add new boilerplate info in readme.md file

tapos007:
i add a new boilerplate to work reactjs and mobX


mweststrate:
Thanks! Merged. Note that I'm about to to remove the section from the readme and give it a standalone page in the official docs, but I'll make sure it has a place over there as well.

PR #367, opened 2016-06-27
Introduce dispose()

andykog:
How about having a function, that disposes reaction, that it was called in.

Before:
```js
const waitForDataLoadingDisposer = reaction(
  () => myStore.data,
  data => {
    if (data) {
      doSomethingWithData(data);
      waitForDataLoadingDisposer();
    }     
  }
)
```

After:
```js
reaction(
  () => myStore.data,
  data => {
    if (data) {
      doSomethingWithData(data);
      dispose();
    }     
  }
)
```

Before:
```js
const timerDisposer = autorun(() => {
   console.log(time.get(), 'seconds left')
   if (time.get() <= 0) timerDisposer();
})
```

After:
```js
autorun(() => {
   console.log(time.get(), 'seconds left')
   if (time.get() <= 0) dispose();
})
```

@mweststrate, I can add tests, if this idea is worth.

### DIFF
--- a/src/api/dispose.ts
+++ b/src/api/dispose.ts
@@ -1,0 +1,9 @@
+import {globalState} from "../core/globalstate";
+
+export function dispose() {
+	if (globalState.currentReaction) {
+		globalState.currentReaction.dispose();
+	} else {
+		console.error('mobx: dispose() must be called inside reaction');
+	}
+}
--- a/src/core/globalstate.ts
+++ b/src/core/globalstate.ts
@@ -45,6 +45,12 @@ export class MobXGlobals {
 	isRunningReactions = false;
 
 	/**
+	 * Current reaction (if any is running).
+	 * Reactions are run after derivations using a trampoline.
+	 */
+	currentReaction: Reaction = undefined;
+
+	/**
 	 * List of observables that have changed in a transaction.
 	 * After completing the transaction(s) these atoms will notify their observers.
 	 */
--- a/src/mobx.ts
+++ b/src/mobx.ts
@@ -47,6 +47,7 @@ export { expr                                                 } from "./api/expr
 export { toJSON, toJS                                         } from "./api/tojs";
 export { ITransformer, createTransformer                      } from "./api/createtransformer";
 export { whyRun                                               } from "./api/whyrun";
+export { dispose                                              } from "./api/dispose";
 
 export { Lambda                                               } from "./utils/utils";
 export { SimpleEventEmitter, ISimpleEventListener             } from "./utils/simpleeventemitter";
--- a/src/core/reaction.ts
+++ b/src/core/reaction.ts
@@ -157,9 +157,12 @@ export function runReactions() {
 		if (++iterations === MAX_REACTION_ITERATIONS)
 			throw new Error("Reaction doesn't converge to a stable state. Probably there is a cycle in the reactive function: " + allReactions[0].toString());
 		let remainingReactions = allReactions.splice(0);
-		for (let i = 0, l = remainingReactions.length; i < l; i++)
+		for (let i = 0, l = remainingReactions.length; i < l; i++) {
+			globalState.currentReaction = remainingReactions[i];
 			remainingReactions[i].runReaction();
+		}
 	}
+	globalState.currentReaction = undefined;
 	globalState.isRunningReactions = false;
 }
 
--- a/test/api.js
+++ b/test/api.js
@@ -18,6 +18,7 @@ test('correct api should be exposed', function(t) {
 		'autorunUntil',
 		'computed',
 		'createTransformer',
+		'dispose',
 		'expr',
 		'extendObservable',
 		'extras',


mweststrate:
Hi Andy!

I really like this idea, I actually think it might make the internal implementation of several reactions simpler as well. I have an unexpected conference talk this weekend (RuhrJS), but I'll come back to you after that with some review comments, questions etc!

src/core/globalstate.ts
```diff
@@ -45,6 +45,12 @@ export class MobXGlobals {
 	isRunningReactions = false;
 
 	/**
+	 * Current reaction (if any is running).
+	 * Reactions are run after derivations using a trampoline.
+	 */
+	currentReaction: Reaction = undefined;
```
mweststrate: Instead of introducing a separate state var it is probably simpler to just peek into the `derivationStack` and check whether it is an instanceof Reaction

src/mobx.ts
```diff
@@ -47,6 +47,7 @@ export { expr                                                 } from "./api/expr
 export { toJSON, toJS                                         } from "./api/tojs";
 export { ITransformer, createTransformer                      } from "./api/createtransformer";
 export { whyRun                                               } from "./api/whyrun";
+export { dispose                                              } from "./api/dispose";
```
mweststrate: The name is a bit generic, maybe `disposeCurrentReaction`? That also makes it more clear that it should be called from inside a reaction

src/api/dispose.ts
```diff
@@ -0,0 +1,9 @@
+import {globalState} from "../core/globalstate";
+
+export function dispose() {
+	if (globalState.currentReaction) {
+		globalState.currentReaction.dispose();
+	} else {
+		console.error('mobx: dispose() must be called inside reaction');
```
mweststrate: `'[mobx] ...'`  for consistency :)

src/mobx.ts
andykog: I thought about it as well.

src/core/globalstate.ts
andykog: Good idea!

mweststrate:
Hey Andy,

See my comments. Should be good to go after adding some tests. I think the `disposeImmediately` variable in the `when` function in `api/autorun.ts` could be eliminated with this feature?

andykog:
@mweststrate yes, I'll make it in the next commit.

src/core/globalstate.ts
andykog: @mweststrate, reaction's `effect` function is not running inside derivation, so if we use looking into `derivationStack` we can't dispose reaction like it was suggested:
```js
reaction(
  () => myStore.data,
  data => {
    if (data) {
      doSomethingWithData(data);
      disposeCurrentReaction();
    }     
  }
)
```

Insead we'll have to do it like this (may be less user-friendly, I think):

```js
reaction(
  () => {
    const data = myStore.data;
    if (data) disposeCurrentReaction();
    return data;
  },
  data => {
    if (data) {
      doSomethingWithData(data);
    }     
  }
)
```

We have 3 options:
1. Pop derivation from `derivationStack` _after_ effect is called (I'm not sure that is ideologically correct)
2. Introduce a state var `currentReaction` (currently in this PR), and remove state var `isRunningReactions` in favour of `currentReaction`.
3. Accept that `disposeCurrentReaction` must be called inside reaction's expression.

Which way you think is better?

mweststrate: Good point. Same holds for `when`, but it's effect disposes the reaction anyway. Approach 1) doesn't gonna work nice, effect is not a tracked function at all, so that might only be confusing. For 2) the same, you would get a fake kind of `currentReaction`, because effect doesn't really running reaction. 

That leaves only 3. Or either of the following: pass in a second parameter to `effect`; the reaction disposer, or have a  separate state var specific for this purpose, something like `currentReactionDisposer` or something. That is a bit iffy though, as this would need to be a stack as autorun's could be nested.

Would a viable alternative to this problem be always passing in the reaction object to the callback functions of reactions? I think meteor does something similar, and then you can always write things like:

```javascript
reaction(
  (r) => {
    r.dispose()
  },
  (data,r) => {
    r.dispose()
  }
)
```

Reactions are nowhere really visible to the outside world, so it could also be that the last argument is just the disposer func. But passing the complete reaction as object seems more flexible towards the future.

andykog: bq. Would a viable alternative to this problem be always passing in the reaction object to the callback functions of reactions?
I think, that's even better then my initial proposition! autorun can work the same way:
```
autorun(r => {   r.dispose();  })
```

mweststrate: yeah that is how I came to it :) passing an additional param to the effect of reaction felt inconsistent with the rest of the reaction api's.

andykog:
Will make a new PR, implementing ”passing in the reaction object to the callback functions of reactions"

mweststrate:
awesome!

Op do 21 jul. 2016 om 21:06 schreef Andy Kogut <notifications@github.com>:

PR #392, opened 2016-07-07
Documentation: @observable should be @computed in "when" example

tetsuo:
... since it throws "@observable can not be used on getters, use @computed instead".


mweststrate:
Thanks!

Issue #364, opened 2016-06-27
Array index assignment does not work in Safari


elliotchance:
In Safari 9.1 assigning an object to an array by index breaks anything after that:

https://jsfiddle.net/yfkkk42r/2/

Open the same URL in Chrome and it will work correctly.

mweststrate:
Do you get an exception or something like that? (sorry, no IOS device at hand)

hellectronic:
I tested on Safari 8.0.8 and got:

[Log] Remaining: Make coffee
[Log] Remaining: Spoil tea, Make coffee
[Log] Remaining: Make coffee
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) https://fiddle.jshell.net/js/babel/foo.js.map

It looks like being a jsfiddle problem

elliotchance:
I originally found the bug in our code, which means its not specifically related to JSFiddle.

You get no errors at all. When you assign an object like, the keys in that object are not observed so that if you change one later it never recognises the change.

`Array.push` works correctly, only on assignment to any index (from 0 to the length).

mweststrate:
@elliotchance sorry for the late response! I'm back in town and just tried in on iOS9 device, indeed it doesn't run as expected. For others, to verify, just use this fiddle: https://jsfiddle.net/yfkkk42r/5/. If you *don't* see "iOS9 bug not present", the bug is present. Thanks for reporting, this is an important issue!

andykog:
@mweststrate, the bug is also present on Desktop Safari 9.1.1

Seems, like Safari doesn't respect setting numbered setters on array-like constructor's prototype:

```js
Object.defineProperty(ObservableArray.prototype, '9999', {
		enumerable: false,
		configurable: false,
		set: function() { alert(1); },
		get: function() { return 0; }
});
Object.defineProperty(ObservableArray.prototype, 'zzzz', {
		enumerable: false,
		configurable: false,
		set: function() { alert(1); },
		get: function() { return 0; }
});
var x = new ObservableArray(1);
x['zzzz']; // 0
x['zzzz'] = 1; // ok, alerts 1
z['0000']; // 0
z['0000'] = 1; // fail, does nothing
z['0000']; // 1 // property redefined
```

andykog:
I'll do additional envestigations tomorrow

mweststrate:
@andykog awesome! Thanks in advance

andykog:
I have more precise example, that represents the problem:
```js
function O() {}

[...]

PR #396, opened 2016-07-07
Fix ObservableArray setters in Safari and buffer autoincreace

andykog:
More info: #364

All the tests are now passing in Safari, Chrome, Firefox.

Chrome & FF assign variable name to function name when assigned to variable, so some test were failing.
```js
var myAnonymous = () => {};
myAnonymous.name // 'myAnonymous'
```

fixed buffer autoincrease bug:
```js
var i = observable(new Array(1000))
i.observe(a => console.log(a))
i[999] = "v" // Ok, triggers reaction
i[1000] = "v" // Fail, nothing happen
i[1001] = "v" // Fail, nothing happen
i[1002] = "v" // Fail, nothing happen
i.length // Fail, 1000
```

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -124,8 +124,8 @@ class ObservableArrayAdministration<T> implements IInterceptable<IArrayWillChang
 		if (oldLength !== this.lastKnownLength)
 			throw new Error("[mobx] Modification exception: the internal structure of an observable array was changed. Did you use peek() to change it?");
 		this.lastKnownLength += delta;
-		if (delta > 0 && oldLength + delta > OBSERVABLE_ARRAY_BUFFER_SIZE)
-			reserveArrayBuffer(oldLength + delta);
+		if (delta > 0 && oldLength + delta + 1 > OBSERVABLE_ARRAY_BUFFER_SIZE)
+			reserveArrayBuffer(oldLength + delta + 1);
 		if (delta > 0)
 			for (let i = 0; i < delta; i++)
 				Object.defineProperty(this.array, "" + (oldLength + i), ENUMERABLE_ENTRIES[oldLength + i]);
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -6,6 +6,15 @@ import {IInterceptable, IInterceptor, hasInterceptors, registerInterceptor, inte
 import {IListenable, registerListener, hasListeners, notifyListeners} from "./listen-utils";
 import {isSpyEnabled, spyReportStart, spyReportEnd} from "../core/spy";
 
+// Detects bug in safari 9.1.1 (or iOS 9 safari mobile). See #364
+const safariPrototypeSetterInheritanceBug = (() => {
+	var v = false;
+	const p = {};
+	Object.defineProperty(p, "0", { set: () => { v = true } });
+	Object.create(p)["0"] = 1;
+	return v === false;
+})();
+
 export interface IObservableArray<T> extends Array<T> {
 	spliceWithArray(index: number, deleteCount?: number, newItems?: T[]): T[];
 	observe(listener: (changeData: IArrayChange<T>|IArraySplice<T>) => void, fireImmediately?: boolean): Lambda;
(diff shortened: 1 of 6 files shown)


src/types/observablearray.ts
```diff
@@ -124,8 +133,14 @@ class ObservableArrayAdministration<T> implements IInterceptable<IArrayWillChang
 		if (oldLength !== this.lastKnownLength)
 			throw new Error("[mobx] Modification exception: the internal structure of an observable array was changed. Did you use peek() to change it?");
 		this.lastKnownLength += delta;
-		if (delta > 0 && oldLength + delta > OBSERVABLE_ARRAY_BUFFER_SIZE)
-			reserveArrayBuffer(oldLength + delta);
+		if (delta > 0 && oldLength + delta + 1 > OBSERVABLE_ARRAY_BUFFER_SIZE)
+			reserveArrayBuffer(oldLength + delta + 1);
+		if (delta > 0)
```
mweststrate: these 2 if checks which result in props on the own array are not needed for the patch right? they make arrays a lot slower see https://github.com/mobxjs/mobx/commit/3635fa233eb19e189ff6147df693ae24f342cb48

mweststrate: just tested, is needed for the patch. Merging and releasing this as 2.3.5. Thanks a lot!

andykog: Basically, we can fix safari without it (I've added your commit because you said that you are going to merge it anyway).
I think, if we don't define own array props, all we need is to always define the only one prop, non-enumerable `0` (revert https://github.com/mobxjs/mobx/pull/396/commits/7565600e293c9a584fc3fd937e45131126f9d747).
So do we need making array props enumerable and loose performance?

mweststrate: I did a quick experiment, but without the other entries it didn't seem to
work on my IPad (only define the own prop on index 0), so I just released
it with the above setup, except that I wrapped the iOS bug detecting var
around it, to not slow down the other environments in the same way.

Op vr 8 jul. 2016 om 09:22 schreef Andy Kogut <notifications@github.com>:

mweststrate: Hmm my first test was probably incorrect, I removed the loop, but forgot to change the condition in the constructor that only applies the fix there for zero-length initialized arrays, which wasn't the case in my test. So probably the patch wasn't applied at all in that test.

andykog: Yep, as i said, https://github.com/mobxjs/mobx/pull/396/commits/7565600e293c9a584fc3fd937e45131126f9d747 needs to be reverted :-)

mweststrate: Ah that works, see https://jsfiddle.net/yfkkk42r/8/ / https://github.com/mobxjs/mobx/commit/0ac2d21a2f9778aac4cf1180a099b2c4fbf49555.

Looking good I think?

andykog: @mweststrate, Yes, thats fine

mweststrate: Btw just let me know if you want npm publish rights.

Op vr 8 jul. 2016 om 11:22 schreef Andy Kogut <notifications@github.com>:

andykog: It may be useful for quick publishing some minor fixes. My npm: https://www.npmjs.com/~andykog

mweststrate: done!

Op vr 8 jul. 2016 om 11:32 schreef Andy Kogut <notifications@github.com>:

andykog: Thanks :-)

jsilvestri:
@mweststrate I'm not sure if this is the correct place to post this, but I do think this bug fix introduced a new bug such that when you have a hash in ios9 that has a non-enumarable object on it, and only the key '0' in the hash, when you iterate over the has, it iterates twice!  Here is an example from https://stackoverflow.com/q/52857259/2611078:

Test this out on https://liveweave.com/kKo88G as well:

    // Define a hash
    var totalItems, sum, type, value
    totalItems = {}
    totalItems['0'] = 3

    // This definition of __ob__ is done dynamically by Vue,
    // but I include it here by way of example of what breaks in ios9
    totalItems.__ob__ = new Object()
    Object.defineProperty(totalItems, '__ob__', {
        enumerable: false,
        writable: true,
       configurable: true
      });

    // Loop through the hash
    sum = 0
    for (type in totalItems) {
      value = totalItems[type];
      sum += value;
    }

    // sum is 6 in ios9 Safari -- it loops through the '0' key twice 
    // sum is 3 in all other browsers and newer ios versions!

jsilvestri:
@mweststrate I'm so sorry.  I posted this in the wrong place.  I found a similar bug in ios9 -- not related to mobxjs!

Issue #375, opened 2016-06-29
Different performance for computed value written in different ways.


kirillsalikhov:
I try to make a simple app which filters json file with ~3000 records. I have slow performance(~2-3s) if I use filter function written with _.reduce and closures inside @computed. And everything works fast if I write it in simpler form(~40ms).

json file looks like:

```
[
  {
   "ID": "3450",
    "Corpus": "1",
    "Section": "1",
    "Floor": 25,
    "FlatNumber": 282, 
  },
]
```

my model class:
```
import {observable, computed} from 'mobx';
import _ from 'lodash';

var startsWith = (filter, name) => (flat) => {
    if (!filter[name]) return true;

    return _.startsWith(flat[name] + '', filter[name]);
};

class FlatsModel {

    flats = [];
    
    @observable filter = {
        Corpus: '',
        Floor: '',
        Section: ''
    };

    filterFns = {
        'Corpus': startsWith(this.filter,'Corpus'),
        'Floor': startsWith(this.filter, 'Floor'),
        'Section': startsWith(this.filter, 'Section')
    };

    filterFn = (flat) => {
        return _.reduce(this.filterFns, function(result, fn) {
            return result && fn(flat);
        }, true);
    };

    @computed get filtered() {
       //SLOW
        return this.flats.filter(this.filterFn);
    }

    populate(flats) {
        this.flats = flats;
    }

    setFilter(name, value) {
        this.filter[name] = value;
    }
}

export default FlatsModel;
```

view is simple
```
import React from 'react';
import {observer} from 'mobx-react';

import Row from './row';

@observer
class List extends React.Component {    
    render() {
        var {model} = this.props;
        return (
            <table>
                <tbody>
                    {model.filtered.slice(0,5).map(flat => <Row flat={flat} key={flat.ID}/>)}
                </tbody>
            </table>
        )
    }
}

export default List;
```

If I rewrite get filtered() like below, everything works fine:
```
@computed get filtered() {
    var filter = this.filter;        
    return this.flats.filter(function(flat) {
        var val = true;
        if (filter['Corpus']) {
            val = val && _.startsWith(flat['Corpus'] + '', filter['Corpus'])
        }        
        if (filter['Floor']) {
            val = val && _.startsWith(flat['Floor'] + '', filter['Floor'])
        }        
        if (filter['Section']) {
            val = val && _.startsWith(flat['Section'] + '', filter['Section'])
        }
        return val;
    });
}
```

[...]

PR #422, opened 2016-07-19
Fast collections

mweststrate:
Addresses among others #375. Basically the `observer` collection is now a set, and the `observing` collection is now a lot more effectiently deduplicated.


src/core/derivation.ts
```diff
@@ -99,25 +111,58 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T) {
...
-		addObserver(added[i], derivation);
+	// This process is optimized by making sure deps are always left 'clean', with value 0, so that they don't need to be reset at the start of this process
+	// after that, all prevObserving items are marked with -1 directly, instead of 0 and doing -- after that
+	// further the +1 and addObserver can be done in one go.
+	for (let i = 0; i < prevLength; i++)
+		prevObserving[i].diffValue--; // expected 0 here, but -1 disables next loop:
+
+	for (let i = 0; i < newLength; i++) {
+		const dep = observing[i];
+		if ((++dep.diffValue) > 0) {
```
mweststrate: double check different value for non unique deps

PR #423, opened 2016-07-19
use created descriptor for defineProperty

bb:
If descriptor was falsy (line 33), a newDescriptor was created (line 35) but it was not actually used in line 55.

This broke Angular 2 zone.js for me with TypeError: desc is undefined and the proposed change fixes that.


hccampos:
Same here. I was actually going to submit a fix for this too. Hopefully it will get merged and published soon.

mweststrate:
Thanks for the PR! I think this is actually a regression bug indeed, If I glance at it quickly seems an incorrect fix of a shadowing issue

bb:
Thanks for quickly merging. However if the real bug is deeper, we should try fixing that as well. I just did the minimal fix which solved my concrete problem.

PR #438, opened 2016-07-24
Pass Reaction as an argument

andykog:
@mweststrate, proposed this in https://github.com/mobxjs/mobx/pull/367/files#diff-b7cd2bd0ae3078ef83015cbfa3499bf6

Reaction instance is being passed to
 - mobx.reaction's expression and effect functions
 - mobx.autorun's view
 - mobx.autorunAsync's view

So you can do:
```js
autorun(r => {
   if (condition) r.dispose();
});

reaction(
  r => {
    if (condition) r.dispose();
    return something.get(); 
  },
  v => {}
)

reaction(
  () => something.get(),
  (v, r) => {
    if (condition) r.dispose();
  }
)
```

Will add this to docs in another PR.

### DIFF
--- a/src/api/autorun.ts
+++ b/src/api/autorun.ts
@@ -1,6 +1,6 @@
 import {Lambda, getNextId, deprecated, invariant, valueDidChange} from "../utils/utils";
 import {assertUnwrapped, ValueMode, getValueModeFromValue} from "../types/modifiers";
-import {Reaction} from "../core/reaction";
+import {Reaction, IReactionPublic} from "../core/reaction";
 import {untrackedStart, untrackedEnd} from "../core/derivation";
 import {action} from "../core/action";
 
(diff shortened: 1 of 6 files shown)


andykog:
I was thing about passing plain object of shape`{ dispose: () => void }` instead of actual Reaction to prevent exposing private APIs. Seems, that meteor's Tracker.autorun is ok with that…

mweststrate:
I'm fine with passing the whole reaction object, but typing it as just `{ dispose: () => void }`. Makes the api contract explicit and people still have an escape hatch if they need it for whatever reason. (Could be fun for debugging actually as you could see the `.observing` property grow while stepping through the autorun function)

andykog:
@mweststrate, you are saying about TypeScript's typings? Added https://github.com/mobxjs/mobx/pull/438/commits/a6f697a0d7ce52afa4befaab6830ff7b22675fb9

mweststrate:
Exactly that indeed :)

Op zo 24 jul. 2016 om 15:26 schreef Andy Kogut <notifications@github.com>:

mweststrate:
Thanks again and released as `2.4.1`

Issue #445, opened 2016-07-29
MobX is obfuscating exceptions


tomaash:
When my render function throws an exception (for example because I'm trying to access property of null object, which is quite common in render functions), I now get an unhelpful MobX message in console instead of the stack trace:

```
[mobx] An uncaught exception occurred while calculating your computed value, autorun or transformer. Or inside the render() method of an observer based React component. These functions should never throw exceptions as MobX will not always be able to recover from them. Please fix the error reported after this message or enable 'Pause on (caught) exceptions' in your debugger to find the root cause. In: 'JobFormComponent#.render()'
```

now, I'm able to find the root cause with debugger, but it's a nuisance. Is there a reason why MobX cannot just rethrow the exception? Those happen quite often during development, and are often a 10 second fix. Now, when I need to turn on debugger "just at the right moment" to catch the correct expection, it takes much more time and it's very frustrating. I'm whinning here because I love MobX, and I want it to be perfect :) I believe this could be a very simple fix on the MobX side, or I'm willing to help if it's not.

tomaash:
I looked into the code.. I understand that some stuff is going on there with `derivations` that might prevent us from rethrowing the exception, but when we're already setting `hasException = true` somewhere, we could also set some exception data at least, to make the ^ warning message more helpful?

andykog:
@tomaash mobx should rethrow exeptions. Are you using fresh mobx version? Is it possible that you didn't notice actual error in the console? Can you show minimal example if the error is really being swallowed in some cases

Duplicate: https://github.com/mobxjs/mobx-react/issues/83

andykog:
See this example, error is being thowed https://jsfiddle.net/Sl1v3r/8xzbts9h/

tomaash:
You are right, I cannot reproduce it. I forked it to add inject and
provider which I use, but it still works:
https://jsfiddle.net/hsyr8tzt/
Last think I did not consider is typescript.. (my project is in typescript)
I'll investigate further..

andykog:
Closed in favour of https://github.com/mobxjs/mobx-react/issues/83

PR #447, opened 2016-07-30
Include original error in console.warn

andykog:
Some people (#445, https://github.com/mobxjs/mobx-react/issues/83), are suffering in situation, when error that happen during rendering being caught unintentionally. For example:
```js
const v = observable('Hello');

@observer
class extends Component {
  render() {
    if (v.get() !== 'Hello') {
      throw new Error('Original error');
    }
    return <h1>{v.get()}</h1>;
  }
}

window.fetch('/data.json')
  .then(data => v.set(data))
  .catch(e => {
    // Oh, may be that's http error, I don't care.
  });
```

With this PR original error (with stack trace) can at least be visible.

### DIFF
--- a/src/core/derivation.ts
+++ b/src/core/derivation.ts
@@ -91,34 +91,31 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T) {
 	let result: T;
 	try {
 		result = f.call(derivation);
-		hasException = false;
-	} finally {
-		if (hasException) {
-			const message = (
-				`[mobx] An uncaught exception occurred while calculating your computed value, autorun or transformer. Or inside the render() method of an observer based React component. ` +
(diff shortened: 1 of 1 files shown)


andykog:
@mweststrate pointed out that using catch and retrow will prevent debugger from stopping in place where original error was thrown, so I don't think this is is good idea anymore.

tomaash:
@andykog You are right, if I include `console.log(error.stack)` in every API call `.catch` clause, then I don't miss the errors. I completely forgot that the offending catch actually happens inside my code :) Well, at least we've found the root cause, right?

andykog:
@tomaash, Cool. But I would use `console.error(error)` as it provides error message in addition to stack.

tomaash:
True. You are right, as always :p

PR #450, opened 2016-07-31
Improve error msg when wrongly using `@computed`

bb:
When using `@computed` on a property, `originalDescriptor` is undefined thus the access to `get` failed before the invariant was reached.

I just (stupidly) wrote a `@computed` decorator in front of a property like this:

    @computed urls: string[] = [];

The error message I saw in browser console was

    mobx.js:180 Uncaught TypeError: Cannot read property 'get' of undefined

With this change a nice error message will be shown instead.

### DIFF
--- a/src/api/computeddecorator.ts
+++ b/src/api/computeddecorator.ts
@@ -11,6 +11,7 @@ export interface IComputedValueOptions {
 
 const computedDecorator = createClassPropertyDecorator(
 	(target, name, _, decoratorArgs, originalDescriptor) => {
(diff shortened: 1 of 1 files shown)


bb:
This might be related to #423 which also addresses the descriptor being undefined in a different use case.

The stacktrace leading to this specific call includes https://github.com/mobxjs/mobx/blob/master/src/utils/decorators.ts#L40 which uses descriptor even though it is undefined/falsy as checked in L33. Another usage which might need fixing is L45.
Maybe a better fix than this PR could be done there.

mweststrate:
Nice addition, thanks!

mweststrate:
Releases as 2.4.2

bb:
Thanks!

PR #452, opened 2016-08-01
Improve memoizing of computed values

asterius1:
This pull requests rewrites bunch of mobx core to 'change direction' of optimally recalculating changes.

Previously it was propagating from changed observables to observing derivations.

In proposed approach propagated up is only fact that some deep dependency might have changed (and reactions scheduled from it). Derivations before computing itself can check if any dependencies indeed changed and recompute when appropriate.
It introduces few big wins:
* When modifications and computations are mixed mobx can still rerun only needed stuff
* When in `autorun(() => a ? b : c)` `a` becomes true `c` is not being recomputed
* Batching of memoization can be introduced. For example `batch(() => { ... })` would make computed memoized for it's duration, giving (as far as I can tell) minimal possible number of recomputations regardless of modifications or reactions.

One issue I had is autoruns rerunning itself. Even though all tests for it are ok, I could point out some corner cases in which behaviour slightly change. In general self-rerunning autoruns, even outside this pull request, have many undefined/unexpected behaviours, although those cases can be quite artificial. I was thinking if we could add a rule so changes in autorun can't schedule it for rerun.

`computed temporary memoization` perf test previews the case, that made me do this in the first place.
Currently it takes 7 seconds. With changes it's down to ~1ms.
Most perf tests are very similar. Few perform worse.

The nicest thing about it is being able to not compromise clear code with performance even while implementing complex dependencies with possibly expensive actions.

Adjusted few tests, mainly lowering computation count.

Right now pull request doesn't pass the test for whyRun. Need to adjust language to new changes.


mweststrate:
So far this looks pretty amazing 😊. I will give detailed feedback and discuss stuff later. this might take a while but I'll come back on this ☺

src/core/computedvalue.ts
```diff
@@ -20,18 +19,21 @@ export interface IComputedValue<T> {
...
 	isComputing = false;
-	staleObservers: IDerivation[] = [];
-	observers = new SimpleSet<IDerivation>();      // nodes that are dependent on this node. Will be notified when our state change
+
+	dependenciesState = -1;
 	observing = [];       // nodes we are looking at. Our value depends on these nodes
+	newObserving = null; // during tracking it's array with new observed observers
+
+	isPendingUnobservation: boolean; // for effective unobserving
+	isObserved = false;
```
mweststrate: the following invariant if I'm correct? `isObserved === dependenciesState !== -1`? That might indicate that `isObserved` might not be explicitly needed?

src/core/derivation.ts
```diff
@@ -1,20 +1,25 @@
...
 	observing: IObservable[];
-	staleObservers: IDerivation[];
-	observers: SimpleSet<IDerivation>;
-	dependencyStaleCount: number;
-	dependencyChangeCount: number;
-	onDependenciesReady(): boolean;
+	newObserving: IObservable[];
+	/**
+	 * Describes state of observing dependencies to know whet it's needed to rerun
+	 * -1 <- not tracking dependencies
```
mweststrate: using an enum would make this cleared: `enum IDerivationState { NOT_TRACKING = -1, UP_TO_DATE = 0, POSSIBLY_STALE = 1, STALE = 2 }`

src/core/derivation.ts
```diff
@@ -24,6 +29,30 @@ export interface IDerivation extends IDepTreeNode, IObservable, ISetEntry {
 	 * amount of dependencies used by the derivation in this run, which has not been bound yet.
 	 */
 	unboundDepsCount: number;
+	onBecomeStale();
+	recoverFromError();
+}
+
+export function shouldCompute(derivation: IDerivation, onError?): boolean {
+	const dependenciesState = derivation.dependenciesState;
+	if (dependenciesState === 0) return false;
```
mweststrate: might benifit from a switch statement

src/core/derivation.ts
```diff
@@ -24,6 +29,30 @@ export interface IDerivation extends IDepTreeNode, IObservable, ISetEntry {
...
+	if (dependenciesState === -1 || dependenciesState === 2) return true;
+	// derivation.dependencyChange === 1
+	let hasError = true;
+	try {
+		for (let i = 0; i < derivation.observing.length; i++) {
+			const obj = derivation.observing[i];
+			if (obj instanceof ComputedValue) {
+				obj.get();
```
mweststrate: I see what this does, but calling `get()` and then relying on the side effects of `derivation.dependenciesState` might justify some elaborate comments :)

src/core/derivation.ts
```diff
@@ -24,6 +29,30 @@ export interface IDerivation extends IDepTreeNode, IObservable, ISetEntry {
...
+				if (derivation.dependenciesState === 2) return true;
```
mweststrate: the following invariant holds here, at this point the dependenciesState is either `2` or `0`?

src/core/derivation.ts
```diff
@@ -185,4 +197,58 @@ export function untrackedStart() {
...
+	const oldState = derivation.dependenciesState;
+	if (oldState === targetState) return;
+	derivation.dependenciesState = targetState;
+	const obs = derivation.observing;
+	const l = obs.length;
+
+	const m = derivation.__mapid;
+
+	// just because it can be bottleneck
+	if (oldState === 0) {
```
mweststrate: this code could benefit from some de-dup. the suggested `move` seems eligable indeed.

src/core/derivation.ts
```diff
@@ -185,4 +197,58 @@ export function untrackedStart() {
...
+			}
+	}
+	else { // oldState === 2
+		if (targetState === 0)
+			for (let i = 0; i < l; i++) {
+				const o = obs[i];
+				delete o.observers.data2[m];
+				o.observers.data0[m] = derivation;
+			}
+		else // targetState === 1
```
mweststrate: does this ever happen?

src/core/derivation.ts
```diff
@@ -136,37 +148,37 @@ function bindDependencies(derivation: IDerivation, prevObserving: IObservable[])
...
+		}
+	}
 }
 
 export function clearObserving(derivation: IDerivation) {
 	const obs = derivation.observing;
 	const l = obs.length;
 	for (let i = 0; i < l; i++)
 		removeObserver(obs[i], derivation);
+	derivation.dependenciesState = -1;
```
mweststrate: might be clearer to have changeDependencieState(-1) and clear the observers there?

src/core/derivation.ts
```diff
@@ -185,4 +197,58 @@ export function untrackedStart() {
...
+				delete o.observers.data0[m];
+				o.observers.data2[m] = derivation;
+			}
+		else // targetState === 1
+			for (let i = 0; i < l; i++) {
+				const o = obs[i];
+				delete o.observers.data0[m];
+				o.observers.data1[m] = derivation;
+			}
+	}
```
mweststrate: conceptually could also transition to -1? should be handled here?

src/core/observable.ts
```diff
@@ -1,10 +1,58 @@
-import {IDerivation, notifyDependencyReady, notifyDependencyStale} from "./derivation";
+import {IDerivation, changeDependenciesState} from "./derivation";
 import {globalState} from "./globalstate";
-import {SimpleSet} from "../utils/set";
+
+export class DerivationsSets {
```
mweststrate: this obsoletes SimpleSet, so that one could be removed

src/core/observable.ts
```diff
@@ -1,10 +1,58 @@
...
+	size: number = 0;
+	data0 = {};
```
mweststrate: I wouldn't be surprised, it is interesting to test that, that keeping one collection for all states and then filtering wherever the sets are used is faster? because `delete` is quite expensive and usually these sets are pretty constant over the lifetime of a derivation, so that would a lot of moving around.

src/core/observable.ts
```diff
@@ -1,10 +1,58 @@
...
+	}
+	remove(value: IDerivation) {
+		const data = this["data" + value.dependenciesState];
+		const m = value.__mapid;
+		if (m in data) {
+			delete data[m];
+			this.size--;
+		}
+	}
+	// move(targetState: number, value: IDerivation) {
```
mweststrate: +1 :) although if all is gathered in one set, the operation isn't needed at all. But not sure whether that is faster or terribly slow, let's investigate

src/core/globalstate.ts
```diff
@@ -48,12 +48,9 @@ export class MobXGlobals {
 	 */
 	isRunningReactions = false;
 
-	/**
-	 * List of observables that have changed in a transaction.
-	 * After completing the transaction(s) these atoms will notify their observers.
-	 */
-	changedAtoms: IAtom[] = [];
+	inBatch: number = 0;
```
mweststrate: note that the version attribute needs to be bumped, as the globalState object's signature changes, meaning that this mobx version is not compatible with older mobx version in same memory

src/core/observable.ts
```diff
@@ -16,46 +64,115 @@ export interface IObservable extends IDepTreeNode {
...
+	isObserved: boolean; // for unobserving only once ber observation
+	// sets of observers grouped their state to only notify about changes.
+	observers: DerivationsSets;
+
 	onBecomeUnobserved();
 }
 
 export function addObserver(observable: IObservable, node: IDerivation) {
 	observable.observers.add(node);
+	if (!observable.isObserved) {
```
mweststrate: just `observable.isObserved =  true` ?

src/core/observable.ts
```diff
@@ -16,46 +64,115 @@ export interface IObservable extends IDepTreeNode {
...
+	globalState.inBatch++;
+}
+
+export function endBatch() {
+	if (--globalState.inBatch === 0) {
+		globalState.pendingUnobservations.splice(0).forEach(observable => {
+			observable.isPendingUnobservation = false;
+			if (observable.isObserved && observable.observers.length === 0) {
+				observable.isObserved = false;
+				observable.onBecomeUnobserved(); // TODO: test if this happens only once, e.g. remove returns bool!
```
mweststrate: double check, this triggers clearObserving for all cases and results in state -1

src/core/observable.ts
```diff
@@ -16,46 +64,115 @@ export interface IObservable extends IDepTreeNode {
...
+		}
 		return;
+	}
 	const derivation = globalState.derivationStack[globalState.derivationStack.length - 1];
 	/**
 	 * Simple optimization, give each derivation run an unique id (runId)
 	 * Check if last time this observable was accessed the same runId is used
 	 * if this is the case, the relation is already known
 	 */
+	observable.isObserved = true;
```
mweststrate: if i'm not mistaken, this can move inside the if. btw if that is the case, `addObserver` might not need to set `isObserved` as well?

src/core/reaction.ts
```diff
@@ -112,6 +104,11 @@ export class Reaction implements IDerivation, IReactionPublic {
 		}
 	}
 
+	recoverFromError() {
+		this._isRunning = false;
+		this._isTrackPending = false;
+	}
+
 	dispose() {
 		if (!this.isDisposed) {
```
mweststrate: reminder, add TODO on `whyRun` update

test/autorunAsync.js
```diff
@@ -6,16 +6,21 @@ test('autorun 1', function(t) {
 	var _result = null;
 	var _cCalcs = 0;
 	var to = setTimeout;
-	
+
 	function expect(fired, cCalcs, result) {
 		t.equal(_fired, fired, "autorun fired");
 		t.equal(_cCalcs, cCalcs, "'c' fired");
+		if(_cCalcs !== cCalcs) {
+			console.log('\n')
```
mweststrate: old debug statement?

test/observables.js
```diff
@@ -1661,7 +1661,7 @@ test('unoptimizable subscriptions are diffed correctly', t => {
 	t.equal(a.observers.length, 2)
 	t.equal(b.observers.length, 1)
 	t.equal(c.observers.length, 1)
-	t.equal(d.$mobx.observing.length, 4) // 3 would be better!
+	t.equal(d.$mobx.observing.length, 3) // 3 would be better!
```
mweststrate: :+1:

test/observables.js
```diff
@@ -1545,23 +1545,23 @@ test('forcefully tracked reaction should still yield valid results', function(t)
     });
 
     t.equal(z, 4);
-    t.equal(runCount, 3);
+    t.equal(runCount, 2); // x is observed, so it should recompute only on dependency change
```
mweststrate: :+1

mweststrate:
Left a bunch of comments and notes to self for now, will look later a bit more in depth to a some individual tests and memory and cpu performance in general

src/core/computedvalue.ts
asterius1: It is the case only for ComputedValue.
Atom doesn't implement derivations, so they don't have dependenciesState
Reaction doesn't implement observable so they don't have isObserved.

isObserved is there to make sure that clearing up batch doesn't flip out.
It's kinda safer with.
I am not sure if it's 100% necessary.

src/core/derivation.ts
asterius1: yep, gonna comment it.
`state === 1` is possible only on `ComputedValue`
at the end of `shouldCompute` `derivation.dependenciesState !== 1`
`ComputedValue.get` will always call shouldCompute

src/core/derivation.ts
asterius1: I'm right now trying to change whole `DerivationsSets` to something more performant, because it seems to be a performance bottleneck.

src/core/observable.ts
asterius1: yep, it does for ComputedValues. Not for Atoms.

asterius1:
Thanks.
I am now rewriting DerivationsSets for better performance, because it's indeed a performance bottleneck.

asterius1:
Made storing observers simpler and way more performant in common cases.
Also bumped globalState version and added some comments.

For performance it seems to be quite faster when data structure for observers is inlined into observable (without proxy set class). I added LegacyObservers, because some tests depended on it.
Probably should clear it up if possible. Is it likely that something depended on it?
Also can I just remove SimpleSet or some outside packagde might depend on it?

Or just depreciate?

src/core/computedvalue.ts
```diff
@@ -32,8 +32,9 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
 	diffValue = 0;
 	runId = 0;
 	lastAccessedBy = 0;
+	lowestObserverState = 0;
 	unboundDepsCount = 0;
-	__mapid = "#" + getNextId();
+	__mapid = "#" + getNextId(); // NOT NEEDED ANYMORE
```
mweststrate: feel free to remove

src/core/observable.ts
```diff
@@ -65,6 +74,7 @@ export interface IObservable extends IDepTreeNode {
 	 */
 	lastAccessedBy: number;
 
+	lowestObserverState: number;
```
mweststrate: deserves some documentation :)

src/core/reaction.ts
```diff
@@ -49,7 +51,9 @@ export class Reaction implements IDerivation, IReactionPublic {
 		if (!this._isScheduled) {
 			this._isScheduled = true;
 			globalState.pendingReactions.push(this);
+			startBatch();
```
mweststrate: push inside reactions?

src/core/computedvalue.ts
mweststrate: never mind :)

mweststrate:
I might be using github wrong, but closing this PR in favor of #489 where further review comments can be processed.

PR #473, opened 2016-08-10
Concider object with null prototype a plain object

andykog:
`isPlainObject(Object.create(null))` should return true


mweststrate:
Nice catch!

mavericken:
I actually just needed this today.  Thanks!

mweststrate:
Released as `2.4.3`

PR #475, opened 2016-08-11
Refactor max iterations exception message

igncp:
Hi,

Updated the exception message for cyclic reactions. It is a matter of taste, so just wanted to check your opinion. I hope it helps.

Cheers


mweststrate:
@igncp yep that is better! Thanks for the contribution :)

Issue #483, opened 2016-08-12
Tell TypeScript API consumers when their objects are observable


spiffytech:
Presently, calling `mobx.observable({})` returns a `T`, and there's no way for my downstream code to assert that it's passed an observable object. Knowing when an object is observable is desirable for e.g., guaranteeing my React components receive objects that will actually mutate my UI when I change them.

My PR uses TypeScript 1.4's union types to change the return type of `mobx.observable()` to `T & IObservable`. This should be fully backwards-compatible, since TypeScript will still treat the object as type `T`. But now consumers have the option of enforcing that various pieces of their apps receive observable objects by demanding a `mobx.IObservable`.

mweststrate:
Merged PR, will be part of 2.5.2

PR #484, opened 2016-08-12
feature(api): tell consumers when their objects are observable

spiffytech:
See issue #483.

### DIFF
--- a/src/api/observable.ts
+++ b/src/api/observable.ts
@@ -4,6 +4,7 @@ import {computed} from "./computeddecorator";
 import {isPlainObject, invariant, deprecated} from "../utils/utils";
 import {observableDecorator} from "./observabledecorator";
 import {isObservable} from "./isobservable";
+import {IObservableObject} from "../types/observableobject";
 import {IObservableArray, ObservableArray} from "../types/observablearray";
 
 /**
@@ -15,7 +16,7 @@ export function observable(target: Object, key: string, baseDescriptor?: Propert
 export function observable<T>(value: T[]): IObservableArray<T>;
 export function observable<T, S extends Object>(value: () => T, thisArg?: S): IObservableValue<T>;
 export function observable<T extends string|number|boolean|Date|RegExp|Function|void>(value: T): IObservableValue<T>;
-export function observable<T extends Object>(value: T): T;
+export function observable<T extends Object>(value: T): T & IObservableObject;
 export function observable(v: any = undefined, keyOrScope?: string | any) {
 	if (typeof arguments[1] === "string")
 		return observableDecorator.apply(null, arguments);
--- a/src/mobx.ts
+++ b/src/mobx.ts
@@ -31,7 +31,7 @@ export { IComputedValue                                       } from "./core/com
 export { asReference, asFlat, asStructure, asMap              } from "./types/modifiers";
 export { IInterceptable, IInterceptor                         } from "./types/intercept-utils";
 export { IListenable                                          } from "./types/listen-utils";
-export { IObjectWillChange, IObjectChange, isObservableObject } from "./types/observableobject";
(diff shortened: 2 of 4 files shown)


mweststrate:
Thanks for the PR, will consider this for the next release

mweststrate:
This typing is actually not correct, an object being converted to an observable is not an `IObservable` itself, it has an IObservable backing object instead, as defined here: https://github.com/mobxjs/mobx/blob/aa8ea15573cf27e0128235b334ca53c9aaceeae1/src/types/observableobject.ts#L51. I don't feel that exposing that interface however adds value, it is just the internal administration.

What could be done is introducing an empty `IObservableObject` interface. That has in itself not that much value, except that it could be used as typeguard for `isObservableObject`, so that people can express that they want to have some observable object as arguments to their functions. see also: https://github.com/mobxjs/mobx/pull/487

spiffytech:
Having `observable({})` return `T & IObservableObject`, with `IObservableObject` being an empty interface would satisfy my request.

Just to be clear, I'm not looking for TypeScript's type guards - they're designed to enhance type safety of code branches used in runtime type checks. I'm looking to improve compile-time checks, like so:

```typescript
@action
function updateMyFoo(foo: MyInterface & IObservableObject) {
  foo.some_prop = true;  // triggers observers
}

const foo: MyInterface = {some_prop: false};
updateMyFoo(foo);  // compiler error - not observable

const observable_foo = observable(foo);  // MyInterfyace & IObservableObject
updateMyFoo(observable_foo);  // success
```

mweststrate:
Yep, that makes sense to me :)

Strate:
@spiffytech unfirtunatelly, due to structural typing system limitation (which TYpeScript is based on), you will not receive compile time errors here, because TS will not check *interfaces* on object, it will check *structure* of object. It means, that having empty interface is almost useless.
There is a suggestion to improve TS in this part, but still needs proposal: https://github.com/Microsoft/TypeScript/issues/202

mweststrate:
dôh indeed

Strate:
There is a cool trick, I've found in comments: https://github.com/Microsoft/TypeScript/issues/202#issuecomment-50199713
Maybe we could use it here :)

spiffytech:
@Strate Can you clarify how your suggestion works with empty interfaces? The below code sample compiles for me, when it should produce an error - `updateMyFoo()` is still allowing an argument of type `MyInterface`.

```typescript
interface IObservableObject {}

interface MyInterface {
    some_prop: boolean;
}

function updateMyFoo(foo: MyInterface & IObservableObject) {
    foo.some_prop = true;
}

const foo: MyInterface = {some_prop: false};
const observable_foo = <MyInterface & IObservableObject> foo;

updateMyFoo(foo);  // compiles :(

updateMyFoo(observable_foo);  // compiles
```

spiffytech:
@mweststrate May I assume from your previous comment that you wouldn't be comfortable exposing the `IIsObservableObject` interface (on an identical interface named `IObservableObject`), even if it's the only way to satisfy this feature request?

mweststrate:
@spiffytech indeed, the `mobx$` is an internal concept and should be hidden to the outside world as much as possible. 

something like `interface IObservableObject { "observable-object-marker": IObservableObject }` is the trick I think. That attribute doesn't have to be exists at runtime, it is just that the compiler needs to think that it is returned from the `observable` function :)

spiffytech:
Ah, I see! Confirmed, that makes the appropriate error occur in my code sample.

I'll update my full request with this.

mweststrate:
@spiffytech did you by any chance proceed on this one ?

spiffytech:
Yeah, I adopted what we discussed, added the type guard to `isObservable`, and added some unit tests for both.

I was hoping to find a way to unit test the failure case for the type guard, but everything I could think of prevented the TypeScript tests from compiling.

mweststrate:
Just merged, thanks a lot!

PR #487, opened 2016-08-16
Make `isObservableArray` and `isObservableMap` type guards.

Strate:
Typescript has a cool feature called `user type guard functions`, which helps to
narrow type of variable. This PR makes `isObservableArray` and `isObservableMap`
user type guard functions.


mweststrate:
cool!

PR #491, opened 2016-08-18
Correct fetchedTodos.forEach parameter usage in docs

gskema:
I removed the timestamp changes in the `.html` since I didn't want to spam the PR.  `search_index.json` got changed as well so I included that.


capaj:
thanks @gskema! LGTM

PR #499, opened 2016-08-22
Typescript: Add proper return type of `runInAction` function.

Strate:
The return type of `runInAction` actually is return type of wrapped function.


mweststrate:
Correct, thanks!

PR #500, opened 2016-08-22
Typescript: Add granular overloads for `action`, which fixes "Parameter XXX implicitly has an 'any' type"

Strate:
Trying to use action as a function wrapper on promises, like this:
```ts
import {action} from "mobx"

fetch().then(action(result => {

}))
```
Produces "Parameter 'result' implicitly has an 'any' type", if compiled with `noImplicitAny` option.
This PR adds some overloads for `action` declaration, which fixes this issue.

### DIFF
--- a/src/api/action.ts
+++ b/src/api/action.ts
@@ -18,6 +18,14 @@ const actionFieldDecorator = createClassPropertyDecorator(
 	true
 );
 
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks! really useful

PR #511, opened 2016-08-24
Remove semvar lock from travis

amilajack:


mweststrate:
Seems sane. Just curious, has this any particular advantage?

amilajack:
Its just for reasons of best practice.

PR #545, opened 2016-09-09
Fix broken markdown for @inject documentation.

osmanzeki:
Due to a missing char, the `@inject` doc at line 135 was not rendering properly in markdown. I let it be a few times and figured someone else might create a PR for it but I might as well do it. :)

Cheers!


mweststrate:
Thanks!

PR #553, opened 2016-09-12
add debug and scope info

bb:
Added a tiny bit of documentation for autorunAsync.

BTW, is there a guideline regarding documentation improvements besides the short contribution note in the Readme file?

E.g. `scope` works same on many methods and does not need repeated explanation everywhere, however it is not mentioned anywhere in the docs yet.

Should the different method signatures be added at the top of each page? Or should the alternatives be mentioned below (as in this PR) or not at all?


mweststrate:
Thanks! There is not really a standard around the docs yet. It grew quite organically, although it would be awesome if somebody could find the time to organize things better :) The api overview is more consistent in showing signatures, as that was written in one go, but some detail pages are very old, and some new.

PR #559, opened 2016-09-16
make autorun available as decorator

bb:
Inspired by the discussion on https://mobxjs.github.io/mobx/refguide/autorun.html / http://disq.us/p/1bthagx why autorun is not usable as decorator, I thought I'd give it a try.

This is by no means complete yet, only decorator without custom name is supported. No docs yet, either. If someone wants to complete that feature: go ahead.

I'm looking for feedback if this approach is going in the right direction or what should be changed/improved before proceeding.

### DIFF
--- a/src/api/autorun.ts
+++ b/src/api/autorun.ts
@@ -23,9 +23,13 @@ export function autorun(view: (r: IReactionPublic) => void, scope?: any);
  */
 export function autorun(name: string, view: (r: IReactionPublic) => void, scope?: any);
 
-export function autorun(arg1: any, arg2: any, arg3?: any) {
+export function autorun(customName: string): (target: Object, key: string, baseDescriptor?: PropertyDescriptor) => void;
+export function autorun(target: Object, propertyKey: string, descriptor?: PropertyDescriptor): void;
+
+
+export function autorun(arg1: any, arg2?: any, arg3?: any) {
 	let name: string, view: (r: IReactionPublic) => void, scope: any;
-	if (typeof arg1 === "string") {
+	if (typeof arg1 === "string" && typeof arg2 !== "undefined") {
 		name = arg1;
 		view = arg2;
 		scope = arg3;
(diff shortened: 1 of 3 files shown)


test/typescript-tests.ts
```diff
@@ -962,3 +962,43 @@ test('computed getter / setter for plan objects should succeed (typescript)', fu
...
+
+	const a = new SomeClass();
+	t.equal(autorunResult, 1);
+	a.foo = 2;
+	t.equal(autorunResult, 2);
+	t.equal(noAutoResult, 0);
+	t.end();
+});
+
+test.skip('autorun as named decorator', function(t) {
```
mweststrate: why is this one skipped?

test/typescript-tests.ts
```diff
@@ -962,3 +962,43 @@ test('computed getter / setter for plan objects should succeed (typescript)', fu
 
 	t.end();
 });
+
+test('autorun as (unnamed) decorator', function(t) {
```
mweststrate: can you copy this test to the babel tests as well?

tsconfig.json
```diff
@@ -6,6 +6,7 @@
         "sourceMap": false,
         "declaration": true,
         "module": "commonjs",
+        "experimentalDecorators": true,
```
mweststrate: note that the tests have their own `tsconfig`, so this one shouldn't need to change

src/api/autorun.ts
```diff
@@ -23,16 +23,28 @@ export function autorun(view: (r: IReactionPublic) => void, scope?: any);
...
 		view = arg2;
 		scope = arg3;
 	} else if (typeof arg1 === "function") {
 		name = arg1.name || ("Autorun@" + getNextId());
 		view = arg1;
 		scope = arg2;
+	} else if (typeof arg1 === "object" && typeof arg2 === "string" && "value" in arg3) { // unnamed decorator
+		name = arg2;
+		view = arg3.value;
+		scope = arg1;
```
mweststrate: I think the current implementation creates one autorun for the entire class, instead of per instance which is probably what one would intend? The `decorator-utils` might come in handy to 'lift' the behavior from the class to specific instance.

mweststrate:
@bb closing the PR for now, since it doesn't seem to be very active anymore ;-).

PR #565, opened 2016-09-19
Fix code example on modifiers docs

leader22:
Some example code became an error when running on browser.


andykog:
@leader22, thanks! Added one more fix: https://github.com/mobxjs/mobx/commit/008395f48c6e4861c752f637a10568650d5c4ea1

PR #568, opened 2016-09-19
Prevent RangeError when calling toJS on HTMLElement

### LINKED ISSUE #566: RangeError when passing an observable HTMLElement to toJS
When you pass an observable `HTMLElement` to `toJS`, a RangeError occurs:

```
mobx.js:427 Uncaught (in promise) RangeError: Maximum call stack size exceeded(…)
```

maxdeviant:
This fixes #566.

I plugged this fix into the compiled MobX lib in our codebase and verified that it fixed the problem we were seeing there.

Once I'm back on my OS X machine I can run the browser tests to ensure that the added test breaks without the fix. Sadly can't run the `small-build.sh` script on Windows :cry:


maxdeviant:
@andykog Here is what I have so far. Let me know if anything needs adjusting.

src/api/tojs.ts
```diff
@@ -58,4 +58,20 @@ export function toJS(source, detectCycles: boolean = true, __alreadySeen: [any,
...
+}
+
+function isDOMNode(source) {
+	const isDOMAvailable = (
+		typeof document === "object" &&
+		document &&
+		document.createElement &&
+		typeof HTMLElement === "function" &&
+		typeof Node === "function"
+	);
```
andykog: All those checks are unnecessarily executed each time `isDOMNode` is called. In my variant (#566) they are called only once. Not a bid deal, but what for?

test/tojs.js
```diff
@@ -0,0 +1,20 @@
...
+	if (typeof document !== 'object') {
+		return t.end();
+	}
+
+	var image = observable(document.createElement('img'));
+
+	try {
+		toJS(image);
+	} catch (err) {
+		t.error(err);
```
andykog: There is `t.doesNotThrow` for this https://github.com/substack/tape#tdoesnotthrowfn-expected-msg

test/tojs.js
```diff
@@ -0,0 +1,20 @@
...
+	}
+
+	t.end();
```
andykog: I would also check if toJS(image) === image

test/tojs.js
```diff
@@ -0,0 +1,20 @@
+var test = require('tape');
```
andykog: I would place this test right after this one: https://github.com/maxdeviant/mobx/blob/d0404bb0e4d95c5a5ea5f888ee5952ed4b256f6e/test/observables.js#L1074

src/api/tojs.ts
maxdeviant: Ah yes, I misunderstood the scope of your example from the issue.

maxdeviant:
@andykog I made the requested edits.

mweststrate:
To be discussed see #568

mweststrate:
closed in favor of #589. Thanks for the effort nonetheless!

PR #582, opened 2016-09-27
Added warning when using autorun + autorunAsync with an action (#576)

### LINKED ISSUE #576: Add warning when passing an action to autorun
I spent 30 minutes today wondering why my `autorun` wasn't triggering. Turns out it was because I was doing this:

```
autorun(action(() => {}))
```

and actions are untracked (as I found out in #490). I think it would be a good idea to add a development-only warning when attempting to set up `autorun` with an action.

Implementing it shouldn't be too difficult - the function being returned from `action` could have a tag (a Symbol or just a privatized key) that is checked when creating an autorun.

Example warning:

```
Warning: attempted to pass an action to autorun. Actions are untracked and will not trigger on state changes. Use `reaction` or wrap only your state modification code in an action.
```

jeffijoe:
Tests do **not** run. If merged, closes #576


mweststrate:
Closed in favor of #587. Thanks!

PR #586, opened 2016-09-29
Fix 571 no instanceof

mweststrate:
eliminate `instanceof` checks by introducing on the prototype to identify the type. Note that different property names are used so that inheritance based checks work (actually only applicable for `ObservableValue` I think, so could be skipped)

@amir-arad / @andykog up for a review?

### DIFF
--- a/src/api/extendobservable.ts
+++ b/src/api/extendobservable.ts
@@ -1,5 +1,5 @@
 import {ValueMode} from "../types/modifiers";
-import {ObservableMap} from "../types/observablemap";
+import {isObservableMap} from "../types/observablemap";
 import {asObservableObject, setObservableObjectInstanceProperty} from "../types/observableobject";
 import {invariant, isPropertyConfigurable, hasOwnProperty} from "../utils/utils";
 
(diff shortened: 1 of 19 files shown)


amir-arad:
sorry for the late response.
change seems ok. you blew my mind with ```as any as T``` in the modifiers section 👍 

how about in ```createInstanceofPredicate``` instead of adding a flag to the class and checking for it,  do something like:
```return isObject(x) && x.constructor && x.constructor.name === clazz.name;``` ?

mweststrate:
That has a few risks: 1) the names are quite generic "Atom" etc, 2) the
names are likely to be uglified away, 3) it doesn't work for inheritance
(ObservableValue -> Atom is the only inheritance in the code base, but
still)

Op wo 5 okt. 2016 om 17:10 schreef Amir Arad <notifications@github.com>:

andykog:
Yep, the approach is ok, react does the same thing: https://github.com/facebook/react/blob/c78464f8ea9a5b00ec80252d20a71a1482210e57/src/isomorphic/modern/class/ReactComponent.js#L33

PR #590, opened 2016-09-29
make sure timer based tests pass in firefox as well

mweststrate:
can someone verify that the tests now succeed in safari as well?

`npm run test-browser-safari`

thanks!


yosiat:
Hi @mweststrate ,
I tried to run the tests on Safari 9.1, Node v6.7.0 and it looks like the test is stuck for ever, while in firefox it runs instantly.

Here is the test output:
```terminal
λ ~/projects/mobx/ fix-firefox-tests npm run test-browser-safari

> mobx@2.5.2 test-browser-safari /Users/yosi/projects/mobx
> npm run small-build && ( browserify test/*.js -t [ babelify --presets [ es2015 ] ] | tape-run --browser safari | faucet )


> mobx@2.5.2 small-build /Users/yosi/projects/mobx
> scripts/single-file-build.sh

WARN: Dropping unreachable code [lib/mobx.js:1505,4]
WARN: Declarations in unreachable code! [lib/mobx.js:1505,4]
WARN: Dropping unreachable code [lib/mobx.js:1618,8]
WARN: Declarations in unreachable code! [lib/mobx.js:1618,8]
WARN: Dropping unreachable code [lib/mobx.umd.js:1507,4]
WARN: Declarations in unreachable code! [lib/mobx.umd.js:1507,4]
WARN: Dropping unreachable code [lib/mobx.umd.js:1620,8]
WARN: Declarations in unreachable code! [lib/mobx.umd.js:1620,8]
WARN: Dropping unused function invariantObservers [lib/mobx.umd.js:1041,9]
WARN: Dropping unused function invariantLOS [lib/mobx.umd.js:1122,9]
WARN: Dropping unused variable define [lib/mobx.umd.js:1,335]
WARN: Dropping unused variable module [lib/mobx.umd.js:1,342]
WARN: Dropping unused variable exports [lib/mobx.umd.js:1,349]
[BABEL] Note: The code generator has deoptimised the styling of "/Users/yosi/projects/mobx/lib/mobx.js" as it exceeds the max of "100KB".
✓ action should wrap in transaction
✓ action modifications should be picked up 1
✓ action modifications should be picked up 1
✓ action modifications should be picked up 3
✓ test action should be untracked
✓ should be possible to create autorun in ation
✓ should not be possible to invoke action in a computed block
✓ action in autorun should be untracked
✓ action should not be converted to computed when using (extend)observable
✓ #286 exceptions in actions should not affect global state
✓ runInAction
✓ action in autorun does not keep / make computed values alive
✓ computed values and actions
✓ correct api should be exposed
✓ test1
✓ array should support iterall / iterable
✓ find and remove
✓ concat should automatically slice observable arrays, #260
✓ observe
✓ array modification1
✓ serialize
✓ array modification functions
✓ array modifications
✓ new fast array values won't be observable
✓ is array
✓ peek
✓ react to sort changes
✓ autoextend buffer length
✓ array exposes correct keys
✓ autorun passes Reaction as an argument to view function
✓ autorun can be disposed on first run
✓ autorun 1
✓ autorun should not result in loop
# autorunAsync passes Reaction as an argument to view function
  ok 10996 should be equal
```

If you could guide to how to run the specific test & debug it I will be happy to help.

mweststrate:
@yosiat it is a bit tricky, you can add a debugger statement to your tests, but you have to make sure you open the devtools quickly enough when the tests start (or at a delay in front) to be able to get at the breakpoint. Pinpointing which tests exactly fail would already be really useful (`autorunAsync` seems to fail from your output at least)!

yosiat:
@mweststrate is there a way to run only "autorunAsync passes Reaction as an argument to view function" so I can debug it with console.log?

mweststrate:
@yosiat yes, see: https://github.com/substack/tape#testonlyname-cb

yosiat:
@mweststrate when I run only this test, it finishes instantly, but when I ran it with the all tests it hangs forever.

I will try to debug it using "debugger"

yosiat:
@mweststrate I can't reproduce it for some reason.. but I found something interesting in the tape documentation:

> If you forget to t.plan() out how many assertions you are going to run and you don't call t.end() explicitly, your test will hang.

https://github.com/substack/tape

This might be the reason for those rare hangs,
If you I can submit a PR to add plan for this function.

mweststrate:
PR would be appreciated :) there is suppossed to be a t.end() everywhere
indeed

Op ma 3 okt. 2016 om 12:21 schreef Yosi Attias <notifications@github.com>:

yosiat:
@mweststrate I finally found what is the strange thing that happens here..
It started that sometimes "auto run 1" and sometimes "autorunAsync passes Reaction as an argument to view function" just waits forever.

I changed the autorunAsync test to logs when we enter setTimeout and when we finishes, and I found that only one/two setTimeout is called and not all timers are called.

When the tests are stuck I open safari dev tools and I see that the tests actually **finishes** 

tape_run_bug
_Left is safari devtool, right is terminal_

mweststrate:
Merged into `mobx3` branch, closing for now

PR #596, opened 2016-10-05
fix stack overflow in toJS when storing any dom element inside observable

bgnorlov:


mweststrate:
Hi @bgnorlov, see also #566 / #589? The latest release `5.6.0` already should have the desired behavior. Did you test it against that version?

bgnorlov:
@mweststrate I don't see it before. I quickly fixed it in my forked repo a week ago and currently remember it. Btw, I don't like the idea about recursive serialization only observables because I used toJS in my debug output and some object may be changes but console show latest current version in all outputs.

bgnorlov:
@mweststrate where is the 5.6.0 version? Current version 2.6.0 still have this error https://jsbin.com/xazonixiku/edit?html,js,console,output

capaj:
@bgnorlov I think that 5.6.0 was just a typo. Right @mweststrate ?

mweststrate:
yep, should be 2.6.0 indeed

Op di 1 nov. 2016 09:52 schreef Jiri Spac <notifications@github.com>:

mweststrate:
@bgnorlov sorry for the late follow up, but it is still unclear whether mobx 2.6.0 actually solved this issue for you? 

I understand that this does not fix your scenario's where you want clones of your entire state, not just the observable parts. But I think that is a generic problem outside the scope of MobX. For these kind of problems I think you should resort to libraries that are more general purpose, like lodash.clone.

bgnorlov:
@mweststrate, in my app I store some html elements directly in a state not in components because I need to move them up and down in html structure, otherwise, it cause refreshing them. For example, an audio element which user can drag and drop between hierarchical blocks and if this  audio element is playing then it will be refreshing. So I store audio element in my NodeStore class. And during any updates, my BaseClass logging any changes and uses toJS to logs internal state of NodeStore object and it cause stack overflow in toJS function.

mweststrate:
@bgnorlov ok, sorry for all the confusion. Something went wrong with the previous release and the earlier mentioned MR wasn't merged (well merged into the wrong branch ;-)). Sorry for all the confusion! I'll release a new version asap

mweststrate:
Released as 2.6.2, @bgnorlov could you verify this is the correct behavior?

bgnorlov:
@mweststrate, there is no longer stack overflow error that's fine, thank you! But now I found some other little issue (or maybe it's design decision) - toJS leaves all setters, getters, and $mobx fields which make my debug logs a little dirty. https://jsbin.com/xumejaqiwi/2/edit?html,js,console,output

mweststrate:
@bgnorlov can you elaborate on that? not sure what you mean (the fiddle seems old)

mweststrate:
@bgnorlov closing the PR for now, feel free to open a new issue if still having the issue you described above

PR #597, opened 2016-10-05
Keep custom props declared on arrays

antitoxic:
In my projects I also have objects that extend Array. Currently mobx ignores all custom props for arrays when it converts them to observable.

That is a very simple change at least to copy over concrete methods and props.

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -245,6 +245,10 @@ export class ObservableArray<T> extends StubArray {
 			// defined on the instance. After that it works fine, even if this property is deleted.
 			Object.defineProperty(adm.array, "0", ENTRY_0);
 		}
+        Object.getOwnPropertyNames(initialValues).forEach(prop => {
+            if (prop == 'length' || !isNaN(prop)) return;
+            this[prop] = initialValues[prop];
+        })
 	}
 
 	intercept<T>(handler: IInterceptor<IArrayChange<T> | IArraySplice<T>>): Lambda {


mweststrate:
But these props aren't observable. Wouldn't that be confusing? (Although
imho, props on arrays are confusing anyway ;-)). What is your use case for
it?

Op wo 5 okt. 2016 om 22:21 schreef Anton Stoychev <notifications@github.com

antitoxic:
> But these props aren't observable. 

I'm guessing you mean the custom props.They don't need to be observable. They're likely just extra methods and properties used with these methods.

> Wouldn't that be confusing?

Missing the properties all together is more confusing :) .

>  (Although imho, props on arrays are confusing anyway ;-)). 

Well :) , mobx does exactly that doesn't it? I also need it. It's a tiny change.

> What is your use case for it?

I have have a `Model` and `ModelCollection` classes. `ModelCollection` is my custom array class and it adds methods like `batchRefresh()` and `toGraphQl()`;

mweststrate:
> I'm guessing you mean the custom props.They don't need to be observable. 
Yep that ones. Well, in your case they don't ;-). But ones it is semantically allowed to preserve props through `observable` on arrays, are should also be a strategy about what that means, and your use case in this (non observable) might be different from someone elses

> Missing the properties all together is more confusing :) .
But at least semantically as clear & simple as it could be ;-).

>  I also need it. It's a tiny change.
That is not a very good reason to change behavior and support it, it is a generic lib ;-)

> have have a Model and ModelCollection classes. ModelCollection is my custom array class and it adds methods like batchRefresh() and toGraphQl();
For that I see some other solutions

### Use factory method
```javascript
function createModelCollection(items /* JS array */) {
  const res = observable(items)
  res.batchRefresh = // etc
  return res
}
```

### Use wrapper class
Wrap your own utility class around it, least convenient for consumers probably

### Extend ObservableArray
```javascript
class ModelCollection extends MobX.ObservableArray {

}
```
Not entirely sure how feasible the last approach is, as the constructor is currently quite costum, needing more props than just the array. Might be that the class is not exposed from the package atm, but that could be changed I think

antitoxic:
ok, I will figure something out.

mweststrate:
Success! Sorry for this limitation, ones proxies are GA this limitation is gone (like with observable objects) as the original array instance could be recycled so nothing would be lost

antitoxic:
Well, I have to use a custom `@observable` decorator. It's ugly but it works.

I know, proxies are awesome. Another thing I was thinking (kind of the opposite of this pull request) is that if isObservableArray is used across mobx why not make ObservableArray just Array and duck-type their class, like `if (Array.isArray(arr) && arr.intercept...) { it's an ObservableArray!)` ?

mweststrate:
the problem is that `Array.length` cannot be intercepted (nor inherited), hence observableArrays are now objects, and don't pass `Array.isArray`. (that could be monkey patched, but felt a bridge too far for me)

antitoxic:
Ok, so, just as proxies, we are hitting ES5 limits here. Got it.

PR #609, opened 2016-10-14
Export ObservableArray class

Mosho1:
For testing or other uses.

### DIFF
--- a/src/mobx.ts
+++ b/src/mobx.ts
@@ -34,7 +34,7 @@ export { IListenable                                          } from "./types/li
 export { IObjectWillChange, IObjectChange, IObservableObject, isObservableObject } from "./types/observableobject";
 export { /* 3.0: IValueDidChange, */ IValueWillChange, IObservableValue } from "./types/observablevalue";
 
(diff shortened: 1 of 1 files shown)


andykog:
@Mosho1 I think, ObservableArray is not exported on purpose, because people would make their own classes that inherit ObservableArray and that may lead to different problems, see https://medium.com/@dan_abramov/how-to-use-classes-and-sleep-at-night-9af8de78ccb4#.dscs4s523

mweststrate:
@Mosho1 could you elaborate on the use cases you have gathered so far? What problems would it solve for you?

Mosho1:
@andykog I don't think "hiding" the class from users is the right way to go.

@mweststrate In this case, I wanted to test whether a property is an `ObservableArray`. I saw that `ObservableMap` *is* exported so I figured there is no harm with exposing `ObservableArray` as well.

mweststrate:
wouldn't isObservableArray suffice for that? now that you mention it,
exposing ObservableMap might actually be a mistake as well. One reason not
to expose it is that instanceof checks are not reliable if mobx is loaded
as dependency of another module

Op za 15 okt. 2016 01:36 schreef Sharon Rolel <notifications@github.com>:

mweststrate:
Closes for now, for type check `isObservableArray()` is the preferred method

PR #619, opened 2016-10-20
Expection -> Expecting

ncammarata:
I ran into this error while doing some MobX code writing. I'm actually not sure Expecting is the correct word here. I assumed my best guess as a PR was better than an issue pointing to the typo.

### DIFF
--- a/src/types/type-utils.ts
+++ b/src/types/type-utils.ts
@@ -42,7 +42,7 @@ export function getAtom(thing: any, property?: string): IDepTreeNode {
 }
 
 export function getAdministration(thing: any, property?: string) {
-	invariant(thing, "Expection some object");
+	invariant(thing, "Expecting some object");
 	if (property !== undefined)
 		return getAdministration(getAtom(thing, property));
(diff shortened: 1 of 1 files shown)


capaj:
it is indeed correct word, `expection` does not exist. We should merge this. Only problem is this changes the Error message that is thrown. So it should probably be released in a new major. If it were just up 2 me, I'd probably just merge it, because it is not very likely that anyone is depending on the message of that error. But we certainly need to get more opinions on this- what do you think @mweststrate @tetsuo @andykog, anyone else?

andykog:
@capaj, don't see any problems in merging this. Depending on error's message is not good way to go any way (btw, I'm using [EnhancedError](https://github.com/andykog/enhancederror) because of this), and this message is not mentioned anywhere in the docs.

capaj:
@andykog cool, one more vote for and I'll merge this

ncammarata:
I'm perhaps biased, but waiting on a major release for an error typo fix seems overkill. I'd vote to merge.

mweststrate:
Yeah feel free to merge. People shouldn't depend on the exact spelling of an error message...

Issue #476, opened 2016-08-11
[ReactNative] ListView.DataSource doesn't work with ObservableArrays


winterbe:
Hi,

I'm currently evaluating MobX to be used with React Native. It seems that `ListView.DataSource` unfortunately doesn't work natively with observable arrays from MobX. I have to create a native array via `toJS()` in order the get the `ListView` show any items.

```js
@observer
export default class SampleList extends Component {
    render() {
        const {store} = this.props;

        const dataSource = new ListView.DataSource({
            rowHasChanged: (r1, r2) => r1.id !== r2.id
        });

        const items = toJS(store.items);    // <= conversion to native array necessary

        return (
            <ListView
                dataSource={dataSource.cloneWithRows(items)}
                renderRow={data => (...)}
            />
        );
    }
}
```

I've just started experimenting with MobX but I'm a little concerned that calling `toJS()` for large collections on every render could lead to performance problems.

Please correct me if I'm wrong and there's another way of getting the DataSource to accept ObservableArrays.

I understand that observable types are a consequence of MobX and that you cannot ensure that every library works out of the box with those types. However in ReactNative `ListView` is such a fundamental component that I hope there's a decent solution when using MobX. 

Thanks.

mweststrate:
If I remember correctly (not an active RN dev myself) the ListViewDataSource itself can work with observable arrays, but you need to make sure that your `renderRow` is an observer component. (It looks like being part of the `SampleList`, but actually these component callbacks have their own lifecycle. So

```javascript
renderRow={RowRenderer}

//...

const RowRenderer = observer(data => {})
```

Should do the trick. Let me know if it doesn't :)

winterbe:
@mweststrate Thanks for your reply. Unfortunately it doesn't work for me. RowRenderer is a function that gets called with the rowData for every element in the underlying DataSource. So my code looks like this:

```js
<ListView
  renderRow={rowData => (
    <MyRow data={rowData} .../>
  )}/>
```

I've modified `MyRow` to be an observer but it doesn't make a difference.

mweststrate:
Does it not react to changes in the row data, or to appending / remove items to the collection? For the latter you might need `items.slice()` to make sure RN recognizes as  array (difference with `toJS` is that the first makes a cheap shallow copy, and the latter a deep copy, which shouldn't be needed as individual rowData objects are observed by `MyRow`).

cc: @danieldunderfelt

danieldunderfelt:
Hi @winterbe!

[...]

PR #623, opened 2016-10-22
docs: mobx with react native list views

sonaye:
as discussed in this [issue](https://github.com/mobxjs/mobx/issues/476), added a short explanation on how to properly use mobx with react native list views, a code snippet was provided as well as a reference.

### DIFF
--- a/docs/best/pitfalls.md
+++ b/docs/best/pitfalls.md
@@ -122,7 +122,39 @@ Observable arrays are actually objects, so they comply to `propTypes.object` ins
 
 #### Rendering ListViews in React Native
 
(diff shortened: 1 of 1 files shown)


mweststrate:
Note that it is not necessary to do the `ds.cloneWithRows(this.list.slice());` in the store, it can be safely done in the `List` component as well (either as computed or just inline). Not sure whether it is nice to leak the RN concept of dataSource to a store?

sonaye:
that's true, it can be taken out of the store. my personal approach is that i try to migrate all the "data" state outside the components, also, whomever is utilizing this piece of data, that is the ListView component in this case, doesn't really need to know anything about how that data is handled or generated, ListView should just receive the array it needs, all the fetching and the updating etc is handled by the mobx store.

Issue #634, opened 2016-10-28
ten minute guide: examples execution is slightly out of order


neil-rubens:
https://mobxjs.github.io/mobx/getting-started.html

`run code` seems to execute slightly out of order (by 1); e.g.: 

in the section `A simple todo store...` executing the first example will print to the console (but that should be done when executing the second example).

or executing examples in the `Making React reactive` already shows icons for assignees (something that is added in the next section)

p.s. `mobx` seems great; am looking froward to using it.

mweststrate:
The examples in the 10 minute tutorial are indeed somewhat confusing, especially when you don't read them carefully. They could benefit from some additional love :)

neil-rubens:
The examples are good; just the execution button is slightly out of order.

I can try to update the doc; however I couldn't find the `getting-started.html` in repo.

mweststrate:
it is in the gh-pages branch. thanks in advance!

Op vr 4 nov. 2016 07:18 schreef Neil Rubens <notifications@github.com>:

hugows:
I coudn't understand this one - shouldn't "try MobX" appear on the console log?

![image](https://cloud.githubusercontent.com/assets/11545/21720301/6f7ed7f6-d409-11e6-8a1e-0d82a44dd2f5.png)

mweststrate:
No, the report function only prints the the title of the _first_ todo, so
changing the title of the second todo won't affect the report, hence it is
not printed

Op vr 6 jan. 2017 om 15:13 schreef Hugo Schmitt <notifications@github.com>:

hugows:
Thank you, sorry for the noise.

eladnava:
I would love to give this a shot. 👍

eladnava:
I was able to reproduce the issue with the first `Run` button in `A simple todo store...`. It simply isn't necessary there, since that code outputs nothing when it is executed. 

```js
const todoStore = new TodoStore();
```

But the code section after that does emit output, so that one should have a `Run` button:

```js
todoStore.addTodo("read MobX tutorial");
console.log(todoStore.report());

todoStore.addTodo("try MobX");
console.log(todoStore.report());

todoStore.todos[0].completed = true;
console.log(todoStore.report());

todoStore.todos[1].task = "try MobX in own project";
console.log(todoStore.report());

todoStore.todos[0].task = "grok MobX tutorial";
console.log(todoStore.report());
```

I could not reproduce the behavior you mentioned with the already showing assignee icons.

[...]

PR #647, opened 2016-11-09
resolves: examples execution is slightly out of order #634

neil-rubens:
some of the code was executed ahead of time; removed those references.


mweststrate:
Thanks! I'll (or somebody else) will test this soon

mweststrate:
Hmm.. shouldn't the buttons not just be removed them? The code snippets themselves have no visible side effects. Or maybe the text then should state nothing visible changed?

neil-rubens:
Good point.  To avoid confusion, it would be good to remove buttons for the code blocks that have no visible effect.

mweststrate:
Thanks! I'll merge this, review and process it further

PR #689, opened 2016-11-30
fix #685 - allow wrapped modifier as expression for box computed value

### LINKED ISSUE #685: computed(expression) doesn't work with asStructure modifier
Hey,

I'm using the boxed-computed form, and I'm trying to apply the asStructure modifier:
```javascript
var computed = mobx.computed(mobx.asStructure(() => {
  return {
    prop1: store.a1 || store.a2,
    prop2: store.b
  }
}));
```

As far as I could figure out, this doesn't work because the ```asStructure``` modifier returns an object, while the computed function expects a function as it's parameter in order to create a ```computerExpr```, otherwise it applies the ```computedDecorator``` method.

Not sure if I'm doing something wrong, or if it's simply not supported at the moment.
Would appreciate any input.

yotambarzilay:


src/mobx.ts
```diff
@@ -28,7 +28,7 @@ export { spy                                                  } from "./core/spy
 export { transaction                                          } from "./core/transaction";
 export { IComputedValue                                       } from "./core/computedvalue";
 
-export { asReference, asFlat, asStructure, asMap              } from "./types/modifiers";
+export { asReference, asFlat, asStructure, asMap, IModifierWrapper, ValueMode } from "./types/modifiers";
```
mweststrate: Why is `ValueMode` being exported from the package? Besides that looking good!

mweststrate: Ok, thanks!

PR #694, opened 2016-12-03
Add note about babel-plugin-mobx-deep-action

Strate:
This PR adds a note about our new [babel-plugin-mobx-deep-action](https://github.com/mobxjs/babel-plugin-mobx-deep-action).


Strate:
As for now, plugin is good enough to merge it.

Issue #719, opened 2016-12-23
Fails when `global` not present in browser


philcockfield:
I am running mobx within a webpack configuration that is not putting setting `global` to `window` on the browser.

This is causing this code to fail upon load:

https://github.com/mobxjs/mobx/blob/6e3df706de1f1fc2408ba919559297cd0d07573e/src/core/globalstate.ts#L91

Error: 
    Uncaught TypeError: Cannot read property '__mobservableTrackingStack' of undefined

Would it make sense check whether `window` is available, and write the global values to that object instead of assuming global is set?

**Some Context:** The webpack configuration that is causing this is coming from the latest version of `nextjs` https://github.com/zeit/next.js/

----

- Mobx Version: `2.7.0`
- Browser: Chrome
- Node: 6.9.1
- Reproduction: Start a new nextjs project (using the latest version 2.x) and reference mobx.

andykog:
Phil, is there any reason to close this?

philcockfield:
This came out of an issue with Next.js (beta) failing to put create a `global` global.  As it hasn't been an issue prior to that - and they are in beta, it felt possibly a bit premature to have it as an issue on Mobx.  

Apparently it will get fixed with Webpack2, and so I didn't want to leave unnecessary noice on this.

Happy to re-open it if folks think this is a wise defensive check to make.

PR #720, opened 2016-12-23
Ensure `global` exists on browser.

philcockfield:
Documented in issue: https://github.com/mobxjs/mobx/issues/719

### DIFF
--- a/src/core/globalstate.ts
+++ b/src/core/globalstate.ts
@@ -2,7 +2,11 @@ import {IDerivation} from "./derivation";
 import {Reaction} from "./reaction";
 import {IObservable} from "./observable";
 
-declare const global: any;
+declare let global: any;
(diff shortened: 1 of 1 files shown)


src/core/globalstate.ts
```diff
@@ -2,7 +2,11 @@ import {IDerivation} from "./derivation";
 import {Reaction} from "./reaction";
 import {IObservable} from "./observable";
 
-declare const global: any;
+declare let global: any;
+if (typeof global === "undefined" && typeof window !== "undefined")
+	global = window;
+
+
```
andykog: @philcockfield, I beleave, with this implementation you'll get `Uncaught ReferenceError` in strict mode

philcockfield: Ah rats! 💩

Issue #721, opened 2016-12-23
report errors from catch block instead of finally


amir-arad:
Hi
I sometimes get Mobx warnings about an uncaught exception from a finally block at `peekUntracked()` in computedValue.ts.
catching the original exception is sometimes tricky and time consuming (like when you run cluncky test tools that has a lot of caught exceptions themselves). It would be much easier if the spy / console.warn would be triggered from within a context that has the uncaught exception - the catch clause, as the line calls console.warn is one click away, so it's simple to add a breakpoint there.
another option is adding the offending exception to the spy report, in addition to throwing it.

mweststrate:
@amir-arad, totally unrelated, tell me what cool stuff is happening here :) https://github.com/wix/mutable

amir-arad:
Mutable is a data class library with type validations, read-only views etc. 
I've recently replaced its internal internal dirty checking mechanism with mobx, now i'm translating it to typescript and generally making it faster and prettier.

amir-arad:
I've just noticed, the readme is very outdated 😊

mweststrate:
Maybe the nice solution to this problem is to not do anything with exceptions in general. I think MobX since the latest versions recovers in all cases nicely from errors. So I think the warning is might be no longer needed at all, and that it should just be documented that errors are not propagated to observers.

mweststrate:
Closed in favor of follow up #731

benjamingr:
@amir-arad mutable sounds cool!

PR #722, opened 2016-12-23
suggestion for verbose error reporting as per #721

amir-arad:
I didn't find the right place to add tests for the new field in spy error message, would love a point to the right direction :)

time report in `npm run perf` seems the same with +-10% range of error , also this change should only slightly affect behavior of systems already in error mode (as the catch blocks were added to already non optimizable functions)

### DIFF
--- a/src/core/derivation.ts
+++ b/src/core/derivation.ts
@@ -119,15 +119,16 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T) {
 	derivation.runId = ++globalState.runId;
 	const prevTracking = globalState.trackingDerivation;
 	globalState.trackingDerivation = derivation;
-	let hasException = true;
+	let hasException = false;
 	let result: T;
 	try {
 		result = f.call(derivation);
-		hasException = false;
+	} catch(e){
+		hasException = true;
+		handleExceptionInDerivation(derivation, e);
+		throw e;
 	} finally {
-		if (hasException) {
-			handleExceptionInDerivation(derivation);
-		} else {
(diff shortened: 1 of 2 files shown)


amir-arad:
fixes #721

src/core/derivation.ts
```diff
@@ -119,23 +119,24 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T) {
...
-	let hasException = true;
+	let hasException = false;
 	let result: T;
 	try {
 		result = f.call(derivation);
-		hasException = false;
+	} catch(e){
+		hasException = true;
+		handleExceptionInDerivation(derivation, e);
+		throw e;
```
mweststrate: The problem with this change is this changes the stack trace of the exception to this line. MobX had something similar in the past, but since the original stack is lost with a `catch`, people will blame MobX for the exception, which results in a plethorea of issues in the MobX issue tracker :). Sadly this cannot be fixed really nicely in Javascript as far as I know :cry:

spion: The stack trace is generated when `new Error` is called, so rethrowing the same error object shouldn't change the stack (unless `handleExceptionInDerivation` mutates the error object)

amir-arad: what @spion said ... 
what do you think about a test to assert the original stack trace?

mweststrate: See: https://github.com/mobxjs/mobx/issues/731

mweststrate:
Merged into, and closed in favor of #736. Thanks for stirring up the discussion!

PR #725, opened 2016-12-27
Update docs to mobx 3

mweststrate:
nuf said

### DIFF
--- a/docs/refguide/action.md
+++ b/docs/refguide/action.md
@@ -3,10 +3,12 @@
 Usage:
 * `action(fn)`
 * `action(name, fn)`
-* `@action classMethod`
-* `@action(name) classMethod`
+* `@action classMethod() {}`
+* `@action(name) classMethod () {}`
 * `@action boundClassMethod = (args) => { body }`
 * `@action(name) boundClassMethod = (args) => { body }`
+* `@action.bound classMethod() {}`
+* `@action.bound(function() {})
 
 Any application has actions. Actions are anything that modify the state.
 With MobX you can make it explicit in your code where your actions live by marking them.
@@ -68,3 +70,39 @@ Example:
 The usage of `runInAction` is: `runInAction(name?, fn, scope?)`.
 
 If you use babel, this plugin could help you to handle your async actions: [mobx-deep-action](https://github.com/mobxjs/babel-plugin-mobx-deep-action).
+
+## Bound actions
+
+The `action` decorator / function follows the normal rules for binding in javascript.
+However, Mobx 3 introduces `action.bound` to automatically bind actions to the targeted object.
+
+Example:
+
+```javascript
+class Ticker {
+	@observable this.tick = 0
+
+	@action.bound
+	increment() {
+		this.tick++ // 'this' will always be correct
+	}
+}
+
+const ticker = new Ticker()
+setInterval(ticker.increment, 1000)
+```
+
+Or
+
+```javascript
+const ticker = observable({
+	tick: 1,
+	increment: action.bound(function() {
+		this.tick++ // bound 'this'
+	})
+})
+
+setInterval(ticker.increment, 1000)
+```
+
+_Note: don't use *action.bind* with arrow functions; arrow functions are already bound and cannot be rebound._
--- a/docs/refguide/action.md
+++ b/docs/refguide/action.md
@@ -75,6 +75,7 @@ If you use babel, this plugin could help you to handle your async actions: [mobx
 
 The `action` decorator / function follows the normal rules for binding in javascript.
 However, Mobx 3 introduces `action.bound` to automatically bind actions to the targeted object.
+Note that `(@)action.bound` does, unlike `action`, not take a name parameter, the name will always be based on the property the action is bound to.
 
 Example:
 
--- a/docs/refguide/api.md
+++ b/docs/refguide/api.md
@@ -1,46 +1,129 @@
 # MobX Api Reference
 
+Applies to MobX 3 and higher. For MobX 2, the old documentation is still available on [githib](https://github.com/mobxjs/mobx/blob/7c9e7c86e0c6ead141bb0539d33143d0e1f576dd/docs/refguide/api.md)
+
 # Core API
 
 _The most important MobX api's. Understanding `observable`, `computed`, `reactions` and `actions` is enough to master MobX and use it in your applications!_
 
 ## Creating observables
 
-### `observable`
+
+### `observable(value)`
 Usage:
 * `observable(value)`
 * `@observable classProperty = value`
 
 Observable values can be JS primitives, references, plain objects, class instances, arrays and maps.
+`observable(value)` is a convenience overload, that always tries to create the best matching observable types.
+You can also directly create the desired observable type, see below.
+
 The following conversion rules are applied, but can be fine-tuned by using *modifiers*. See below.
 
-1. If *value* is wrapped in the *modifier* `asMap`: a new [Observable Map](map.md) will be returned. Observable maps are very useful if you don't want to react just to the change of a specific entry, but also to the addition or removal of entries.
+1. If *value* is an wrapped is an instance of an [ES6 Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map): a new [Observable Map](map.md) will be returned. Observable maps are very useful if you don't want to react just to the change of a specific entry, but also to the addition or removal of entries.
 1. If *value* is an array, a new [Observable Array](array.md) will be returned.
-1. If *value* is an object *without* prototype, all its current properties will be made observable. See [Observable Object](object.md)
+1. If *value* is an object *without* prototype, the object will be cloned and all its current properties will be made observable. See [Observable Object](object.md)
 1. If *value* is an object *with* a prototype, a JavaScript primitive or function, a [Boxed Observable](boxed.md) will be returned. MobX will not make objects with a prototype automatically observable; as that is the responsibility of it's constructor function. Use `extendObservable` in the constructor, or `@observable` in it's class definition instead.
 
 These rules might seem complicated at first sight, but you will notice that in practice they are very intuitive to work with.
 Some notes:
-* To create dynamically keyed objects use the `asMap` modifier! Only initially existing properties on an object will be made observable, although new ones can be added using `extendObservable`.
+* To create dynamically keyed objects always use maps! Only initially existing properties on an object will be made observable, although new ones can be added using `extendObservable`.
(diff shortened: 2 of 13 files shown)


docs/refguide/action.md
```diff
@@ -68,3 +70,40 @@ Example:
...
+
+Example:
+
+```javascript
+class Ticker {
+	@observable this.tick = 0
+
+	@action.bound
+	increment() {
+		this.tick++ // 'this' will always be correct
```
dmitriid: Please explain why this might not be correct using plain `action`

docs/refguide/action.md
```diff
@@ -68,3 +70,40 @@ Example:
 The usage of `runInAction` is: `runInAction(name?, fn, scope?)`.
 
 If you use babel, this plugin could help you to handle your async actions: [mobx-deep-action](https://github.com/mobxjs/babel-plugin-mobx-deep-action).
+
+## Bound actions
+
+The `action` decorator / function follows the normal rules for binding in javascript.
+However, Mobx 3 introduces `action.bound` to automatically bind actions to the targeted object.
```
dmitriid: Please shortly explain the difference as in:

> The `action` decorator / function follows the normal rules for binding in javascript... That is it [only] binds this and that
> ... `action.bound` to automatically bind actions to the targeted object. That is in this and that situation

docs/refguide/api.md
```diff
@@ -1,46 +1,129 @@
...
 * `@observable classProperty = value`
 
 Observable values can be JS primitives, references, plain objects, class instances, arrays and maps.
+`observable(value)` is a convenience overload, that always tries to create the best matching observable types.
+You can also directly create the desired observable type, see below.
+
 The following conversion rules are applied, but can be fine-tuned by using *modifiers*. See below.
 
-1. If *value* is wrapped in the *modifier* `asMap`: a new [Observable Map](map.md) will be returned. Observable maps are very useful if you don't want to react just to the change of a specific entry, but also to the addition or removal of entries.
+1. If *value* is an wrapped is an instance of an [ES6 Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map): a new [Observable Map](map.md) will be returned. Observable maps are very useful if you don't want to react just to the change of a specific entry, but also to the addition or removal of entries.
```
dmitriid: > is an wrapped is an instance

should probably be

> is a wrapped instance

docs/refguide/computed-decorator.md
```diff
@@ -30,33 +44,34 @@ class OrderLine {
 }
 ```
 
-If your environment doesn't support decorators or field initializers,
-`@computed get funcName() { }` is sugar for [`extendObservable(this, { funcName: func })`](extend-observable.md)
-
+## `computed` modifier
 
-`@computed` can be parameterized. `@computed({asStructure: true})` makes sure that the result of a derivation is compared structurally instead of referentially with its preview value. This makes sure that observers of the computation don't re-evaluate if new structures are returned that are structurally equal to the original ones. This is very useful when working with point, vector or color structures for example. It behaves the same as the `asStructure` modifier for observable values.
+If your environment doesn't support decorators, use the `computed(expression)` modifier incombination with `extendObservable` / `observable` to introduce new computed properties.
```
dmitriid: > incombination

docs/refguide/api.md
```diff
@@ -1,46 +1,129 @@
...
+```
+
+[&laquo;details&raquo;](modifiers.md)
+
+
 ## Computed values
 
 Usage:
 * `computed(() => expression)`
+* `computed(() => expression, (newValue) => void)`
```
dmitriid: `computed(() => expression, (newValue) => void)` isn't explained in the `computed` docs, only `computed(() => expression)` and `computed(() => expression, options)` are explained

docs/refguide/extend-observable.md
```diff
@@ -22,4 +20,57 @@ extendObservable(matthew, {
...
-(N.b:  `observable(object)` is actually an alias for `extendObservable(object, object)`)
+N.b:  `observable.object(object)` is actually an alias for `extendObservable({}, object)`.
+
+Note that the property maps are not always copied literally onto the target, but they are considered property descriptor.
+Most values are copied as-is, but values wrapped in a modifier as treated specially. And so are properties that have a getter.
+
+## Modifiers
+
+[Modifiers](modifiers.md) can be used to define special behavior for certain properties.
+For example `observable.ref` creates an observable reference which doesn't automatically convert it's values into observables, and `computed` introduces a derived property:
```
dmitriid: `it's` -> `its`

docs/refguide/map.md
```diff
@@ -1,17 +1,18 @@
 # Observable Maps
 
-`observable(asMap(values?, modifier?))` (and `map(values?, modifier?)`) creates a dynamic keyed observable map.
+## `observable.map(values)`
+
+`observable.map(values?)` creates a dynamic keyed observable map.
```
dmitriid: It probably makes sense to add 

> To create dynamically keyed objects always use maps!

here as well. You can never have too few warnings :)

docs/refguide/modifiers.md
```diff
@@ -1,132 +1,115 @@
 # Modifiers for observable
 
-By default, `observable` recursively makes all the values of _plain_ objects and arrays recursively observable.
-Besides that, it automatically converts functions without arguments into reactive views or derived properties.
-For all other types of values just a reference is stored.
-In general, this should just do what you need, but if you want you can override the default behavior using _modifiers_.
-Note that modifiers are 'sticky', they are interpreted as being annotations.
-They do not only apply to the current value, but also to all values that are assigned in the future to the same attribute.
+Modifiers can be used decorator or in combination with `extendObservable` and `observable.object` to change the autoconversion rules for specific properties.
```
dmitriid: `can be used decorator` -> `can be used as decorators`

docs/refguide/modifiers.md
```diff
@@ -1,132 +1,115 @@
...
-## asReference
+```javascript
+class Message {
+    @observable message = "Hello world"
 
-The most common modifier is `asReference`.
-If this modifier is used, `observable` will not attempt to make the value observable.
-Use this for example if you want to store a reference to a function, instead of creating a view based on that function.
-You can also use it to prevent that plain objects or arrays are made observable automatically.
+    // ficitonal example, if author is immutable, we just need to store a reference and shouldn't turn it into an mutable, observable object
```
dmitriid: `ficitonal ` -> `fictional`

docs/refguide/modifiers.md
```diff
@@ -1,132 +1,115 @@
...
 
-However, if the window size is smaller than the `minSize`, the resize doesn't need to influence the rendering anymore, as the computed
-will return the same dimensions after each run. `asStructure` signals to MobX that observers of this computation should only be triggered
-if the value returned by the computed has _structurally_ changed (by default strict equality is used to determine whether observers need to be notified).
-This means that a new object that is returned from `viewPortSize` won't trigger a `render` if its contents are (structurally) the same as the previous value.
+## Action & Computed
 
-To use the `asStructure` modifier in combination with the `@computed` decorator, use the following:
+`action`, `action.bound`, `computed` and `computed.struct` can be used as modifiers as well.
+See [`computed`](computed-decorator.md) respectively [`action`](action.md).
```
dmitriid: > See computed and action respectively
(word order)

docs/refguide/action.md
```diff
@@ -68,3 +70,40 @@ Example:
...
+	tick: 1,
+	increment: action.bound(function() {
+		this.tick++ // bound 'this'
+	})
+})
+
+setInterval(ticker.increment, 1000)
+```
+
+_Note: don't use *action.bind* with arrow functions; arrow functions are already bound and cannot be rebound._
```
BurntCaramel: Typo: should say`action.bound`

docs/refguide/api.md
BurntCaramel: Yes, does the setter behave as an action?

docs/refguide/computed-decorator.md
```diff
@@ -1,17 +1,31 @@
...
+Computed values can be optimized away in many cases by MobX as they are assumed to be pure.
+For example, a computed property won't re-run if none of the data used in the previous computation did change.
+Nor will a computed property re-run if is not in use by some other computed property or reaction.
+In such cases it will be suspended.
 
-Computed properties can be optimized away in many cases by MobX as they are assumed to be pure.
-So they will not be invoked when their input parameters didn't modifiy or if they are not observed by some other computed value or autorun.
+This automatic suspension is very convenient; it means that a computed value that is not observed anymore (for example because the UI in which it was used has disappeared)
+can automatically be garbabged collected by MobX, and unlike `autorun`'s you don't have to dispose them yourself.
+This sometimes confuses people new to MobX; if you create a computed property but don't use it anywhere in a reaction, it will not cache it's value and recompute more often than seems necesarray.
```
BurntCaramel: Typo: it's -> its
Typo: necesarray -> necessary

docs/refguide/modifiers.md
```diff
@@ -1,132 +1,115 @@
...
+
+const todos = [{ test: "value" }]
+const store = new Store()
+
+store.collection1 = todos;
+store.collection2 = todos;
+store.collection3 = todos;
 ```
+
+After these assignments:
```
BurntCaramel: This is a great example!

docs/refguide/object.md
```diff
@@ -1,7 +1,7 @@
 ## Observable Objects
 
-If a plain JavaScript object is passed to `observable` all properties inside that object will be made observable.
-(A plain object is an object that wasn't created using a constructor function)
+If a plain JavaScript object is passed to `observable` all properties inside will be copied into a clone and made observable.
+(A plain object is an object that wasn't created using a constructor function / but has `Object` as it's prototype, or no prototype at all.)
```
BurntCaramel: its

docs/refguide/reaction.md
```diff
@@ -1,20 +1,32 @@
...
-It takes two functions, the first one is tracked and returns data that is used as input for the second one, the side effect.
+It takes two functions, the first one (the *data* function) is tracked and returns data that is used as input for the second one, the *effect* function.
 Unlike `autorun` the side effect won't be run directly when created, but only after the data expression returns a new value for the first time.
 Any observables that are accessed while executing the side effect will not be tracked.
+
 The side effect can be debounced, just like `autorunAsync`.
 `reaction` returns a disposer function.
-If a string is passed as first argument to `reaction`, it will be used as debug name.
-The functions passed to `when` will receive one argument when invoked, the current reaction, which can be used to dispose the when during execution.
+The functions passed to `reaction` will receive one argument when invoked, the current reaction, which can be used to dispose the when during execution.
```
BurntCaramel: `dispose the when`?

PR #737, opened 2017-01-07
Add link to my articles :-)

alexhisen:


mweststrate:
Thanks! Small suggestion: mention observable maps in the extendObservable section. When using observable maps mobx can handle property additions reliably without being dependent on the computation order :)

alexhisen:
Thanks Michel. I made the mention of observable maps more explicit. Any other suggestions are most welcome.

PR #739, opened 2017-01-07
Update docs

rlesniak:
Remove obsolete not about componentWillReact mobx-react@4 has changed semantic of `props` and `state`, which became observables. So, this not became obsolete too.

### DIFF
--- a/docs/refguide/observer-component.md
+++ b/docs/refguide/observer-component.md
@@ -164,7 +164,6 @@ import {observer} from "mobx-react";
 
 * `componentWillReact` doesn't take arguments
 * `componentWillReact` won't fire before the initial render (use `componentWillMount` instead)
-* `componentWillReact` won't fire when receiving new props or after `setState` calls (use `componentWillUpdate` instead)
 
 ## Optimizing components
 


docs/refguide/observer-component.md
```diff
@@ -164,7 +164,6 @@ import {observer} from "mobx-react";
 
 * `componentWillReact` doesn't take arguments
 * `componentWillReact` won't fire before the initial render (use `componentWillMount` instead)
-* `componentWillReact` won't fire when receiving new props or after `setState` calls (use `componentWillUpdate` instead)
```
capaj: I am thinking it would be better to just change  `won't fire` into `will fire`. It's useful to leave this note here, because a lot of programmers won't know how react-mobx works under the hood and they would expect this to be true.

mweststrate: @capaj what do you mean with `will fire`? comment seems to be correct

capaj: @mweststrate well this PR is about removing it. I propose to leave it there, but just making it true by removing the negative.

It is my understanding that for mobx-react 4.0.0 and up this must have changed since the `state` and `props` is now observable as well.

mweststrate: Check! will update

PR #759, opened 2017-01-14
removes deviation from ecmascript array, fixes #745

### LINKED ISSUE #745: observableArray toString method seems fishy
I was debugging a bug in our app, where calling `encodeUriComponent()` on an observableArray produced very weird results. Turns out toString method is returning something else than the ecmascript array:
```
> var mobx=require('mobx')
undefined
> a = mobx.observable([])
ObservableArray {}
> encodeURIComponent(a)
'%5Bmobx.array%5D%20'
> encodeURIComponent([])
''
> a.toString()
'[mobx.array] '
> [].toString()
''
> [1].toString()
'1'
> a.push(1)
1
> a.toString()
'[mobx.array] 1'

```

@mweststrate/anyone why is it like this? Could we maybe match the behaviour in regular array? That would certainly fix any issues people might have when they try to encode their observableArrays into query strings.

EDIT: Looking into the source https://github.com/mobxjs/mobx/blob/2cc98ae92c56ed8f8371875436aba75657a8bab1/src/types/observablearray.ts#L383 it is indeed done like this. What is the reasoning for explicitly not matching the behaviour of ecmascript array here?

capaj:
issue: #745
PR checklist:

* [X] Implementation
* [x] Added tests
* [x] Updated (either in the description of this PR as markdown, or as seperate PR on the `gh-pages` branch. Please refer to this PR). For new functionality, at least [API.md](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md) should be updated
* [x] Added type annotations (if you are unfamiliar with typescript, untyped, `:any` based PR's are welcome as well

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -380,11 +380,13 @@ export class ObservableArray<T> extends StubArray {
 	}
 
 	toString(): string {
-		return "[mobx.array] " + Array.prototype.toString.apply(this.$mobx.values, arguments);
+		this.$mobx.atom.reportObserved();
+		return Array.prototype.toString.apply(this.$mobx.values, arguments);
 	}
 
 	toLocaleString(): string {
-		return "[mobx.array] " + Array.prototype.toLocaleString.apply(this.$mobx.values, arguments);
+		this.$mobx.atom.reportObserved();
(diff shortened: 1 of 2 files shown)


mweststrate:
Great, thanks!

PR #765, opened 2017-01-16
Fixes runInAction flow type, name is not mandatory

wietsevenema:
Fixes runInAction flow type, name is not mandatory


mweststrate:
Thanks! Will include in next release

PR #766, opened 2017-01-16
Also publish flow typings to npm

wietsevenema:
Also publish the flow typings to npm, so we can use them. 

Where should I add the documentation on how to use this? You should add this to the `[libs]` section of your .flowconfig. 
```
[libs]
node_modules/mobx/lib/mobx.js.flow
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,13 @@ And finally kudos for all the people that believed in, tried, validated and even
 
 > Note: Before testing, make sure to run `npm run small-build`.
 
+## Flow support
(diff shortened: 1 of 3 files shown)


mweststrate:
@wietsevenema could you add that to the readme, probably somewhere near the installation notes. Thanks!

wietsevenema:
- Added documentation next to Bower support.
 - Rewrote using new dependency: ncp, to keep the build windows friendly. Had to after #770 was fixed.

mweststrate:
Awesome, thanks a lot!

Issue #652, opened 2016-11-13
Introduce `dynamicObject` for dynamically keyed objects in Proxy supporting environments


mweststrate:
See also #336 

Introduce `object = dynamicObject(object)` and implement that using ES6 proxies. This allows plain objects to be used as observable objects, without the current limitation of not supporting dynamic keys.

I think that is a nice case to start working with Proxies, learn about their caveats, performance etc. It has a clear use case and advantage over the current getter / setter implementation, yet there is a fallback for those needing to support browsers without proxies (probably the most of us): `ObservableMaps`

Example:
```javascript
const todos = dynamicObject({
  "a23bf": { completed: false}
})

autorun(() => console.log(Object.keys(todos)))
// prints "a23bf"

todos["a8235"] = { completed: true}
// prints "a23bf", "a8235", something that currently can only be achieved through observable maps
```

Should throw clear error if proxies (or too limited polyfill) are not available at runtime and hint at using ObservableMaps instead

solkimicreb:
Stale data in transparent reactivity is new to me as I never worked on getter/setter based solutions and I never experienced the issue with Proxies. I have a few questions about why stale derivations could occur. I have a feeling that proxies can solve it for free.

- Does MobX save actual save data (values) of computed properties?
- Does MobX transform computed props into getters? If so, is it the reason why it needs to save a cached value? (Otherwise it could be simply the native equivalent of computed, which is a getter.)
- Finally, are these the reason of possible staleness? Are there any other possible reason, apart from computed derivations?

With Proxies I don't see any reason to save any data on the framework level. Proxies can wrap getters/setters and even other Proxies. Without storing data the issue is gone in my opinion. With Proxies I never met with stale derivations. I might be missing something and I might be just lucky though.

amir-arad:
MINE!

mull:
Did anyone start working on this yet?

amir-arad:
I have #776 with pending tasks, but it's not going to get any serious boost from my end

ascoders:
Hi @mweststrate , I implemented mobx in proxy, although not all of the API has been implemented, but the core functionality has been opened and has passed 50 unit tests and has been run online.

ascoders:
And implemented a similar mobx-react called [dynamic-react](https://github.com/ascoders/dynamic-react), i wish I could merge into mobx

mweststrate:
@ascoders didn't see the comments earlier, thanks! Will check the code out when starting on proxies. 

[...]

PR #776, opened 2017-01-17
ES6 Proxy

amir-arad:
naive solution proposal for #652

uses shallow observable map internally, because it fits the data structure's dynamic nature.

done as part of a #goodnessSquad hackathon


amir-arad:
@idoros

src/api/observable.ts
```diff
@@ -116,6 +116,35 @@ export class IObservableFactories {
...
+				}
+				return target.get(property);
+			},
+			set(target, property:string, value, receiver){
+				if (property === '$mobx'){
+					return false;}
+				target.set(property, value);
+				return true;
+			},
+			ownKeys: function(target) {
```
solkimicreb: This is used by `for in` loops internally too, so they work out of the box with expando props.

src/api/observable.ts
```diff
@@ -116,6 +116,35 @@ export class IObservableFactories {
...
+				return target.keys();
+			},
+			getOwnPropertyDescriptor: function(target, property:string) {
```
solkimicreb: Is this necessary? Why not just leave the default property descriptor? If it is necessary shouldn't it include `writable: true` and `value: target.get(prop)`?

src/api/observable.ts
```diff
@@ -116,6 +116,35 @@ export class IObservableFactories {
 		return res as any;
 	}
 
+	dynamic<T>(props: T, name?: string): T & IObservableObject {
```
solkimicreb: Maybe check of the props object is already wrapped by a dynamic observable Proxy. Accidental double wrapping could cause some nasty bugs.

src/api/observable.ts
```diff
@@ -116,6 +116,35 @@ export class IObservableFactories {
...
+		if (arguments.length > 2)
+			incorrectlyUsedAsDecorator("object");
+		invariant(typeof Proxy === 'function', "dynamic objects are not supported in this environment");
+		const internalMap = new ObservableMap(props, referenceEnhancer, name);
+		const result = new Proxy(internalMap, {
```
solkimicreb: I think you could also add a `deleteProperty` and maybe a `has` trap for the sake of completeness.

solkimicreb:
I don't know much about MobX, but I added a few general tips. Nice and clean approach in my opinion 🙂

src/api/observable.ts
amir-arad: before we added `getOwnPropertyDescriptor`, Object.keys(foo) always returned an empty array, no matter what `ownKeys` actually returned.

amir-arad: AFAIK he rest of the fields have default fallbacks that fits our use case. `configurable` specifically breaks the Proxy when it does not match the internal property's `configurable` value, that is why we added it as well.

src/api/observable.ts
amir-arad: I didn't think about that. thanks!

amir-arad:
TODOs:
 - [ ] test and avoid double-wrap on already boxed argument
 - [ ] implement other proxy traps (`has`, `delete`)
 - [ ] test and support spy and observe events
Thanks for the feedback @solkimicreb !

solkimicreb:
Deep wrapping will be a tricky one. It won't be enough to traverse the structure at start and wrap everything into an observable, because of the potential object valued expando properties.

mweststrate:
Proxies are no on their way! So closing this PR in favor #1380, but thanks for the setup @amir-arad! Will borrow some stuff from this. Sad it took a year, but finally browser support is up to par :)  (although, IE, RN android, sigh)

PR #779, opened 2017-01-18
Missing es6 map error

dirtyrolf:
In an environment where the ES6 Map is missing, any call to observable() will throw an uncaught error at the `isES6Map` check in `utils/utils.ts` and the observable won't be created. This pull request is a simple fix that makes sure that we don't try to call `type instanceof undefined`.

PR checklist:

* [x] Implementation
* [x] Added tests
* [ ] Updated (either in the description of this PR as markdown, or as seperate PR on the `gh-pages` branch. Please refer to this PR). For new functionality, at least [API.md](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md) should be updated
* [ ] Added type annotations (if you are unfamiliar with typescript, untyped, `:any` based PRs are welcome as well

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -199,7 +199,7 @@ export function isArrayLike(x: any): boolean {
 }
 
 export function isES6Map(thing): boolean {
-	if (thing instanceof getGlobal().Map)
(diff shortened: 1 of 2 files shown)


mweststrate:
Great work @dirtyrolf! Will probably be released later today :)

PR #786, opened 2017-01-20
Fix sentence in api overview

k15a:
There is nothing React specific above the line but below. Correct me please if I am wrong.


mattruby:
Thanks!

PR #787, opened 2017-01-20
Link to mobx-utils instead of rx-mobx in FAQ

k15a:
`rx-mobx` is deprecated in favour of `mobx-utils`.


mattruby:
Thanks!  You're on a roll!

PR #805, opened 2017-01-29
Autorun & computed values should believe NaN === NaN

andykog:


capaj:
@andykog JS master badge earned :+1:

mweststrate:
Great stuff @andykog!

PR #817, opened 2017-02-03
observe: Use typescript index types

Strate:
This enables better typecheck and inference.


hccampos:
This will break compatibility for everyone using TS < 2.1 if I am not mistaken. Would probably warrant a major version bump

mweststrate:
Great @Strate! Thanks for pointing out @hccampos, will indeed be breaking for TS 2.0 users. Will merge it as soon as MobX 4 is nigh. Note that MobX 3 won't be as long lived as MobX 2, so shouldn't take too long :)

mweststrate:
Merged into #1321, closing here.

PR #818, opened 2017-02-03
observe, typescript: index & mapped types

Strate:
Enable index types and mapped types on `IObjectdidChange` for better typecheck and inference

### DIFF
--- a/src/api/observe.ts
+++ b/src/api/observe.ts
@@ -1,6 +1,6 @@
 import {IObservableArray, IArrayChange, IArraySplice} from "../types/observablearray";
 import {ObservableMap, IMapChange} from "../types/observablemap";
-import {IObjectChange} from "../types/observableobject";
+import {IObjectDidChange} from "../types/observableobject";
 import {IComputedValue} from "../core/computedvalue";
 
 import {IObservableValue, IValueDidChange} from "../types/observablevalue";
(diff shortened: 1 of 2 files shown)


src/api/observe.ts
```diff
@@ -11,7 +11,7 @@ export function observe<T>(value: IObservableValue<T> | IComputedValue<T>, liste
 export function observe<T>(observableArray: IObservableArray<T>, listener: (change: IArrayChange<T> | IArraySplice<T>) => void, fireImmediately?: boolean): Lambda;
 export function observe<T>(observableMap: ObservableMap<T>, listener: (change: IMapChange<T>) => void, fireImmediately?: boolean): Lambda;
 export function observe<T>(observableMap: ObservableMap<T>, property: string, listener: (change: IValueDidChange<T>) => void, fireImmediately?: boolean): Lambda;
-export function observe(object: Object, listener: (change: IObjectChange) => void, fireImmediately?: boolean): Lambda;
+export function observe<T>(object: T, listener: (change: IObjectDidChange<T, keyof T>) => void, fireImmediately?: boolean): Lambda;
```
Igorbek: Setting type of keys to `keyof T` is not quite correct in TS, unless `T` would represent an exact type (not supported yet).
Consider your `T` is `{ a; b; }`, and your runtime observable object is `{ a: 1, b: 2, c: 3 }` which is totally ok as it is assignable to `T`.
Now your `listener` can at most only expects `"a" | "b"` which is not aligned with the runtime possibility to receive `"c"`.

Strate: Yes, sure. Same thing exists in `Object.keys` function definitions in lib.d.ts: this function returns `string[]`, not a `keyof` passed object. I think this can stay as it is.

PR #824, opened 2017-02-07
Improve flow typing for action method

newoga:
This PR adds two new type definitions for the `action()` method:

`action(name: string, func: T): T`
`action(func: T): T`

More specifically, this types the non-decorator implementation of `action()`.

The existing `action()` type definition returns `any`, which causes type information associated with the wrapped function to be lost. At some point, we should probably investigate flow-type support for decorators. I could be mistaken but I think decorators in flow-type either causes warnings or are ignored, so the current definition may not provide the intended value.


mweststrate:
Thanks!

PR #842, opened 2017-02-20
Clarify English

chasm:
Generally, "but also" is preceded by "not only", which makes this sentence clearer (it's a correlative conjunction). The previous text said that the values _can't_ be primitive values, but of course they can. It is that they are not _limited_ to being primitive values.


mattruby:
Thanks!

PR #844, opened 2017-02-20
Remove types in Javascript code examples

emiloberg:
Some sneaky numbers managed to find their way in from typed land.


mattruby:
Good catch, thanks!

PR #849, opened 2017-02-24
[docs] lately → late

radex:
The use of `lately` is confusing (and I think, incorrect) in this context. "lately" means "recently" -- "not long ago". What the docs mean is to dereference _late_ in the process.


mattruby:
Thanks!

PR #857, opened 2017-02-28
Fix typo, clarify comparison

kourge:
The original typo made it seem like the two approaches can happen sequentially when, after re-reading, it seems like the intent is to convey that `untracked` is a lower-level operation than `(@)action`, and that if you're using `(@)action`, then `untracked` isn't necessary.


mattruby:
Thanks!

Issue #837, opened 2017-02-16
Reaction with delay is not fired, when value changes for the first time before: time = creationTime + delay


SaboteurSpk:
When reaction is created with delay option, the reaction will not react to changes of value if the value changes before the delay runs out after creation of the reaction.

Example:

reaction(
   () => user.name,
   (name) => console.log(name),
   {delay: 5000}
);
// will do nothing
user.name = 'test';
// Just this will change will run the reaction
setTimeout(() => user.name = 'test2', 10000);

see this fiddle:
https://jsfiddle.net/js1k4hsn/

andykog:
With `fireImmediately: true` it works as @SaboteurSpk expected it to work, but I believe that option should make it fire _really_ immediately without the initial delay

SaboteurSpk:
Hmm, so with @andykog comment I see two unexpected behaviours now :)

1) reaction should be called even if value changes before delay timespan expires after creation of reaction, but it is not. 
2) fireImmediately: true with delay is not run immediately as it's name says, but with delay.

SaboteurSpk:
#860

mweststrate:
Released as 3.1.5. Thanks for the PR!

SaboteurSpk:
Works like a charm. Thanx for publishing and cooperation! Have a nice day!

PR #860, opened 2017-03-01
Fix reaction with delay - first run of expression should not be delayed - just the reaction call

SaboteurSpk:


mweststrate:
Double check: this change fixes both issues mentioned in #837?

SaboteurSpk:
Just the delay thing. If this behaviour was not intended.

SaboteurSpk:
And this was my first modification of MobX sources so ... watch out :)

PR #867, opened 2017-03-04
Improve class action autonaming

iotch:
I'm not sure this is the right place, but AFAIK it is the only way for now to include class name with action name without changing the ObservableValue API. I think this change is too minor to follow the checklist.

* [ ] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/src/api/action.ts
+++ b/src/api/action.ts
@@ -81,7 +81,9 @@ export var action: IActionFactory = function action(arg1, arg2?, arg3?, arg4?): 
 	if (arguments.length === 1 && typeof arg1 === "string")
 		return namedActionDecorator(arg1);
 
-	return namedActionDecorator(arg2).apply(null, arguments);
+    const targetName = arg1.constructor.name;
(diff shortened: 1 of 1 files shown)


mweststrate:
Some comments:

1. Note that there is a potential NPE in `target.constructor.name`. 
2. Also please update the tests to show in which cases the behavior changes / should change. 
3. Please describe in which cases the names should change. For all automatically named action? For actions with a custom name etc?
4. For a PR to be accepted, please make sure the CI build succeeds.

iotch:
1 and 2. I've commented earlier that I'm not sure the action name should be manipulated in decorator directly, the better place is debug tools, but they need to have the access to the decorated function name. So treat this PR as a quick demonstration. I don't think it should be merged even if I fix the null check and tests.
3. This is applies only to class functions when custom decorator name is not specified.

mweststrate:
Closing PR for inactivity / unclarity what should be done with

Issue #847, opened 2017-02-22
Generic IMapChange


hediet:
Even though IMapChange is generic, it does not really use its type parameter:
```ts
export interface IMapChange<T> {
	object: ObservableMap<T>;
	type: "update" | "add" | "delete";
	name: string;
	newValue?: any;
	oldValue?: any;
}
```
(https://github.com/mobxjs/mobx/blob/master/src/types/observablemap.ts#L44)

Is there a reason why `newValue` and `oldValue` aren't of type T?

BTW.: I really like mobx and I couldn't imagine how to get along without it! Thanks for your effort!

mweststrate:
@hediet nope, I think that is just a missing type! feel free to submit a PR

mweststrate:
Note that the type would be ` T | undefined`, or to nail it precisely down, the interface could be splitted up further (for `"add"` oldValue is always `undefined` and newValue is `T` and never `undefined`, etc)

tcrognon:
How about something like this?

```ts
type IMapChange<K, T extends keyof IMapChangeTypeValueLookup<K>> = {
    object: ObservableMap<K>;
    type: T;
    name: string;
} & IMapChangeTypeValueLookup<K>[T];

interface IMapChangeTypeValueLookup<K> {
    update: {
        oldValue: K;
        newValue: K;
    };
    add: {
        newValue: K;
    };
    delete: {
        oldValue: K;
    };
}
```

hediet:
Yeah, `keyof` rules :D 
However, `keyof` does not play very well with type inference:
```typescript
const c = null as IMapChange<any, "delete" | "add" | "update">;
if (c.type === "add") {
    c. // does not suggest newValue
}
```

This is why I would suggest to make it more explicit:
```typescript
export type IMapChange<T> = IMapChangeUpdate<T> | IMapChangeAdd<T> | IMapChangeDelete<T>;

interface IMapChangeBase<T> {
  object: ObservableMap<T>;
  name: string;
}

export interface IMapChangeUpdate<T> extends IMapChangeBase<T> {
	type: "update";
	newValue: T;
	oldValue: T;
}

export interface IMapChangeAdd<T> extends IMapChangeBase<T> {
	type: "add";
	newValue: T;
}

export interface IMapChangeDelete<T>  extends IMapChangeBase<T> {
	type: "delete";
	oldValue: T;
}
```

mweststrate:
@hediet yep I think that is the way to go. Interested in submitting a PR?

[...]

PR #877, opened 2017-03-07
Stricter typing of IMapChange and IMapWillChange

hediet:
This PR addresses #847 and splits `IMapChange` into `IMapChangeUpdate`, `IMapChangeAdd` and `IMapChangeDelete`.
The type of `IMapWillChange` is also updated so that its `newValue` field is of type `T` (and because the field is optional its type is implicitly `T|undefined`).

### DIFF
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -37,19 +37,34 @@ export type IMapEntry<V> = [string, V];
 export type IMapEntries<V> = IMapEntry<V>[];
 
 // In 3.0, change to IObjectMapChange
-export interface IMapChange<T> {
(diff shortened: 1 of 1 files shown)


hediet:
I don't get that `.build/mobx.ts(3101,22): error TS4081: Exported type alias 'IMapChange' has or is using private name 'IMapChangeUpdate'` error: Why is `IMapChangeUpdate` private? Everything is exported!

mweststrate:
@hediet ah, that is because the source files are preprocessed before the final build, so that only things explicitly mentioned in mobx.ts will be exposed to the public (basically to achieve what is package protected in a language like java). Will fix it during the merge. Thanks!

PR #879, opened 2017-03-08
Expose ES2015 modules

### LINKED ISSUE #868: Compile to ES2015 modules and expose it at  jsnext and module in package.json
> What problem would it solve for you?

It helps shaking off unused MobX code when building for production 

> Do you think others will benefit from this change as well and it should in core package (see also mobx-utils)?

Rollup and Webpack users will get free tree shaking benefits 

> Are you willing to (attempt) a PR yourself?

Yes, all we need to do is to tell tsc to compile again with modules set to ES2015 and update package.json to have a "module" filed.

@developit and @Rich-Harris has a lot of experience with this

mohsen1:
Fixes #868

PR checklist:

* [ ] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "3.1.5",
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
+  "module": "lib/es2015/mobx.js",
+  "jsnext:main": "lib/es2015/mobx.js",
+  "umd:main": "lib/mobx.umd.js",
   "typings": "lib/mobx.d.ts",
   "scripts": {
     "test": "npm run quick-build && npm run tape",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "3.1.5",
+  "version": "3.1.6",
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
   "module": "lib/es2015/mobx.js",
--- a/src/api/autorun.ts
+++ b/src/api/autorun.ts
@@ -98,7 +98,7 @@ export function when(arg1: any, arg2: any, arg3?: any, arg4?: any) {
 		if (predicate.call(scope)) {
 			r.dispose();
 			const prevUntracked = untrackedStart();
-			effect.call(scope);
+			(effect as any).call(scope);
 			untrackedEnd(prevUntracked);
 		}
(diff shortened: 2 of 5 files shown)


src/api/autorun.ts
```diff
@@ -98,7 +98,7 @@ export function when(arg1: any, arg2: any, arg3?: any, arg4?: any) {
 		if (predicate.call(scope)) {
 			r.dispose();
 			const prevUntracked = untrackedStart();
-			effect.call(scope);
+			(effect as any).call(scope);
```
mohsen1: These need to go. Had to do this get build working for now. TSC is not happy about current state of master when targeting ES6:


```
.build/mobx.ts(318,4): error TS2684: The 'this' context of type 'Lambda' is not assignable to method's 'this' of type 'Function'.
.build/mobx.ts(4127,10): error TS2684: The 'this' context of type 'Lambda' is not assignable to method's 'this' of type 'Function'.
  Property 'name' is optional in type 'Lambda' but required in type 'Function'.

```

mohsen1:
@mweststrate is this good to merge? Also can you figure out why when targeting ES6 TSC fails?

package.json
```diff
@@ -1,8 +1,11 @@
 {
   "name": "mobx",
-  "version": "3.1.5",
+  "version": "3.1.6",
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
+  "module": "lib/es2015/mobx.js",
+  "jsnext:main": "lib/es2015/mobx.js",
```
capaj: Do we really need `jsnext:main`? Rollup certainly prefers `module` instead of `jsnext:main`: https://github.com/rollup/rollup/wiki/jsnext:main

Is there a tool which relies on `jsnext:main` and doesn't work with `module`?

mohsen1: Also, it doesn't hurt...

capaj: I know it's just one extra line, but IMHO there's no reason to promote this proprietary, obsolete package.json property.

This only leads to people getting confused.

mohsen1: 🤷🏼‍♀️ 

https://github.com/search?q=%22jsnext%3Amain%22&type=Code&utf8=%E2%9C%93

developit: Slightly older versions of rollup use this.

PR #910, opened 2017-03-28
Fix heading syntax in README.md

pine:
Hi. I've fixed README.md syntax. Thanks!

## before
screen shot 2017-03-29 at 1 31 22 am

## after
screen shot 2017-03-29 at 1 31 05 am


mattruby:
Thanks!

PR #912, opened 2017-03-29
Fix missing `this` in code example

buunguyen:
Fix code example:

 ```javascript
 @observer class MyComponent extends React.component {
     author;
     constructor(props) {
         super(props)
         this.author = props.message.author;
     }
 
     render() {
        return <div>{author.name}</div> // ---->   should be this.author.name
     }
 }
 ```


mattruby:
Thanks!

Issue #916, opened 2017-03-30
Abnormal behavior when access computed value and modify it at same reaction 


fedeoo:
I have a Question: 
Chrome 56, Mobx version 3.1.7
```
const store = observable({
  a: 3,
  b: 4,
  get sum() {
    return this.a + this.b;
  },
});
autorun(() => {
  console.log('sum is', store.sum);
  store.b = 5;
});
store.a = 6;
```
The above code print the result only once. 

I doubt that there is a bug.

andykog:
@fedeoo, nice catch! Working on it.

mweststrate:
Released fix as `3.1.8`.

PR #923, opened 2017-04-01
Fix: new observed derivations become stale

andykog:
See #916


mweststrate:
Great stuff @andykog! Will merge

PR #930, opened 2017-04-04
Fix MobX flow typings for when

jamsea:
A user can pass a debug name to a `when` function that isn't reflected in the flow typings.


jamsea:
@mweststrate let me know if there's anything I need to change to get this merged in! :)

mweststrate:
@jamsea sorry, somehow I created a review but didn't submit it :) (still getting used to that new workflow thing). Anyway:

```
declare function when(nameOrFunction: string | () => boolean, effect: Lambda, scope?: any): any
```

is not entirely correct, as if the first arg is a string, then the third arg should be a function as well. Although the above would work, I think it would be nicer to define it as overload, just as in TS. Afaik that is supported nowadays in Flow?

```
declare function when(name: string, cond:() => boolean, effect: Lambda, scope?: any): any
declare function when(cond:() => boolean, effect: Lambda, scope?: any): any
```

mweststrate:
Thanks!

Issue #943, opened 2017-04-10
Custom comparator for reactions


yixunx:
I have a*:

1. [x] Question: 
I am using some objects generated by Scala.js as a field in the `data` object in a `reaction`. The Scala.js objects contain circular references, causing `deepequal` to fail. Is it possible to pass a custom comparator to replace `deepequal`, or is there any way to achieve something similar in effect?

jamiewinder:
I asked something much the same in #802. It's not possible at the moment, but @mweststrate has indicated he'd be happy with a custom comparator being an option. If I manage to claw some free time from somewhere, I'll give it a go.

mweststrate:
Version 3.2.1 added support for custom equality checks. Thanks @jamiewinder!

Issue #802, opened 2017-01-26
Value-like equality in computed


jamiewinder:
I have the following code:

```typescript
import { LocalDate, Instant } from 'js-joda';
import { computed } from 'mobx';
import { now } from 'mobx-utils';

const currentDate = computed(() => {
    return LocalDate.ofInstant(Instant.ofEpochMilli(now(10000)));
});
```
This basically produces an observable `LocalDate` value based on mobx-utils' `now` utility which will be accurate enough for my use (never more than 10 seconds out). However, the computed itself will observably change every 10 seconds, despite the practical 'value' of the returned object very rarely changing. 

This for example will log every 10 seconds, not ~ once a day:
```typescript
autorun(() => {
  document.write(currentDate.get().toString());
});
```

Is there a better way I could be doing this? It'd be nice to be able to instruct MobX how to determine equality of the return values (i.e. `(a, b) => a.equalTo(b)`) so it knows when the computed value should observably change.

EDIT: This appears to achieve what I'm after, but is a little hacky:
```typescript
let lastDate = null;
const currentDate = computed(() => {
  const nextDate = LocalDate.ofInstant(Instant.ofEpochMilli(now(10000)));
  if (!nextDate.equals(lastDate)) {
    lastDate = nextDate;
  }
  return lastDate;
});
```

mweststrate:
@jamiewinder did you try `computed.struct`?

urugator:
Ability to provide own comparator has been already proposed, but it's probably not needed that often...
Another option would be the ability to implement [Mobx.Symbol.equals] on arbitrary objects. 

Some other workarounds: 
1) In the end autorun uses comperable string, so you could introduce another computed   
```javascript
const currentDate = computed(() => {
  const nextDate = LocalDate.ofInstant(Instant.ofEpochMilli(now(10000)));
});
const currentDateAsString = computed(() => {
  return currentDate.get().toString(); // or timestamp or anything comparable by value/reference/structure
});
autorun(() => {
  document.write(currentDateAsString);
  // it's a bit hacky but you can still access currentDate without subscribing via Mobx.untracked 
});
```

2) Use other observable and update it when needed via reaction
```javascript
const boxedLocalDate = Mobx.observable.box(LocalDate.ofInstant(Instant.ofEpochMilli(Date.now()));
reaction(() => MobxUtils.now(10000), epochMilli => {  
  const newDate = LocalDate.ofInstant(Instant.ofEpochMilli(epochMilli));  
[...]

PR #951, opened 2017-04-15
Computed / Reaction comparer

jamiewinder:
Addresses #802 and #943

Allows a custom comparer to be specified on computed values and properties. Useful for situations where the inbuilt identity and structural comparisons aren't sufficient. 

This should be completely backwards compatible with the current major version.

Still to do:
* [x] Update ComputedValue JSDoc
* [x] Change createComputedDecorator to accept equals comparer rather than boolean 'struct'
* [x] Add `@computed.equals(equals: IEqualsComparer)` 

PR checklist:
* [x] Added unit tests
* [ ] Updated changelog


jamiewinder:
WIP because of a couple of build / test issues which I'd appreciate help with:

I was hoping to change the signature of `createComputedDecorator` from:

```
function createComputedDecorator(compareStructural: boolean)
```
to
```
function createComputedDecorator(equals: IEqualsComparer<any>)
```
and therefore change the following lines:
```
const computedDecorator = createComputedDecorator(false);
const computedStructDecorator = createComputedDecorator(true);
```
to...
```
const computedDecorator = createComputedDecorator(defaultComparer);
const computedStructDecorator = createComputedDecorator(structuralComparer);
```
However, I can't because for some reason they're `undefined` at this point; I have to access them late for the values to be set. I think this is because of the custom build script which is calling the above lines before the `defaultComparer` and `structuralComparer` variables are set. 

Any advice on what the problem here might be?

Also, I want to add a function similar to `@computed.struct` which will allow the custom comparer to be specified for class properties. Something similar to:

```typescript
const isSameid = (a, b) => a.id === b.id;

class Foo {
     @computed.equals(isSameid ) get selected() {
        ...
     }
}
```
I'll do this once I've solved my build problem, but any comments on the name would be appreciated.

I'll sort the docs, tests, etc. once the above are sorted. Any comments welcome!

jamiewinder:
I started writing some tests for this, and come up with something I'd like input on:

Do we think that I should hide away the change detection for the first value, and from/to CaughtException such that:
- `(no value) -> value` is considered a change, does not call the comparer
- `value -> CaughtException` is always considered a change, does not call the comparer
- `CaughtException -> value` is always considered a change, does not call the comparer
- `CaughtException -> CaughtException` is always considered a change, does not call the comparer

As such, there won't be any 'surprising' values being passed to the comparer function; if you have a basic computed property like so:

```typescript
get time() {
    return { hour: this.hour, minute: this.minute }
}
```
I'll *know* that if I add an `.equals` comparer to this, it can be written simply as:
```typescript
const sameTime = (a, b) => a.hour === b.hour && a.minute === b.minute;
```
i.e. i don't have to check if `a` or `b` is undefined, or a CaughtException.

mweststrate:
@jamiewinder sorry for not having time earlier to look into this! Imho `(no value) -> value` can be considered always a change indeed, but only if you can express this to only happen for the first run. As `undefined -> value` should call the comparer generically speaking.

Circular references in the build process are annoying indeed. I want to move to rollup in the near feature, as that hopefully fixes without blowing the bundle size. In the mean time, if it is still an issue, just let me know and I'll investigate it. Moving imports to top / bottom, or using functions instead of constants (as functions get hoisted but constant assignments not) can often help

jamiewinder:
No problem at all, we're all busy! 

The behaviour of the `equals` comparer is currently just that. A comparer along the lines of `(a, b) => a.id === b.id` doesn't (necessarily) need to guard against the arguments being undefined as it'll only be called with values returned from the computed property, or the reaction expression.

I've included some tests which I believe covers this for both computed-s and reaction-s. 

As far as I'm aware, the only things missing is the documentation, which I can do if we're happy with everything else.

Thanks.

mweststrate:
@jamiewinder this PR looks great, let's proceed with it!

Could you take at the the following points:
* [ ] A unit test where `computed` with comparer is used, without using decorators, (e.g. extendObservable({ prop: computed(expr, getter, setter, comparer})`
* [ ] Docs & changelog updates
* [ ] I think it is nice to not expose the compare functions on top level, but at a namespace to group them together? Suggestion: `compare.structural`, `compare.identity`, `compare.default`?

Thanks a lot so far! I think this is a great PR :)

jamiewinder:
I've made the code changes (updated to master, moved comparers to a namespace, non-decorator tests). I can't seem to figure out why the CI is now failing? 
```
TypeError: Object prototype may only be an Object or null: undefined
    at setPrototypeOf (native)
    at r (C:\Projects\GitHub\mobx\lib\mobx.js:2:560)
```
I'll keep looking, but can you spot anything obvious?

mweststrate:
This means that rollup didn't properly resolve ciruclar dependencies, and
some inheritance is failing because the base class is not defined. Usually
this needs to be fixed by moving some imports a round, for example to the
bottom of a file


Op do 29 jun. 2017 om 11:47 schreef jamiewinder <notifications@github.com>:

jamiewinder:
Thanks for the tip! Seems to be sorted. Only thing left is documentation, which I'll try to update soon.

jamiewinder:
Documentation in separate pull request #1084

mweststrate:
Released as 3.2.1. Thanks!

PR #1022, opened 2017-05-28
[DOCS] Fix examples so they can be copy-pasted

danawoodman:
It isn't obvious for beginners to use the guide and not know what the imports should look like. 

Also, the `<RenderCounter />` component was missing from the example making copy-pasting throw an error. For people new to React/MobX, they may assume that component is builtin somehow so putting in the example makes it more explicit.

### DIFF
--- a/getting-started.html
+++ b/getting-started.html
@@ -161,6 +161,8 @@ console.log(todoStore.report());
                             object:
                         </p>
                         <textarea spellcheck="false" class="prettyprint" id="code3" rows="8">
+import mobx, { observable, computed } from 'mobx'
+
 class ObservableTodoStore {
 	@observable todos = [];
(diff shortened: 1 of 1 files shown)


mweststrate:
Are the examples still runnable with these changes?

danawoodman:
I'm not sure, how can I run the examples locally @mweststrate ?

mweststrate:
Sorry, overlooked the arlier comment @danawoodman. You can run the examples by simply pressing the Run buttons below each piece of code :) The first ones don't have one (but they are run as part of the later examples). An alternative is to put the imports in comments so that they don't break the running

mweststrate:
Closing issue; main goal is to get the examples running on the page itself

PR #1023, opened 2017-05-29
remove single-file-build.js, add build.js based on rollupjs

rossipedia:
Addresses #1009 

Currently this depends on a branch I have (PR submitted) for rollup-plugin-typescript2 which enables .d.ts support, so that it can be tested and verified until the PR is merged.

`npm test` passes

PR #1025, opened 2017-05-29
fix typescript tests

rossipedia:
The `tape` package depends on electron, which depends on `@types/node`. This caused a resolution conflict between node's declaration of `require()` and the one in `test/typescript/require.d.ts`.

Also, since `@types/node` has a different return type signature for `setInterval` than the one in the `dom` lib, we need to add a `as any` cast to force the correct overload is resolved.


rossipedia:
Closing. This will be fixed by #1023

PR #1047, opened 2017-06-14
add regression tests for webpack 1 and 2

rossipedia:
I'm not exactly sure where, but @mweststrate asked in a comment if anybody would up for adding some tests to verify that building with webpack didn't break in the future.

This is my attempt at doing so. It adds a `webpack-regression-tests.js` file to `scripts/`, and a corresponding `test-webpack` npm script which runs a few builds using v1 and v2 (using both `commonjs` and `es2015` modules) of webpack. It then executes the resulting bundle and verifies that the output is what's expected.

There's probably a lot more that could be done here, but this should catch the majority of webpack breakages before they make it to `npm`

Comments and suggestions welcome!


mweststrate:
Thanks!

Issue #1043, opened 2017-06-10
"export 'default' (imported as 'mobx') was not found in 'mobx'


themre:
Installed version 3.1.14 and I get:
`"export 'default' (imported as 'mobx') was not found in 'mobx'`
whereas with 3.1.10 everything works fine.

I'm using https://github.com/timarney/react-app-rewired

rossipedia:
how are you consuming mobx? are 

* `import * as mobx from 'mobx';`
* `import mobx from 'mobx';`?

rossipedia:
mobx's module build (which it seems CRA uses by default) does not have a default export. There's some hacky code in there (that should probably be removed at this point - thoughts @mweststrate?) to mimic a default export, but that's more for CommonJS code that's expecting a `default` property on the module.

For ES2015 modules (default in Webpack 2 and CRA) you should import what you need from mobx specifically:

```js
import { observable, extendObservable } from mobx;
```

or import everything and give it an alias:

```js
import * as mobx from 'mobx';

// use mobx.observable, mobx.extendObservable, etc...
```

mweststrate:
I think we should now mimic the hack, until it is removed in 4.0? This
keeps the commonjs and module version as close as possible to each other?
Did you actually run into this, or were you testing?

Op za 10 jun. 2017 om 21:31 schreef Bryan J. Ross <notifications@github.com

rossipedia:
The trick is that ES2015 modules need to be static, and problem here is we don't have a `export default` in the module build. That's going to prevent module consumers from using a default import with it. To support _both_ scenarios, we'd need to have both:

```js
export {
  extras,
  Reaction,
  untracked,
  IDerivationState,
  Atom,
  BaseAtom,
  useStrict,
  isStrictModeEnabled,
  spy,
  asReference,
  asFlat,
  asStructure,
  asMap,
  isModifierDescriptor,
  isObservableObject,
  isObservableValue as isBoxedObservable,
  isObservableArray,
  ObservableMap,
  isObservableMap,
  map,
  transaction,
  observable,
  IObservableFactories,
  computed,
  isObservable,
  isComputed,
  extendObservable,
  extendShallowObservable,
  observe,
  intercept,
  autorun,
  autorunAsync,
  when,
  reaction,
  action,
  isAction,
  runInAction,
  expr,
  toJS,
  createTransformer,
  whyRun,
  isArrayLike
};
```

_and_

```js
export default {
  extras,
  Reaction,
  untracked,
[...]

PR #1050, opened 2017-06-16
Explicit default export

andykog:
See #1043,  https://github.com/mobxjs/mobx-react-devtools/issues/64 ...

### DIFF
--- a/src/mobx.ts
+++ b/src/mobx.ts
@@ -104,22 +104,99 @@ export const extras = {
 	setReactionScheduler
 };
 
+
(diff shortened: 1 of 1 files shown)


rossipedia:
Thanks for this, I _really_ didn't want to do the grunt work there, so I appreciate it that you took the effort :)

mweststrate:
Thanks, excellent idea with the warning @andykog !

mweststrate:
Released as 3.1.16

PR #1052, opened 2017-06-19
[DOCS] Update references `asMap` to `observable.map`

caesarsol:
Update documentation: invite users to use `observable.map` instead of the old `asMap` modifier.

Let me know if there is other sections referring to legacy modifiers that you would like to see updated!


mweststrate:
Thanks! This was outdated indeed

PR #1053, opened 2017-06-19
Add links to awesome list

hawkins:
Hey all,

Per my work with @mweststrate , I've added links to the `gh-pages` docs to the newly-revitalized official awesome list.

Not included in this PR is a link in the readme, since it looks like that'll reside in `master` branch.

PR #1054, opened 2017-06-19
Add link to awesome mobx to readme

hawkins:
Continuing work from #1053 , I've added a link on the README to the awesome list.

Is this what you were thinking, @mweststrate , or would you like it more prominent?


mweststrate:
@hawkins this is fine. I think there is also an official awesome list badge that we could add as well, in the badges list :)

PR #1057, opened 2017-06-22
new Map()

Ciantic:
Important since ES5 maps are standard

I used the edit this page, but some reason it adds that last redundant edit there no matter what.


mweststrate:
Thanks!

PR #1070, opened 2017-07-02
Add size-limit

ai:
[size-limit](https://github.com/ai/size-limit) is a tool to show how many kilobytes your JS library increases the user bundle. You can set size budget and size-limit will prevent JavaScript libraries bloat on Travis CI.

It already used in Autoprefixer, PostCSS and JSS. For instance, I reduce Logux Client size 2x because I found some unexpected huge dependencies.

@mweststrate what do you think about this tool?

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test-browser-chrome": "npm run small-build && ( browserify test/*.js | tape-run --browser chrome | faucet )",
     "test-browser-safari": "npm run small-build && ( browserify test/*.js -t [ babelify --presets [ es2015 ] ] | tape-run --browser safari | faucet )",
     "test-browser-firefox": "npm run small-build && ( browserify test/*.js  | tape-run --browser firefox | faucet )",
-    "test-travis": "npm run full-test && npm run quick-build && istanbul cover tape test/*.js test/typescript/typescript-tests.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -18,15 +18,16 @@
...
     "test-flow": "node_modules/.bin/flow check",
     "test-webpack": "node scripts/webpack-regression-tests.js",
     "coverage": "npm run quick-build && npm run build-tests && istanbul cover tape test/*.js test/typescript/typescript-tests.js",
     "build-tests": "npm run build-typescript-tests && npm run build-babel-tests",
     "build-typescript-tests": "tsc -p test/typescript",
     "build-babel-tests": "babel test/babel/babel-tests.js -o test/babel-tests.js",
     "use-minified": "cp lib/mobx.min.js lib/mobx.js",
-    "lint": "tslint -c tslint.json src/*.ts src/types/*.ts src/api/*.ts src/core/*.ts src/utils/*.ts"
+    "lint": "tslint -c tslint.json src/*.ts src/types/*.ts src/api/*.ts src/core/*.ts src/utils/*.ts",
+    "size": "size-limit 19KB lib/mobx.js"
```
mweststrate: the 19KB is quite arbitrary, why not 20? I assume the goal is to check whether the size is not unexpectedly blown up?

ai: Sure, fixed.

mweststrate:
Merged, thanks!

Issue #1072, opened 2017-07-03
Exception in TS if @observable has no initializer


farwayer:
If you try to observe (intercept, ...) property without initial value before first access then you will get exception. I think we need to add `runLazyInitializers` to TS also. Or even rewrite logic to make init more consistency between TS and JS. It's really bit hacky now.

```typescript
import {observable, observe} from 'mobx'

class User {
  @observable
  loginCount;
}

const user = new User();
observe(user, 'loginCount', () => {});
```

```
tsc --experimentalDecorators test.ts
```

```js
/mnt/data/dev/react/mobx-decorators/node_modules/mobx/lib/mobx.js:2211
        throw new Error("[mobx] Invariant failed: " + message + (thing ? " in '" + thing + "'" : ""));
        ^

Error: [mobx] Invariant failed: Cannot obtain atom from [object Object]
    at invariant (/mnt/data/dev/react/mobx-decorators/node_modules/mobx/lib/mobx.js:2211:15)
    at fail (/mnt/data/dev/react/mobx-decorators/node_modules/mobx/lib/mobx.js:2206:5)
    at getAtom (/mnt/data/dev/react/mobx-decorators/node_modules/mobx/lib/mobx.js:3129:12)
    at getAdministration (/mnt/data/dev/react/mobx-decorators/node_modules/mobx/lib/mobx.js:3134:34)
    at observeObservableProperty (/mnt/data/dev/react/mobx-decorators/node_modules/mobx/lib/mobx.js:3178:12)
    at Object.observe (/mnt/data/dev/react/mobx-decorators/node_modules/mobx/lib/mobx.js:3170:16)
    at Object.<anonymous> (/mnt/data/dev/react/mobx-decorators/test.js:19:8)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
```

mweststrate:
Can you set up a reproduction? Ideally a PR with an additional typescript unit test?

mweststrate:
Released fix as `3.1.17`

PR #1077, opened 2017-07-04
ts test: #1072 @observable without initial value and observe before first access

farwayer:
test for #1072


mweststrate:
Awesome, thanks!

mweststrate:
Merged indirectly through #1081

PR #1080, opened 2017-07-05
Improve typings of intercept method of observable array

bvanreeven:
The `intercept` method of `ObservableArray` was using the types
`IArrayChange` and `IArraySplice`, but should have used the types
`IArrayWillChange` and `IArrayWillSplice` instead, as there are less
properties available when intercepting a change than when observing one.


mweststrate:
Thanks! Note that `intercept` should always return a change or `null`, fixed that in: https://github.com/mobxjs/mobx/commit/1690b9ca1371054a0b8790e01e2828f17cd1e66c

mweststrate:
Released as 3.1.17

PR #1083, opened 2017-07-06
Introduced `runInSandbox` and added warnings when multiple instances of mobx are in a project.

mweststrate:
See #1082


src/core/globalstate.ts
```diff
@@ -88,7 +88,34 @@ export class MobXGlobals {
...
+{
+	const global = getGlobal();
+	if (!global.__mobxInstanceCount) {
+		global.__mobxInstanceCount = 1;
+	} else {
+		global.__mobxInstanceCount++;
+		setTimeout(() => {
+			if (!shareGlobalStateCalled && !warnedAboutMultipleInstances && !_runInSandbox) {
+				warnedAboutMultipleInstances = true;
+				console.warn("[mobx] Warning: there are multiple mobx active instances active. This might lead to unexpected results. See https://github.com/mobxjs/mobx/issues/1082 for details.")
```
andykog: > active instances active 

typo :-)

andykog:
Looks good

src/core/globalstate.ts
```diff
@@ -88,7 +88,34 @@ export class MobXGlobals {
...
+			}
+		})
+	}
+}
+
+export function runInSandbox() {
```
urugator: Maybe reconsider the name, there is no concept of actual sandbox and there is `runInAction`, so someone may (I would) expect something more like this: 
```javascript
runInSandbox(() => {
  // runs isolated
} /*, sandbox/global/namespace/whatever*/)
```
`supressMultipleInstancesWarning` perhaps?

src/core/globalstate.ts
```diff
@@ -88,7 +88,34 @@ export class MobXGlobals {
 
 export let globalState: MobXGlobals = new MobXGlobals();
 
+let shareGlobalStateCalled = false;
+let warnedAboutMultipleInstances = false;
+let _runInSandbox = false;
```
urugator: For the sake of consistency, could `_runInSandbox` be named `runInSandboxCalled` just like `shareGlobalStateCalled`...?

src/core/globalstate.ts
mweststrate: Renamed to it to **isolateGlobalState**, which is a clearer counterpart to **shareGlobalState**

src/core/globalstate.ts
mweststrate: Renamed to `runInIsolationCalled`

src/core/globalstate.ts
mweststrate: Fixed :)

src/core/globalstate.ts
urugator: I don't want to nitpick, but was it meant to be `isolateGlobalStateCalled`?

Issue #1087, opened 2017-07-07
Expand createTransformer to support multiple arbitrary arguments in the transformer


SamPruden:
Currently, `createTransformer` has some major limitations, such as:

- It only allows a single argument
- That argument must be either a string, number, or object - null, undefined, boolean, etc throw errors
- It doesn't respect `this`, so works poorly to transform class methods
- The documentation lacks details about these limitations

I think it would be beneficial to expand the functionality to remove these limitations, such that any arbitrary function could be transformed and memoized.

## Feature wishlist for `createTransformer`:

- Allow any arbitrary type to be used as arguments in the transformer
- Allow any number (zero or more) arguments to be used in the transformer
- Allow the value of `this` to pass through, such that a method can be a transformer and still have access to its class members
- Allow constructors to become transformers, such that using `new` respects memoization
- Provide a decorator to make methods transformers - either `@transformer` or maybe `@computed` on methods
- MAYBE provide some shortcut for memoizing constructors - a class decorator or a `transformer.makeFactory(Klass)` approach perhaps, but there are some design and technical challenges surrounding this point

I currently have a WIP PR for this at #1088. I've been having fun putting this stuff in and I think the quality of the work is pretty good for now, but it could use some stress testing (perhaps in the form of more unit tests) by a more experienced javascript developer as I'm cautious about claiming it supports all types (there's one acknowledged limit to that already in frozen objects) and doesn't break some backwards compatibility somewhere.

TheOtherSamP:
I've edited this issue to tidy it up a lot, taken out some fluff and made things clearer.

mweststrate:
`createTransformer` will move to the `mobx-utils` package from Mobx 4 and upward. Closing the issue here

PR #1088, opened 2017-07-08
WIP: Extending createTransformer to allow multiple arbitrary arguments in the transformer

SamPruden:
I made an issue for this at #1087. That hasn't gotten any response yet and I know this may not be accepted, but I wanted to have a play with it to help familiarise myself with mobx anyway.

## Stuff to do
- [x] Allow any arbitrary type to be used as arguments in the transformer
- [x] Allow any number (zero or more) arguments to be used in the transformer
- [x] Allow the value of `this` to pass through, such that a method can be a transformer and still have access to its class members
- [x] Allow constructors to become transformers, such that using `new` respects memoization
- [x] Provide a `@transformer` decorator for methods
- [ ] Provide `transformer.global` function (needs a better name) to use a default global cache, calling `transformer.global(f)` multiple times with the same `f` should share a cache, see below
- [ ] Maybe allow `@transformer` to be applied directly to classes
- [ ] Maybe merge `@transformer` functionality into `@computed`
- [ ] Maybe rename the transformer concept to better reflect the generalised concept
- [ ] Document changes
- [ ] Work out if there are more places tests need to be added - maybe Babel? Performance?

**Bug of note**: Currently it throws an error if one of the arguments to the transformer is not extensible. To the best of my knowledge, this is not fixable without ES6 features, specifically Map. An argument could be made for this not throwing, and just silently not memoizing instead.

### `transformer.global`
Probably implemented as some version of:
```TS
transformer.global = createTransformer(createTransformer).bind(transformer);
```
Calling `createTransformer` directly creates a unique local cache, calling `createTransformer.global` shares a cache among all global calls. This is useful when interfacing with third-party libraries that don't expose their functions as transformers. `createTranformer` works transparently with constructors, so this also provides a way to memoize constructor calls without creating an explicit boilerplate factory.

## PR to do
* [x] Added unit tests
* [ ] Updated changelog

I'm not a particularly experienced javascript/TS developer, so I'm uncomfortable making hard claims about this working with any arbitrary object and not breaking backwards compatibility. To the best of my knowledge it does both of those things, but it could use some more stressful unit tests from someone who might know some weird edge cases.

### DIFF
--- a/src/api/createtransformer.ts
+++ b/src/api/createtransformer.ts
@@ -22,7 +22,8 @@ export function createTransformer<F extends (...args: any[]) => R, R>(transforme
 	return (function (...objects: any[]) {
 		// Limit the length to that of the transformer
 		// Without this, use in functions like .map() break because of extra unwanted arguments
-		objects = objects.slice(0, transformer.length);
+		// this is added at the start so that createTransformer can be run on prototypes
+		const memoizeObjects = [this, ...objects.slice(0, transformer.length)];
 
 		if (resetId !== globalState.resetId) {
 			objectCache = {};
@@ -30,7 +31,7 @@ export function createTransformer<F extends (...args: any[]) => R, R>(transforme
 			resetId = globalState.resetId;
 		}
 
-		const identifier = getMemoizationId(symbolCache, objects);
+		const identifier = getMemoizationId(symbolCache, memoizeObjects);
 		let reactiveTransformer = objectCache[identifier];
 		if (reactiveTransformer)
 			return reactiveTransformer.get();
--- a/src/api/createtransformer.ts
+++ b/src/api/createtransformer.ts
@@ -44,7 +44,18 @@ export function createTransformer<A, B>(transformer: ITransformer<A, B>, onClean
 	};
 }
 
-function getMemoizationId(object): string {
+function getMemoizationId(...objects: any[]): string {
+	// Get a key for each arg
+    const keys = objects.map(getObjectMemoizationId);
+
+    // Get the lengths of each key
+    const keyLengths = keys.map(key => key.length);
+
+    // The key lengths are added to the front so that f("a", "b") !== f("ab")
+    return `${keyLengths.join(",")}:${keys.join("")}`;
+}
+
+function getObjectMemoizationId(object): string {
 	switch (typeof object) {
 		case "symbol": throw new Error("Symbols are not supported as createTransformer arguments.");
 		case "undefined": return "undefined";
@@ -60,7 +71,7 @@ function getMemoizationId(object): string {
         default: {
             if (object === null) return "null";
 
-			let tid = object.$transformId;
+			let tid = object.$transformId as number;
 			if (tid === undefined) {
 				tid = getNextId();
 				addHiddenProp(object, "$transformId", tid);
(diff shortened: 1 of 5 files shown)


mweststrate:
Thanks for the extensive PR. Just a heads up: It's gonna take some time before I have time to study, agree and review an extensive PR for this (probably couple of weeks, want to focus on releasing MST first), if there is some real hurry for any reason, let me know! 

On a side note, making createTransformer becomes a lot easier if you use `WeakMap`, which will be allowed in MobX 4 (as the runtime requirements will be upped). If that enables more items on your wishlist, that might be something to consider.

TheOtherSamP:
@mweststrate No worries, and there're still some kinks to work out with this for sure. Actually, WeakMap doesn't help too much in this case as I'm serialising everything to string to support multiple arguments. I guess with WeakMap we could move to some form of tree structure for the memoizing, but I suspect that's less desirable than this single flat lookup. Where it _would_ be helpful is in replacing the `$transformId` that's added to track object IDs because it's cleaner not to mutate tracked objects, and currently this can't support non extensible objects. `$transformId` is already used by the existing `createTransformer` implementation though.

Access to ES6 features in general may be of significant value, I'm yet to fully investigate `Proxy` and `Reflect` but if they become available they may tidy up a number of problems. I'm currently trying to work out what to do about a nasty problem with rest parameters not being serialised because they're not indicated in `length`, I don't think there's any way to detect them.

TheOtherSamP:
If I were to post updates on this with details of tricky (design or technical) problems and things, would you prefer those go here in the PR or in the feature request issue?

mweststrate:
PR preferably :)

Op do 13 jul. 2017 om 16:50 schreef TheOtherSamP <notifications@github.com>:

TheOtherSamP:
I thought I'd write up a bit of a progress-report on how this has gone so far, as much for my own records as anything else. I'm working in a fairly agile-esque manner, just playing around with some rough ideas and goals and letting those develop as I go. 

### Arbitrary argument types
I started off by extending `createTransformer` to handle arbitrary argument types. This was mostly fairly easy, I'm just taking any object and mapping it to a unique string key. The main challenges here are mapping references (but the existing implementation already handled them) and mapping `Symbol`. Symbol is dealt with by caching symbols to IDs, with access to `(Weak)Map` the same could be done for arbitrary objects.

### Arbitrary argument counts
Next I extended `createTransformer` to handle arbitrary numbers of arguments. The basic mechanism for this is a function that takes an array of arbitrary objects, and produces a unique string key by combining the keys of each object as described above. As far as I know, that also works completely fine.

However, the naive usage (just memoizing based on `arguments`) of this posed a major problem. Many functions that might be used as transformers are used as callbacks and passed more arguments than they use. The example of this that got thrown up by a unit test is `Array.prototype.map`. I only wanted to memoize based on the item, but the transformer was returning different results if the _index_ of some object changed, even if the object remained the same, as the index is a second parameter to the callback.

My naive solution to this, and how the code currently stands, was to only memoize up to `transformer.length` arguments. That seemed like a simple and elegant solution, but now (as of ES6) has a weird problem. Rest parameters are not detectable, and contribute zero to length. As such, they are completely ignored by the memoization. The only way I know to resolve this cleanly is with function decompilation and parsing, but I suspect that's something mobx does not want to be doing.

The remaining option seems to be to introduce additional parameters that specify desired behaviour. This does have the potential advantage of allowing comparison modes (ref, struct, etc) to be specified for each parameter. I suspect this is what I'll be forced to do, but I fear it may be a frequent cause of bugs if rest parameters are silently ignored.

mweststrate:
Thanks for the write @SamPruden. Now closing this PR until somebody champions further improvements on it :). I noticed in that in general using view models and associate them with models to 'capture' the memoization is easier to grasp for people although it will involve more boilerplate

PR #1099, opened 2017-07-15
Another couple of spelling fixes

realityforge:


src/core/reaction.ts
```diff
@@ -147,9 +147,10 @@ export class Reaction implements IDerivation, IReactionPublic {
 	dispose() {
 		if (!this.isDisposed) {
 			this.isDisposed = true;
+			// if disposed while running, clean up later. Maybe not optimal, but rare case
```
capaj: @realityforge can you move this into the if block? so one line down? I know it's petty little thing, but I think it's nicer when the comment about the if branch is actually inside the if branch itself not above it

realityforge: Done

Issue #1060, opened 2017-06-23
Consider using Prettier


roperzh:
**What problem would it solve for you?**

In some places (especially tests files) the code is a bit hard to read because of formatting issues like bad indentation. [Prettier](https://prettier.github.io/prettier) is an opinionated code formatter, and can help us here.

Also, @mweststrate mentioned in https://github.com/mobxjs/mobx/issues/1035#issuecomment-307141578 the will to implement Prettier in the near future.

**Do you think others will benefit from this change as well and it should in core package (see also mobx-utils)?**

Yes, but it will not make a difference in the core package.

**Are you willing to (attempt) a PR yourself?**

Yes, but I'm not sure if I'm the appropriate person.

mweststrate:
Prettier is a welcome addition :). Can you make sure the same settings as in mobx-state-tree are used?

realityforge:
The pull request #1103 was raised to satisfy this issue.

realityforge:
This can be closed now that #1103 has been merged

PR #1103, opened 2017-07-19
Introduce prettier to keep the format of the codebase consistent

realityforge:
This PR attempts to address issue #1060.

In short it introduces the `prettier` and `lint-staged` dependencies and setups the same tasks and configuration as in [mobxjs/mobx-state-tree](https://github.com/mobxjs/mobx-state-tree). The prettier was also run across the codebase to ensure it aligns with the new codestyle. 

It should be notes that the codestyle in `mobx-state-tree` is not the same as in this codebase and thus this commit introduces a lot of changes. If you prefer I could align the rules used by prettier to more closely match this repository to minimize the amount of changes in the pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "module": "lib/mobx.module.js",
   "typings": "lib/mobx.d.ts",
   "scripts": {
+    "prettier": "prettier --write --print-width 100 --tab-width 4 --no-semi \"**/*.js\" \"**/*.jsx\" \"**/*.tsx\" \"**/*.ts\"",
     "test": "npm run quick-build && npm run tape",
     "full-test": "npm run small-build && npm run build-tests && npm run use-minified && npm run tape && node --expose-gc test/perf/index.js && npm run test-flow && node test/mixed-versions/mixed-versions.js && npm run test-webpack",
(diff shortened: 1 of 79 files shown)


package.json
```diff
@@ -88,5 +92,11 @@
     "functional-reactive-programming",
     "state management",
     "data flow"
-  ]
-}
\ No newline at end of file
+  ],
+  "lint-staged": {
+    "*.{ts,tsx,js,jsx}": [
+      "prettier --write --print-width 100 --tab-width 4 --no-semi",
```
capaj: @mweststrate are you okay with no semi and tab with 4?

mweststrate:
Yep! Sorry for not responding earlier

Op ma 7 aug. 2017 10:18 schreef Jiri Spac <notifications@github.com>:

capaj:
awesome. Since this builds fine, I'll merge it.

phiresky:
Am I missing something or is there a dependency to `husky` missing for this to run as a pre-commit script?

capaj:
@phiresky yeah we're missing husky. We have `precommit` script, but that won't run without husky. I'll open a PR.

PR #1112, opened 2017-07-26
fix access to global object (fixes #1109)

### LINKED ISSUE #1109: Mobx not getting imported and bundled by Rollup
I'm trying to use Mobx in an existing project using Rollup, Babel, Uglify to bundle and transpile.  I import like this:
```javascript
import {observable, autorun} from 'mobx'
```
Normally that will get included in the resulting bundle. Yet nothing gets included and in the browser I get the error that Mobx is undefined. When building, there is no error about not finding the imports. They simply never get included by Rollup in the final bundle. 

I'm using Rollup with Gulp as a gulp plugin: gulp-better-rollup: 1.1.1

I have Mobx version 3.2.1 installed from NPM.

rossipedia:
webpack UMD builds pass in the global object (`global` in node and `window` in the browser) as a parameter named `global`. In RollupJS this is not the case. Access to the global object has to be handled manually, so let's just check for `window`, and if it doesn't exist fall back to `global`.

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -3,7 +3,7 @@ Object.freeze(EMPTY_ARRAY);
 
 declare var global;
 export function getGlobal() {
-	return global;
+	return typeof window !== 'undefined' ? window : global;
 }
 
 export interface Lambda {


mweststrate:
Thanks!

PR #1119, opened 2017-07-29
Use Index Type for property name in intercept

smithgeek:
Use an index type for the property name argument when using intercept for better type checking.


capaj:
@smithgeek sorry I've merged codestyle change which affected the whole codebase. Would you be as kind and rebase your change? LGTM so I'd merge it when you do.

mweststrate:
Thanks!

mweststrate:
Merged as part of 3.3.0

PR #1124, opened 2017-08-08
Update create-transformer.md

chestozo:
It is good to mention that `A` should be an observable value.
Not sure about `B`.

### DIFF
--- a/docs/refguide/create-transformer.md
+++ b/docs/refguide/create-transformer.md
@@ -2,7 +2,7 @@
 
 `createTransformer<A, B>(transformation: (value: A) => B, onCleanup?: (result: B, value?: A) => void): (value: A) => B`
 
-`createTransformer` turns a function (that should transform one value into another value) into a reactive and memoizing function.
+`createTransformer` turns a function (that should transform observable value `A` into another value `B`) into a reactive and memoizing function.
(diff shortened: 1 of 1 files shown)


chestozo:
I had a look into the source code and it looks like `A` and `B` - are general js objects.

Hmm
Somehow I could not make `createTransformer` memoize results.
For example:
```js
const trans = createTransformer((foo) => {
    console.log('SHOULD RUN ONCE FOR EACH foo');
    return foo.bar;
});

const foo = { bar: 'baz' };
console.log(trans(foo));
console.log(trans(foo));
```
here I am expecting to see `SHOULD RUN ONCE FOR EACH foo` only once.
But I see it twice in console:
```
SHOULD RUN ONCE FOR EACH foo
baz
SHOULD RUN ONCE FOR EACH foo
baz
```
What I am doing wrong?

mweststrate:
That behavior is correct, to quote https://mobx.js.org/refguide/create-transformer.html:

> Always use transformations inside a reaction like @observer or autorun. Transformations will, like any other computed value, fall back to lazy evaluation if not observed by something, which sort of defeats their purpose.

So, this should produce the correct behavior: 

```
autorun(()=> {
  console.log(trans(foo));
  console.log(trans(foo));
})
```

mweststrate:
Removed the word `observable` and merged

PR #1125, opened 2017-08-10
Fix example to use `this`

danawoodman:
I believe this is a typo?


capaj:
@danawoodman indeed it is

PR #1131, opened 2017-08-14
Rename ESM output to .mjs

geelen:
I've just been playing around with the [@std/esm](https://medium.com/web-on-the-edge/es-modules-in-node-today-32cff914e4b) loader, which makes reference to the fact that Node's ES6 module support is going to be tied to the file extension `*.mjs`. I admit I don't know much context around this issue but I saw this and thought it might make sense to follow the new convention.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
   "umd:main": "lib/mobx.umd.js",
-  "module": "lib/mobx.module.js",
+  "module": "lib/mobx.mjs",
   "typings": "lib/mobx.d.ts",
   "scripts": {
     "prettier": "prettier --write --print-width 100 --tab-width 4 --no-semi \"**/*.js\" \"**/*.jsx\" \"**/*.tsx\" \"**/*.ts\"",
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -100,7 +100,7 @@ function build() {
 
         generateBundledModule(
             path.resolve(".build.es", "mobx.js"),
-            path.resolve("lib", "mobx.module.js"),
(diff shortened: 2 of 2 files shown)


mweststrate:
@rossipedia could you review this one?

rossipedia:
@geelen there's another minor change you'll need to make: [`pkg.module`](https://github.com/mobxjs/mobx/blob/master/package.json#L7) would also need to change to `mobx.mjs`.

I'm not 100% on the ramifications of this change at the moment, either. While _ideally_ moving to `.mjs` is the Right Thing™ to do, and at least Webpack handles that extension currently, I'm not sure how many other bundlers / ecosystems would get tripped up by this. Although, to be fair, we gotta start _somewhere_. As far as my personal opinion goes, though, I'm all for it 👍

capaj:
I personally dislike `mjs`, but it is the standard for node.js and browsers don't really care for it. So makes sense to have `mjs` instead of `js`.

mweststrate:
Thanks & merged. Will be part of next release. Probably blowing up in somebodies face like any module / bundling related change ever did in the past :cry:

mweststrate:
Can somebody confirm this will work on React Native as expected as well?

mweststrate:
React native does not support `.mjs` extension yet. Also, webpack doesn't support it yet (will be added in 4). See: https://github.com/webpack/webpack/issues/5686

Rolled back this PR and created a fresh follow up issue: https://github.com/mobxjs/mobx/issues/1172

geelen:
Also, since I posted this PR, Node has gotten [experimental support for modules](http://2ality.com/2017/09/native-esm-node.html) which... I don't think should change things but who knows!

> Probably blowing up in somebodies face like any module / bundling related change ever did in the past

That's my experience too. Good luck!

PR #1141, opened 2017-08-17
Replace '@state' with '@observable state'

alex-at-work:


mattruby:
Thanks for catching that!

Issue #1113, opened 2017-07-27
Why `componentWillUpdate` method will be invoked when I use observable properties in React Component?


MrErHu:
Document say:

>The reactive state will be picked up by render but will not explicitly invoke other React lifecycle methods like componentShouldUpdate or componentWillUpdate. 

but  when I run the code below:

```javascript
@observer
class Timer extends React.Component {
    @observable secondsPassed = 0

    componentWillMount() {
        setInterval(() => {
            this.secondsPassed++
        }, 1000)
    }


    componentWillUpdate(){
        console.log('componentWillUpdate')
    }

    componentDidUpdate(){
        console.log('componentDidUpdate')
    }

    render() {
        return (<span>Seconds passed: { this.secondsPassed } </span> )
    }
}
```

The console logs the message:

```
componentWillUpdate
componentDidUpdate
```
That confuses me.

mweststrate:
Yes, but the cause of these hooks firing is that a re-rendering is
triggered by MobX, as a result of seconds being used in render. And these
hooks are part of the normal render process. If you would use
`this.secondsPassed` in the lifecycle hooks, but *not* in the render
function, they wouldn't trigger.

Op do 27 jul. 2017 om 03:36 schreef 请叫我王磊同学(Wanglei) <
notifications@github.com>:

OleksandrSachuk:
Create PL according to this issue https://github.com/mobxjs/mobx/pull/1149

PR #1149, opened 2017-08-23
Clarify docs about React lifecycles hooks #GoodnessSquadKyiv

OleksandrSachuk:
According to issue #1113 (https://github.com/mobxjs/mobx/issues/1113) create an explanation like this: "The reactive state will be picked up by render but will not explicitly invoke other React lifecycle methods except componentWillUpdate and component did update. If you need other React lifecycle methods, just use the normal React state based APIs."


mweststrate:
Cool! Thanks!

PR #1163, opened 2017-09-11
Support weex

CodeFalling:
## About this PR
[Weex](https://github.com/alibaba/weex) is a framework like `ReactNative`. It runs javascript in JavaScriptCore/V8 runtime and freezed `Array.prototype`.

Mobx is a awesome library so I want make it works in weex. `ObservableArray` in mobx override `push/sort/etc` could not run in weex. We'd better make this original array method writeable in `StubArray.prorotype`.

See https://github.com/alibaba/weex/pull/1529

## PR checklist:

* [ ] Added unit tests
* [ ] Updated changelog
* [X] Added typescript typings
* [X] Verified that there is no significant performance drop (`npm run perf`)

ezgif-4-d59f90397f

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -119,6 +119,29 @@ function inherit(ctor, proto) {
 }
 inherit(StubArray, Array.prototype)
 
+// Weex freeze Array.prototype
+// https://github.com/alibaba/weex/pull/1529
+;[
+	"constructor",
+    "push",
+    "shift",
+    "concat",
+    "pop",
+    "unshift",
+    "replace",
+	"find",
+	"findIndex",
+	"splice",
+	"reverse",
+	"sort"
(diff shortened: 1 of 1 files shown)


src/types/observablearray.ts
```diff
@@ -119,6 +119,29 @@ function inherit(ctor, proto) {
 }
 inherit(StubArray, Array.prototype)
 
+// Weex freeze Array.prototype
```
mweststrate: Thanks for the PR.

Could you make the patch conditional? That is, only apply the patch if the properties are actually not writable (such as in the weex case)? This avoids putting more methods in the prototype chain when it is not really needed.

CodeFalling:
Done. Now the patch only works when `Array.prototype`  has been freezen.

PR #1179, opened 2017-09-22
Ignore IDEA's project files

realityforge:
Ignore some IDE project files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ docs/_book
 docs/node_modules
 refguide/*.md
 .DS_Store
+*.iml
+*.ipr
+*.iws
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ refguide/*.md
 *.iml
 *.ipr
 *.iws
+/.idea


krnaveen14:
It should also include .idea/ folder

VolCh:
Why not simple use [.git/info/exclude file](https://help.github.com/articles/ignoring-files/#explicit-repository-excludes) or [git global (your user wide) ignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) ? There are a lot of IDEs/editors in the world and add all theirs file to repo seems not reasonable.

realityforge:
The convention for other mobx projects seems to be to ignore IDE files. (i.e. See master branches .gitignore at  https://github.com/mobxjs/mobx/blob/master/.gitignore). 

Generally you can not ignore IDE files globally as there is sometimes good reasons for committing them in. So rather than doing it per checkout you can just apply the rule to the repository. It is not uncommon across many github projects. See for example the search https://github.com/search?q=.iml&type=Code&utf8=%E2%9C%93

realityforge:
Added the additional ignore for directory based idea projects

capaj:
I have to agree with @VolCh. It doesn't seem like we can add all the folders for all the editors out there. IMHO this ignore should be placed in your global gitignore @realityforge.
It would make sense to remove it from the master as well. We should only ignore files that are part of this project. Not everyone's editor working folder.

mweststrate:
Besides that this is no generic solution indeed, it isn't really in the way either. If this is a pragmatic solution that lowers the boundary of people to contribute / submit correct PR's I can live with that.

PR #1187, opened 2017-10-01
Fix #1120: isComputed should return false for non existing property

### LINKED ISSUE #1120: isComputed should return `false` for non existing property
`isComputed(observable({}), "x")`  currently throws, but should return `false` (`isComputed({}, "x")` does behave correct)

DyingAlbatross:
Fix for #1120: isComputed now returns false for non existing properties.

PR checklist:

* [x] Added unit tests


mweststrate:
Great! Thanks for fixing :)

PR #1202, opened 2017-10-13
Updating docs (api.md, computed-decorator.md)

kamilmielnik:
1. Updated description of `observable` rules for type of received argument which contradicted the [source code](https://github.com/mobxjs/mobx/blob/a867e87788a3aeb5d03a0a99f8379c13edd6be25/src/utils/utils.ts#L78)
2. Few typos/missing words


mweststrate:
Thanks!

PR #1211, opened 2017-10-19
Fix failing tests when running `npm test`

alex-deas:
Added generation of umd files through browserify which are required when running `npm
test`


mweststrate:
Thanks! In the end this problem has been solved slightly different (with reusing rollup instead of browserify) but the issue has been addressed now :)

PR #1213, opened 2017-10-19
Don't pass undefined to custom equality checks on becoming observed a…

### LINKED ISSUE #1208: Non-value passed to custom computed equality check
The documentation for `computed`'s `equals` option states

> This acts as a comparison function for comparing the previous value with the next value. If this function considers the previous and next values to be equal, then observers will not be re-evaluated.

I interpreted this to mean that custom equality functions will only ever be passed values that were returned from the derivation function. Is that correct? I believe I've found a situation where this isn't the case.

A snippet that reproduces the behaviour:

```ts
const mobx = require('mobx');
const observable = mobx.observable(0);
const computed = mobx.computed(() => observable.get() + 1, {
  equals: (old, new_) => {
    console.log(`old=${old}, new_=${new_}`);
    return old === new_;
  },
});

// Logs nothing
const dispose = mobx.autorun(() => computed.get());
// Logs "old=1, new_=2"
observable.set(1);
dispose();
// Logs "old=undefined, new_=2"
mobx.autorun(() => computed.get());
```

`undefined` was never returned from the derivation function `() => observable.get() + 1`, so I didn't expect it to be passed to the equality function. After disposing the observer and establishing a new one, I either expected the equality function not to be called, or to be called with 2 and 2.

I think this is happening because `ComputedValue::value` is assigned `undefined` when it becomes unobserved ([code](https://github.com/mobxjs/mobx/blob/ea6c823a5e6871e4f5352ed3db216003887c60b3/src/core/computedvalue.ts#L113)) and this is then passed to `equals` when it becomes observed again ([code](https://github.com/mobxjs/mobx/blob/ea6c823a5e6871e4f5352ed3db216003887c60b3/src/core/computedvalue.ts#L174-L180)). Perhaps it should be assigned `new CaughtException(null)` as it is in the constructor ([code](https://github.com/mobxjs/mobx/blob/ea6c823a5e6871e4f5352ed3db216003887c60b3/src/core/computedvalue.ts#L78)) which would bypass the equality check ([code](https://github.com/mobxjs/mobx/blob/ea6c823a5e6871e4f5352ed3db216003887c60b3/src/core/computedvalue.ts#L177))?

chrisdoble:
…gain

This PR attempts to fix #1208.

I've extended one of the tests added in #951 to cover this bug, but I'm not sure if that's sufficient. The difference between `test/babel/*.js`, `test/typescript/*.js`, and `test/*.js` isn't clear to me so please let me know if I should add more tests.

PR checklist:

* [x] Added unit tests

### DIFF
--- a/test/errorhandling.js
+++ b/test/errorhandling.js
@@ -682,7 +682,7 @@ test("peeking inside autorun doesn't bork (global) state", t => {
         t.equal(b.diffValue, 0)
         t.equal(b.lowestObserverState, 0)
         t.equal(b.unboundDepsCount, 1)
-        t.equal(b.value, undefined)
+        t.notEqual(b.value, 3)
         t.equal(b.isComputing, false)
 
         t.equal(c.dependenciesState, -1)
--- a/src/core/computedvalue.ts
+++ b/src/core/computedvalue.ts
@@ -110,7 +110,7 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
 
     onBecomeUnobserved() {
         clearObserving(this)
-        this.value = undefined
+        this.value = new CaughtException(null)
     }
 
     /**
--- a/test/observables.js
+++ b/test/observables.js
@@ -1853,7 +1853,7 @@ test("computed equals function only invoked when necessary", t => {
     )
 
     const values = []
-    const disposeAutorun = mobx.autorun(() => values.push(combinedToLowerCase.get()))
+    let disposeAutorun = mobx.autorun(() => values.push(combinedToLowerCase.get()))
 
     // No comparison should be made on the first value
     t.deepEqual(comparisons, [])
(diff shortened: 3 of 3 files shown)


test/errorhandling.js
```diff
@@ -682,7 +682,7 @@ test("peeking inside autorun doesn't bork (global) state", t => {
         t.equal(b.diffValue, 0)
         t.equal(b.lowestObserverState, 0)
         t.equal(b.unboundDepsCount, 1)
-        t.equal(b.value, undefined)
+        t.notEqual(b.value, 3)
```
chrisdoble: I'm not particularly happy with this assertion. I wanted to write something like `t.ok(isCaughtException(b.value))` but as far as I can tell neither `CaughtException` nor `isCaughtException` are exported from `mobx.ts`. I considered exporting them from `mobx.ts`, but that's changing public API. Also considered importing them from `core/derivation.ts` directly, but that's inconsistent with all other tests.

This feels like it's still achieving the goal of the assertion (ensuring that the cached value is cleared) while avoiding those problems.

I'm open to suggestions of other approaches.

mweststrate:
Made a minor improvement and merged. Thanks!

PR #1216, opened 2017-10-20
Update extendObservableHelper for Typescript 2.6

### LINKED ISSUE #1215: mobx fails to compile with Typescript 2.6
In Typescript 2.6, Object.getOwnPropertyDescriptor now returns `PropertyDescriptor | undefined`. extendobservable.ts uses `getOwnPropertyDescriptor` in a safe way, but the compiler can't tell, so it fails to compile.

```ts
            if (definedProps[key] !== true && hasOwnProperty(propSet, key)) {
                definedProps[key] = true
                if ((target as any) === propSet && !isPropertyConfigurable(target, key)) continue
                const descriptor = Object.getOwnPropertyDescriptor(propSet, key)
                defineObservablePropertyFromDescriptor(adm, key, descriptor, defaultEnhancer)
                                                                 ~~~~~~~~~~
                                                                 Error, descriptor might be undefined.
            }
```

The fix is just to change the use of `descriptor` to `descriptor!`.

sandersn:
In Typescript 2.6, Object.getOwnPropertyDescriptor now returns `PropertyDescriptor | undefined`. Assert that the type is still just `PropertyDescriptor`, because it is checked with `hasOwnProperty` just above.

Fixes #1215

### DIFF
--- a/src/api/extendobservable.ts
+++ b/src/api/extendobservable.ts
@@ -46,7 +46,7 @@ export function extendObservableHelper(
                 definedProps[key] = true
                 if ((target as any) === propSet && !isPropertyConfigurable(target, key)) continue // see #111, skip non-configurable or non-writable props for `observable(object)`.
                 const descriptor = Object.getOwnPropertyDescriptor(propSet, key)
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks!

PR #1218, opened 2017-10-22
observable-decorator.md doc fix

SheetJSDev:
The example also uses `computed` but didn't import the symbol


mweststrate:
Thanks!

PR #1231, opened 2017-11-05
improve flow typings for observable

Kriegslustig:
This is basically just copied from the typescript definition and slightly modified to match the flow syntax.


Kriegslustig:
CI failure because of auto cancellation. Can't really do anything about this. The tests pass locally.

> This job was cancelled because the "Auto Cancellation" feature is currently enabled, and a more recent build (#1232) for pull request #1231 came in while this job was waiting to be processed.

mweststrate:
Thanks a lot for your effort!

Issue #1243, opened 2017-11-15
ObservableMap replace triggers a change on unchanged values


elado:
ObservableMap `.replace` uses `.clear`, which causes reactions on all values in the map, even if they haven't changed.

```js
import { observable, computed, autorun } from 'mobx'

class Store {
  m = observable.map({ a: 1, b: 2, c: 3 })

  @computed
  get a() {
    console.log('a recompute')
    return this.m.get('a')
  }
}

const store = new Store()

autorun(() => {
  console.log(store.a)
})

console.log('changing b...')

store.m.set('b', 4)

console.log('setting a to same value...')
// won't recompute a
store.m.set('a', 1)

console.log('replacing...')

// store.a will be recomputed even when it wasn't changed
store.m.replace({ a: 1, d: 5 })
```

https://codesandbox.io/s/61kwwnjvlr

I would expect it to remove missing keys instead of `clear` and use `merge` on the rest, and then it won't trigger changes on unchanged values.

https://github.com/mobxjs/mobx/blob/master/src/types/observablemap.ts#L305-L307

mweststrate:
Good catch! Interested in create a PR?

Op wo 15 nov. 2017 om 23:15 schreef Elad Ossadon <notifications@github.com>:

elado:
yes! will do in the next few days.

mweststrate:
Awesome! Thanks in advance

Op vr 17 nov. 2017 om 19:52 schreef Elad Ossadon <notifications@github.com>:

elado:
https://github.com/mobxjs/mobx/pull/1247

PR #1247, opened 2017-11-19
Map: Trigger reactions only on changed values

elado:
Fixes https://github.com/mobxjs/mobx/issues/1243


mweststrate:
Great PR, thanks!

PR #1249, opened 2017-11-21
make deep NaN === deep NaN for computed values when using compareStructural

Ulmanb:
There seemed to be an inconsistent behavior where NaN === NaN but deep NaN !== NaN.
I think @benjamingr talked with you about it. 

PR checklist:

* [X] Added unit tests
* [X] Verified that there is no significant performance drop (`npm run perf`) - it does seem to add around 1-3 second on my machine, not sure if this is significant.


benjamingr:
@mweststrate this is the one we discussed.

mweststrate:
Thanks for noticing & PR, @Ulmanb  & @benjamingr!

Issue #1232, opened 2017-11-05
Improve Flow Typed Support


Kriegslustig:
Flow has this `.js.flow` file pattern. If utilized correctly, a consumer won't have to change anything in their `.flowconfig` in order to get the flow typings for `mobx`.

We'd have to remove the wrapping `module` block from the `.js.flow` file and add `export` to all declared variables that are exported. I've already done this locally to test it, so I could immediately open a PR.

Projects using MobX with flow typings would have to remove the `[libs]` line from their config when they upgrade.

mweststrate:
The PR has been merged, however if I understand it correctly it does not solve the issue described above? (Properly exposing the types). Would you be interested in submitting a PR for that as well?

Kriegslustig:
Correct. Sure thing! I'll submit a PR.

mweststrate:
Released as 3.4.0

PR #1254, opened 2017-11-25
correctly expose flow typings

Kriegslustig:
With this PR consumers can use the Flow type definitions without needing any configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Next
+
+* Improve Flow support support by exposing typings regularly. Flow will automatically include them now. In your `.flowconfig` will have to remove the import in the `[libs]` section (as it's done [here](https://github.com/mobxjs/mobx/pull/1254#issuecomment-348926416)). Fixes [#1232](https://github.com/mobxjs/mobx/issues/1232).
+
 # 3.3.2
 
 * Fix bug where custom comparers could be invoked with `undefined` values. Fixes [#1208](https://github.com/mobxjs/mobx/issues/1208)
--- a/README.md
+++ b/README.md
@@ -300,11 +300,7 @@ And finally kudos for all the people that believed in, tried, validated and even
 * Use `npm test` to run the basic test suite, `npm run coverage` for the test suite with coverage and `npm run perf` for the performance tests.
 
 ## Flow support
-MobX ships with flow typings. To use them in your project, add this to the `[libs]` section of your `.flowconfig`:
-```
-[libs]
-node_modules/mobx/lib/mobx.js.flow
-```
+MobX ships with flow typings. Flow will automatically include them. Though you cannot ignore `node_modules` in your `.flowconfig` and a you **cannot** import it explicitly in the `[libs]` section.
 
 ## Bower support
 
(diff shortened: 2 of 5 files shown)


Kriegslustig:
Apparently references don't work in titles... #1232

mweststrate:
@Kriegslustig just double check: you verified this will work when installing the package and import stuff using `import { x } from "mobx"`?

Kriegslustig:
Yes, this works. With one caveat though, consumers will need to remove MobX from flow config `[libs]` section. I will update the README accordingly.

```
--- a/.flowconfig
+++ b/.flowconfig
@@ -5,7 +5,6 @@
 [include]
 
 [libs]
-./node_modules/mobx/lib/mobx.js.flow
 /flow-typed
 
 [options]
```

Kriegslustig:
I documented it in the README. Should I also write something into the changelog?

mweststrate:
@Kriegslustig yep, please put it in the changelog.md as well :)

Kriegslustig:
done 👍

mweststrate:
Thanks! Merged and will release soon as 3.4.0

zhaocnus:
@Kriegslustig, @mweststrate:

I think a lot of people might accidentally update MobX from v3.3.3 to v3.4.0 (by just removing and reinstalling all the node_modules). If they do that but they still use MobX's flow type definition based on the old [README](https://github.com/mobxjs/mobx/tree/c6dd0d852157fcebfaafc2d9ac119eb6a7f6c65f#flow-support), they will get a ton of flow errors. 

I have a blog post to explain this: https://github.com/zhaocnus/mobx-flow-typed-test

Maybe we need to provide more info in the readme?

Kriegslustig:
@zhaocnus have you read the [changelog](https://github.com/mobxjs/mobx/blob/master/CHANGELOG.md#340)? Maybe we can extend this information, if it's not sufficient.

zhaocnus:
@Kriegslustig Yes I think extending the information will be helpful (especially for people who are not an expert in flow). But it not necessary. 

I did check the changelog but I wish I have seen it earlier :) 

This is how I encountered the issue:

1. I did a refresh install of all my node modules (removed node_modules folder and package-lock.json, and run `npm i`). I have `"mobx": "^3.2.2",` in my package.json. `npm i` installed mobx 3.4.1. 
2. I still have `node_modules/mobx/lib/mobx.js.flow` in `.flowconfig`

So this exposes all the mobx type definitions to the global scope, and flow throws a lot of errors. It took me a while to figure the issue. I hope I could have checked the README and changelog earlier. 

So I figure adding some more info might save other people's time later.

mweststrate:
PR for changelog is welcome! (and tip: make sure that removing node_modules
doesn't upgrade your modules unintentionally. Use yarn, npm lock files or
npm shrinkwrap, without them it is impossible to create reliable /
reproducible builds for your project)

Op vr 29 dec. 2017 om 09:56 schreef Luca Nils Schmid <
notifications@github.com>:

zhaocnus:
@mweststrate Thanks for the tip! I've learned my lesson. I will create a PR.

Issue #1238, opened 2017-11-13
Issue: Cannot assign to read only property 'length' of function


dannsam:
We have a problem happening with 92 users (it's way less than 1% of our users)

The error is `Cannot assign to read only property 'length' of function`
`at IObservableFactories line 1935, column 30 (webpack:///./node_modules/mobx/lib/mobx.js:1935)`
![image](https://user-images.githubusercontent.com/12251341/32761537-cd06391a-c958-11e7-84f5-046f041ef787.png)

It maps to line 195 in the following snippet
https://github.com/mobxjs/mobx/blob/b7fef45ebd7120730b3bdb13ff50b13412ebf03d/src/api/observable.ts#L191-L195

Somehow, `IObservableFactories.prototype` have a property called `length`. Attempting to assign `observable['length']` fails in strict mode, because `length` is a read-only property.

I found no way to reproduce it. I am really keen on blaming browser extensions, but this issue happens across multiple browsers:

4 browsers affected | %
-- | --
Chrome | 61%
Mobile Safari | 35%
IE | 2%
Edge | 2%

I guess the only idea I have is be wrap the code assigning props into `try{} catch(e){}` or to check filter out non-writable properties.

I don't mind creating a PR, but it'd be nice to know what you guys think and what approach is preferable before I do it

dannsam:
The PR I posted is not going to solve the issue, we've just had another error report 
![image](https://user-images.githubusercontent.com/12251341/32816659-b865fa1a-ca1e-11e7-8e36-ddb4b400b866.png)

Going to investigate deeper

dannsam:
Weird, `IObservableFactories.prototype` must have a value of some kind of rubbish object but it does not make sense how that value gets assigned...

mweststrate:
@dannsam I would be fine by introducing a simple workaround, that filters out `length` as well, or just make everythign explicit, kill the loop, and copy the relevant functions by hand. Interested in changing your PR to that if that solves the issue?

dannsam:
Thanks @mweststrate I'll check it out and change my PR if it solves the issue!

dannsam:
I am going to trial these changes https://github.com/mobxjs/mobx/commit/df5002cadf72d2915c4f81dac99c42d0ad6334df over the weekend and create a PR if the issue is resolved

dannsam:
[...]

PR #1257, opened 2017-11-26
Split `IObservableFactories` type definition and implementation

dannsam:
It seems the only reason to keep `IObservableFactories` as a class is to avoid declaring type definition and implementation separately. However, it does not seem to be entirely correct approach at runtime, because it's never expected that a new instance of `IObservableFactories` will be created. In addition, reading properties of prototype cause issues with some browsers.

Issue #1238


mweststrate:
Merged, thanks!

Issue #1248, opened 2017-11-20
array remove by predicate should be implemented


fehergeri13:
I use this pattern so often:

`myObservableArray.replace(myObservableArray.filter(item => item.bad !== true));`

so I think it should be implemented by default.

`myObservableArray.removeAll(item => item.bad === true)`

mweststrate:
Ehh.. it exists already?

Op ma 20 nov. 2017 om 15:14 schreef Gergely Fehérvári <
notifications@github.com>:

fehergeri13:
Meh, not really. because currently only item match is existing but this is kinda different, since i want to use a predicate

mweststrate:
Feel free to submit a PR :) `removeAll` sounds like a sane name

mweststrate:
Closed, see: https://github.com/mobxjs/mobx/pull/1384#issuecomment-372024393

PR #1260, opened 2017-11-28
Added ObservableArray.removeAll issue #1248

thomasd16:
`myArray.removeAll(predicate)` is logically equivalent to `myArray.replace(myArray.filter(predicate))`.
`.removeAll` has yet to be added to github pages I will submit a second PR later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Next
+
+* Added `ObservableArray.removeAll`, [#1248](https://github.com/mobxjs/mobx/issues/1248)
+
 # 3.3.2
 
 * Fix bug where custom comparers could be invoked with `undefined` values. Fixes [#1208](https://github.com/mobxjs/mobx/issues/1208)
--- a/flow-typed/mobx.js
+++ b/flow-typed/mobx.js
@@ -156,7 +156,10 @@ declare module "mobx" {
             thisArg?: any,
             fromIndex?: number
         ): number,
-        remove(value: T): boolean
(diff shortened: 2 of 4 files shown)


thomasd16:
I think travis-ci broke. The log only shows a strange docker error.

mweststrate:
Correct me if I am reading this wrong, but it seems that you did implement `keepAll` rather then `removeAll`? `removeAll` should remove all the items for which the predicate returns true, instead of keeping them and removing the other ones. 

For examle like in linq: https://msdn.microsoft.com/en-us/library/wdka673a(v=vs.110).aspx

thomasd16:
That is correct my mistake. I mis-read the issue slightly I need to invert the predicate.

I will do that later today. Would you like me to keep the current implementation and re-name it `keepAll`?

mweststrate:
No I don't think that is overkill. Adding a `!` to the predicate should
suffice for people ;-)

Op ma 4 dec. 2017 om 16:07 schreef Thomas Devries <notifications@github.com

mweststrate:
No. I think it is overkill.... Sorry for the confusing formulation :-P

Op ma 4 dec. 2017 om 16:22 schreef Michel Weststrate <mweststrate@gmail.com

mweststrate:
Closed per https://github.com/mobxjs/mobx/pull/1384#issuecomment-372024393

Thanks for the effort! 

But since we can finally towards using proxies, observable arrays should become more 'normal' arrays and we can express these kind of utilities using the standard utility belts like lodash.

Issue #992, opened 2017-05-12
Atom and onBecome(Un)Observed


urugator:
First some simplified self-explanatory code of what I was trying to accomplish:
```javascript
function onBecomeObserved(observable, callback) {
  // Obtain atom from observable
  const atom = Mobx.extras.getAtom(observable);
  // Save original onBecomeObserved
  const onBecomeObserved = atom.onBecomeObserved;
  // Replace onBecomeObserved
  atom.onBecomeObserved = () => {
    onBecomeObserved(); // call original
    callback(); // call additional logic
  }
}

// Usage:
onBecomeObserved(observable, setInterval(/*...*/));
// additionally I would like to provide unobserved version
onBecomeUnobserved(observable, clearInterval(/**/));
// and property versions:
onBecomeObserved(this, "users", setInterval(/*...*/));
onBecomeUnobserved(this, "users", clearInterval(/*...*/));
```

The point is the ability to hook additional resource management to an existing atom, instead of creating another one and synchronizing their subscription/unsubscription through a common getter.
However it doesn't work for multiple reasons, which can be expressed by following questions:

Why `onBecomeObserved` is not defined on `IObservable` interface similary to `onBecomeUnobserved`?

Why `BaseAtom` doesn't provide default implementation for `onBecomeObserved` similary to `onBecomeUnobserved`?  [(note the comments)](https://github.com/mobxjs/mobx/blob/master/src/core/atom.ts#L21)  

How so, that `getAtom(observable("hello")).onBecomeObservedHandler/onBecomeUnobservedHandler` (both) returns undefined? [Shouldn't they be functions?](https://github.com/mobxjs/mobx/blob/master/src/core/atom.ts#L58)

If I replace `onBecomeObserved/onBecomeUnobserved/onBecomeObservedHandler/onBecomeUnobservedHandler` on atom obtained by `getAtom(observable("hello"))` it seems like non of it gets called when `autorun` is created/disposed (replacement is done before the reaction is created).

urugator:
Fiddle presenting the problems: https://jsfiddle.net/wv3yopo0/1212/

jamiewinder:
This could really help with the suggestions I made [here](https://github.com/mobxjs/mobx-utils/issues/55) and [here](https://github.com/mobxjs/mobx-utils/issues/51).

urugator:
I've noticed that `ObservableValue` extends from `BaseAtom`, not from `Atom`, which explains the lack of default `...Handler` callbacks.

[...]

PR #1270, opened 2017-12-11
WIP: create onBecome(Un)observed hooks

mweststrate:
See #992


src/api/become-observed.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	if (typeof orig !== "function")
+		return fail("Not an atom that can be (un)observed")
+
+	atom[hook] = function() {
+		orig.call(this)
+		cb.call(this)
+	}
+	return function() {
+		this[hook] = orig
```
jamiewinder: What if the 'original' hook (at the time) was removed in the meantime? I think this would restore it?

mweststrate:
Closed in favor of #1321, all relevant changes are moved over.

Issue #1282, opened 2017-12-24
Flow Typecheck Error


wadeanthony0100:
Hey, all!

There looks to be a small error in our flow typings definitions.

Having the following code
```js
  import { observable } from 'mobx';
```
and running flow typechecking yeilds the following:

```
node_modules/mobx/lib/mobx.js.flow:273
273:     <T>(IModifierDescriptor<T>): T,
                                ^ Unexpected token <
```

Looking at the source in this repo, I think its just a simple syntax error that _should_ look like this:

```js
<T>(value: IModifierDescriptor<T>): T,
```

That being said, I'm a Flow noob, so I could be missing some big picture details or something. I can open a PR with this one liner change if you guys want.

Thanks!

wadeanthony0100:
Somewhat related: fixing this problem reveals the following:

```
node_modules/mobx/lib/mobx.js.flow:239
239:     oldValue: T | typeof undefined
                              ^^^^^^^^^ typeof-annotation `undefined`. Could not resolve name
```

After fixing this, my project is successfully type checked! I can add this to the PR

mweststrate:
Thanks for the PR! Will ship with the next release.

PR #1283, opened 2017-12-24
Fix flow typing errors

wadeanthony0100:
Addresses issue #1282 

Like I said in the issue, I'm a Flow noob, so feel free to suggest other ways to do this.


mweststrate:
Thanks!

PR #1294, opened 2018-01-02
add example of reaction invoke once

takefumi-yoshii:
As long as I see the src code, I can invoke a reaction only once with this example. Is there any reason for being hidden?

### DIFF
--- a/docs/refguide/reaction.md
+++ b/docs/refguide/reaction.md
@@ -1,6 +1,6 @@
 # Reaction
 
-Usage: `reaction(() => data, data => { sideEffect }, options?)`.
+Usage: `reaction(() => data, (data, reaction) => { sideEffect }, options?)`.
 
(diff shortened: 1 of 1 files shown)


mweststrate:
@takefumi-yoshii not sure whether this is a good example, as `when` does exactly do this. Maybe the example could be a bit more artificial and dispose after 5 times?

takefumi-yoshii:
@mweststrate Thank you for your reply. In some usecase, reaction hope to be only invoked once.With my recognition, `when` will only react bool type value. Is there any way to reaction another type with `when`?

mweststrate:
@takefumi-yoshii ah fair point, the example makes that not really clear though, as `items.length`  seems seems truthy. Maybe you can mention that explicitly?

takefumi-yoshii:
@mweststrate For clarity, I moved the example to another section.

mweststrate:
Thanks!

PR #1305, opened 2018-01-15
Allow autorunAsync to take a debouncing function in addition to a delay

TrentHouliston:
When mobx autorun is used to execute canvas rendering functions it can cause slowdowns as it tries to rerun the rendering function for every model update even though the system can't render yet.
This PR proposes to add a way to fix this by allowing autorunAsync to take a function instead of a delay.
This function (for the case of canvas requestAnimationFrame) would allow mobx to only run the expensive render function once the system was ready to render.

This effect can be seen in the two examples below
 * https://jsbin.com/sineton/2/edit?js,output
 * https://jsbin.com/sineton/1/edit?js,output
The code is rendering 10000 lines in a spinning circle updating on an interval of 1.
The first of these is using autorun as normal and a significant slowdown can be seen compared to how fast it should be spinning.
The second is using requestAnimationFrame to let mobx know when it can update the canvas which gives much better performance.

Before I go and update the docs etc, is this a feature that would be wanted in the mobx codebase?
Are there any suggestions on how this could be improved?

PR checklist:
* [x] Added unit tests
* [x] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 3.5.0
+
+* `autorunAsync` now accepts a scheduler function to allow improved performance for tasks such as rendering to canvas
+
 # 3.4.1
 
 * Republished 3.4.0, because the package update doesn't seem to distibute consistently through yarn / npm
--- a/src/api/autorun.ts
+++ b/src/api/autorun.ts
@@ -120,21 +120,28 @@ export function autorunAsync(
     delay?: number,
     scope?: any
 ): IReactionDisposer
+export function autorunAsync(
+    func: (r: IReactionPublic) => any,
+    scheduler: (func: () => any) => any,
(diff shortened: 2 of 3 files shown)


mweststrate:
@TrentHouliston yeah that makes sense! Could you add unit tests and prepare a PR on the docs?

TrentHouliston:
Is this good @mweststrate?

mweststrate:
@TrentHouliston great, thanks! I'll merge soonish; as I am processing a few other api changes related to autorun, and will process this together with that.

mweststrate:
@TrentHouliston merged into the mobx4 branch. Thanks! https://github.com/mobxjs/mobx/pull/1321/commits/6abdb58a6ec11798c45f37660acea0de157d89c4

PR #1306, opened 2018-01-15
adds husky so that precommit runs when it should

capaj:
I think we forgot to include this in https://github.com/mobxjs/mobx/pull/1103


mweststrate:
Thanks!

Issue #1301, opened 2018-01-09
Should be able to specify reaction onError at initialization


adamabeshouse:
It's not possible to add a custom error handler to an autorun in a way that handles if the error is thrown on initialization and first run.

https://jsfiddle.net/akz7rz1p/1/

A fix would be to be able to specify the error handler at initialization of the reaction.

adamabeshouse:
awesome, thanks!

mweststrate:
Closing as this will be added to mobx4

Issue #1263, opened 2017-12-04
remove bower.json


capaj:
Since bower is deprecated I think it would be a good idea to drop it's support. Probably in the upcoming 4.0 release? What are your thoughts? Anyone will miss having bower.json?

vadim-ch:
Can I try contribute?

mweststrate:
Done so in MobX 4

Issue #1241, opened 2017-11-14
Dev mode which doesn't catch exceptions


NaridaL:
I've read https://github.com/mobxjs/mobx/issues/731, but the resolution isn't very satisfying. The current behavior makes debugging unnecessarily difficult. `extras.onReactionError(e =>{ throw e; });` makes things slightly better but doesn't fix the main issue that the debugger will not stop at the location of the actual error, but rather on the rethrow, which makes it impossible the inspect the values of the problematic stack frames.

Halting on every exception will have one stepping through a bunch of caught exceptions in other frameworks which isn't helpful either.

3. [X] Idea:
  * [X] What problem would it solve for you? ^^
  * [X] Do you think others will benefit from this change as well and it should in core package (see also mobx-utils)? Yes, this would help everyone debugging.
  * [X] Are you willing to (attempt) a PR yourself? If necessary.

mweststrate:
@NaridaL I see why the idea is relevant :). With the limatations of exception handling in JS it is really hard to find a behavior that works optimal for all causes. Feel free to attempt a pr, for example by introducing `extras.disableErrorBoundaries()` (which prints a clear warning that this is a dev / debug only feature)

mweststrate:
Will be implemented in Mobx4. Thanks!

PR #1321, opened 2018-01-25
WIP: Mobx4 

mweststrate:
see #1076, #1316 

# Mobx 4

* [x] remove deprecated stuff
* [x] remove default export
 * [x] Upgrade to latest TS
 * [x] Error messages cleanup
 * [x] Reaction api cleanup
 * [x] Remove most invariants in prod
 * [x] Support Babel 7: See https://github.com/mobxjs/mobx/issues/1352
 * [x] Process TODOs
 * [x] ~~Fix strict type errors~~ skipped, it made the source less comprehensible, due to all the reflection an dynamic scoping of property descriptors etc
 * [x] make observable api more stricter
 * [x] improve spy events, drop custom events
 * [x] double check cleaned up api
 * [x] decorate api
 * [x] ~~extendObservable => extend nope: impactful without a lot of added value~~
 * [x] get / set / delete / keys / values api
 * [x] fix map .values and other iterables
 * [x] deep argument + plain constructors
 * [x] fix onbecomeobservedhooks. see #992
  * [x] ~~kill modifiers? nope; used by many~~
 * [x] move transform to mobx-utils?
 * [x] ~~use boolean flags for all the internal states? Didn't gain enough to be interesting~~
 * [x] add custom error handlers when creating reactions, see https://github.com/mobxjs/mobx/issues/1301
 * [x] update flow typings
 * [x] introduce `legacy_observable`
 * [x] reintroduce deprecated api's with warning & redirect
 * [x] implement #817 
 * [x] implement #800 
 * [x] check code coverage (both for tests and dead code)
 * [x] #1361 ?
 * [x] Flow?
 * [x] clean docs
* [x]  remove moved functions from mobx-utils
 * [ ] release mobx-utils / mobx-react
 * [x] blog post
 * [ ] new sandboxes

# MobX 5

* [ ] Proxies!
* [ ] Remove non-native utilities like `array.remove` and such?
* [ ] Maps & Sets!


CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
...
+* If there are multiple mobx instances active in a single project, an exception will be thrown. Previously only a warning was printed. Fixes #1098. For details, see [#1082](https://github.com/mobxjs/mobx/issues/1082).
+* Dropped the `shareGlobalState` feature. Instead, projects should be setup properly and it is up to the hosting package to make sure that there is only one MobX instance
+* Dropped bower support. Fixes #1263
+* The `spyReportStart`, `spyReportEnd`, `spyReport` and `isSpyEnabled` are no longer public. It is no longer possible to emit custom spy events as to avoid confusing in listeners what the possible set of events is.
+* Dropped `isStrictModeEnabled`
+* `observable(value)` will only succeed if it can turn the value into an observable data structure (a Map, Array or observable object). But it will no longer create an observable box for other values to avoid confusion. Call `observable.box(value)`  explictly in such cases.
```
benjamingr: Does this mean that `@observable foo = 5` won't work anymore and we'd need to `@observable.box foo = 5`?

spion: afaik this is just for the function API

benjamingr: Good, I think `@observable.box foo = 5` would be tedious to do every time since it's a pretty common use case for me to have observable value properties.

mweststrate: @benjamingr correct, it is only for the function version, to make the result more predicatable (previously you would 'suddenly' get a box around your value if the value itself could not be made observable)

benjamingr: Thanks @mweststrate, any update on using proxies for arrays? :D

test/base/autorunAsync.js
```diff
@@ -130,6 +130,94 @@ test("autorunAsync passes Reaction as an argument to view function", function(do
...
+test("autorunAsync accepts a scheduling function", function(done) {
+    var a = m.observable({
+        x: 0,
+        y: 1
+    })
+
+    var autoRunsCalled = 0
+    var schedulingsCalled = 0
+
+    debugger
```
TrentHouliston: I don't think this is supposed to be here?

mweststrate: Correct, removed, thanks!

mweststrate:
Additional issues being fixed in this MR:

#813

mweststrate:
#961

mweststrate:
#992 
#1098 
#1122
#1148 
#1197 
#1241
#1263

CHANGELOG.md
```diff
@@ -5,171 +5,121 @@
 
 This is the extensive list of all changes.
 
-Most noteable changes:
+### New features
 
+The changes mentioned here are discussed in detail in the [release highlights](https://medium.com/p/c1fbc08008da/), or were simply updated in the docs.
```
benjamingr: 🎉

CHANGELOG.md
```diff
@@ -5,171 +5,121 @@
...
+
+* MobX 4 introduces separation between the production and non production build. The production build strips most typechecks, resulting in a faster and smaller build. Make sure to substitute process.env.NODE_ENV = "production" in your build process! If you are using MobX in a react project, you most probably already have set this up. Otherwise, they idea is explained [here](https://reactjs.org/docs/add-react-to-an-existing-app.html).
+* Observable maps are now backed by real ES6 Maps. This means that any value can be used as key now, not just strings and numbers.
+* Introduced `decorate(thing, decorators)` to decorate classes or object without needing decorator syntax.
```
benjamingr: Any reason why this isn't in utils?

CHANGELOG.md
```diff
@@ -5,171 +5,121 @@
...
+* Introduced `onBecomeObserved` and `onBecomeUnobserved`. These API's enable hooking into the observability system and get notified about when an observable starts / stops becoming used. This is great to automaticaly fetch data from external resources, or stop doing so.
+* `computed` / `@computed` now accepts a `requiresReaction` option. If it set, the computed value will throw an exception if it is being read while not being tracked by some reaction.
+* To make `requiresReaction` the default, use `mobx.configure({ computedRequiresReaction: true })`
+* Introduced `mobx.configure({ disableErrorBoundaries })`, for easier debugging of exceptoins. By [NaridaL](https://github.com/NaridaL) through [#1262](https://github.com/mobxjs/mobx/pull/1262)
+* `toJS` now accepts the options: `{ detectCycles?: boolean, exportMapsAsObjects?: boolean }`, both `true` by default
+* Introduced `flow` to create a chain of async actions. This is the same function as [`asyncActions`](https://github.com/mobxjs/mobx-utils#asyncaction) of the mobx-utils package
```
benjamingr: I think that we should figure out how to do this with `async` functions rather than generators

CHANGELOG.md
```diff
@@ -5,171 +5,121 @@
...
+* The flow typings have been updated. Since this is a manual effort, there can be mistakes, so feel free to PR!
+
+* `computed(fn, options?)` / `@computed(options) get fn()` now accept the following options:
+  * `set: (value) => void` to set a custom setter on the computed property
+  * `name: "debug name"`
+  * `equals: fn` the equality value to use for the computed to determine whether it's output has changed. The default is `comparer.default`. Alternatives are `comparer.structural`, `comparer.identity` or just your own comparison function.
+  * `requiresReaction: boolean` see above.
+
+* `autorun(fn, options?)` now accepts the following options:
+  * `delay: number` debounce the autorun with the given amount of milliseconds. This replaces the MobX 3 api `autorunAsync`
```
benjamingr: Why is this a part of autorun rather than just calling a `_.debounce` on an existing autorun?

mweststrate: Good question :) I think the major advantage is that this way the effect is cancellable.

CHANGELOG.md
mweststrate: We cannot automatically wrap the `then` of every promise returned. I think still in general the best practice is to just queue all your changes to the end of an async process.

CHANGELOG.md
mweststrate: Yes, this makes it possible to use MobX in the same when when using decorator syntax, plain objects, prototypes through `Object.create` etc.

src/utils/utils.ts
```diff
@@ -199,6 +181,10 @@ export function toPrimitive(value) {
     return value === null ? null : typeof value === "object" ? "" + value : value
 }
 
+export function isThennable(thing: any): thing is PromiseLike<any> {
```
benjamingr: In general, it's preferable to not check if something is `then`able if suspected but to `Promise.resolve` it.

mweststrate: Lol, committed this in the wrong branch.. :)

PR #1324, opened 2018-01-29
IObservableArray.find() typing fix

outring:
Hi! Currently, `IObservableArray<T>.find` returns `T` instead of `T | undefined` which leads to runtime errors.

I’ve tried to write a test, but it seems that `typescript-tests.ts` is not being typechecked, because even `var x:string = 1` doesn’t fail. Is there any way to test this?


mweststrate:
Thanks!

PR #1340, opened 2018-02-05
Introduce `.trace()` to ease debugging

mweststrate:


src/core/observable.ts
```diff
@@ -224,8 +232,48 @@ export function propagateMaybeChanged(observable: IObservable) {
...
+    )
+    if (derivation.isTracing === TraceMode.BREAK) {
+        const lines = []
+        printDepTree(getDependencyTree(derivation), lines, 1)
+
+        // prettier-ignore
+        new Function(
+`debugger;
+/*
+Tracing '${derivation.name}'
```
eric-wieser, line 256: I've not seen this pattern before; did you copy it from somewhere else (and does it have a name), or is this your own invention?

mweststrate:
Didn't see it before, so the latter :)

Issue #1204, opened 2017-10-15
TypeScript - ObservableMap to Map


eddiegroves:
Just wondering if using an ObservableMap as an es2015 Map is a supported thing? TypeScript is giving a few errors 

IDE: VS Code Version 1.17.0 (1.17.0)
TypeScript: 2.5.3

```
import { ObservableMap, observable } from "mobx";

const observableMap: ObservableMap<string> = observable.map();

const map: Map<string, string> = observableMap;
/*
[ts]
Type 'ObservableMap<string>' is not assignable to type 'Map<string, string>'.
  Property '[Symbol.toStringTag]' is missing in type 'ObservableMap<string>'.
*/

const map2: Map<string, string> = new Map(observableMap)
/*
[ts]
Argument of type 'ObservableMap<string>' is not assignable to parameter of type 'Iterable<[string, string]>'.
  Property '[Symbol.iterator]' is missing in type 'ObservableMap<string>'.
*/

const map3: Map<string, string> = observableMap.toJS();
/*
[ts]
Type 'IKeyValueMap<string>' is not assignable to type 'Map<string, string>'.
  Property '[Symbol.toStringTag]' is missing in type 'IKeyValueMap<string>'.
*/
```

`tsconfig.json` includes:
```
{
  "compilerOptions": {
    "target": "es2015",
    "module": "es2015",
    "lib": ["dom", "es2015", "es2016"]
  }
}
```

mweststrate:
I think we could ass `Symbol.iterator` and `Symbol.toStringTag` support (I think the first might already be there, just not in the typings). 

But not that `observable.map() instanceof Map` will always yield `false` until we use proxies. An alternative could be to introduce a utility function that converts observable maps back to normal `Map`'s (`toJS` will create a plain key/value object from the top of my head).

Alternatively, I think `new Map(theObservableMap)` will probably do the trick as well, although the typings might be in the way

Strate:
There was a very similar issue about Bluebird and it assignable to Promise: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/11027

As @mweststrate says, `instanceof` check will not work, so it is a definetely not a Map. Also, for example, if you add some methods to Map's prototype, and then assign ObservableMap to variable declared as `Map`, and then tries to use that method, you will get an runtime error.

[...]

PR #1361, opened 2018-03-01
WIP: Make observable map structurally match an ES6 Map

marcfallows:
Observable map now has the typings to structurally match `Map`. Also implements `toStringTag` properly as if it were a Map. This should resolve https://github.com/mobxjs/mobx/issues/1204

***

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "quick-build": "tsc --pretty",
     "small-build": "node scripts/build.js",
     "lint": "tslint -c tslint.json src/*.ts src/types/*.ts src/api/*.ts src/core/*.ts src/utils/*.ts",
-    "size": "size-limit 20KB lib/mobx.js",
+    "size": "size-limit 21KB lib/mobx.js",
     "precommit": "lint-staged"
   },
   "repository": {
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -11,7 +11,8 @@ import {
     deprecated,
     isES6Map,
     getMapLikeKeys,
-    fail
+    fail,
+    declareStringTag
 } from "../utils/utils"
 import {
     IInterceptable,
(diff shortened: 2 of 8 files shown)


tsconfig.json
```diff
@@ -13,7 +13,7 @@
         "noImplicitAny": false,
         "moduleResolution": "node",
         "lib": [
-            "es5",
+            "es2015",
```
marcfallows: Had to change this so I could use `Symbol.toStringMap` and `Symbol.iterator` in the typings.

src/types/observablemap.ts
```diff
@@ -40,9 +41,19 @@ export interface IMap<K, V> {
     get(key: K): V | undefined
     has(key: K): boolean
     set(key: K, value?: V): this
+    // prettier-ignore
+    [Symbol.toStringTag]: "Map"; // necessary semicolon
```
marcfallows: prettier gets rid of this semi, but it is needed. Not familiar enough with prettier to know if there's a better way.

(Or maybe there's a better way to let TS know it's the end of the statement, but I don't know it).

package.json
```diff
@@ -22,7 +22,7 @@
     "quick-build": "tsc --pretty",
     "small-build": "node scripts/build.js",
     "lint": "tslint -c tslint.json src/*.ts src/types/*.ts src/api/*.ts src/core/*.ts src/utils/*.ts",
-    "size": "size-limit 20KB lib/mobx.js",
+    "size": "size-limit 21KB lib/mobx.js",
```
marcfallows: Apparently I've bumped the size over the limit by 10 bytes. I don't know if this is the change you would want - but this should get things to pass

mweststrate:
Thanks! Would it ok for you to target the `mobx4` branch in this PR? That avoids a potential breaking change where people have to change their TS setup to be able to type Symbol.

marcfallows:
Ah, I hadn't realised there was a mobx4 branch with some of the TS configuration already changed. I'll have to resolve conflicts before this one is ready to go.

mweststrate:
@marcfallows don't bother, Ill merge your changes, that is propbably

mweststrate:
Merged, thanks!

PR #1380, opened 2018-03-08
WIP: MobX 5

mweststrate:
Nothing to see here yet

* [x] proxy array
* [x] proxy object
* [x] ~ObservableMap extends Map~ Not supported by TS / babel
* [x] disable spy and other dev features in production builds? (for smaller build)
* [x] remove deprecated api's
* [x] optimize `reportObserved` recording using sets?
* [x] ~release `mobx 5` which always throws: pick mobx 4 or 15~ Meh. Just mobx 5 should be fine
* [x] process all open todos
* [x] ~introduce `seal` option?~ `proxy: false` is a new option instead
* [x] Polish changelog
* [x] Update documentation
* [ ] Bump peer deeps in (should be completely backward compatable)
  * [x] mobx-react
  * [x] mobx-utils
  * [ ] mobx-state-tree
  * [ ] notify @andykog / @adamkleingit / @philmander to bump dev-tools / angular binding dependencies


src/mobx.ts
```diff
@@ -16,6 +16,12 @@
  *
  */
 
+if (typeof Proxy === "undefined") {
+    throw new Error(
```
benjamingr: :tada:

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "4.0.0-beta.4",
+  "version": "15.0.0-alpha.1",
```
tkrotoff: Do you mean "5.0.0-alpha.1"? :-)

test/base/array.js
```diff
@@ -295,7 +284,7 @@ test("is array", function() {
     expect(x instanceof Array).toBe(true)
 
     // would be cool if this would return true...
-    expect(Array.isArray(x)).toBe(false)
+    expect(Array.isArray(x)).toBe(true)
```
hiroppy: nit: we can use `.toBeTruthy()`:)

capaj:
@mweststrate what do you mean by `pick mobx 4 or 15`? You're considering releasing next version as 15.0.0?

benjamingr:
@Bnaya ? Is this something you and michel discussed offline?

Bnaya:
@benjamingr you mean the `enforceActions: "strict"`? I just got these errors, pin-point them to the transpiled code,
```js
class Foo {
    @observable a = 1;
}
```
is
```js
class Foo {
    constructor() {
        this.a = 1;
    }
}
```
Which is considered as writing to an observable prop not in action context.
So i've added this comment on a specific commit.
Not sure why it ended up in this PR

package.json
mweststrate: Actually might be 15 :-) .Completely in the spirit of React.

But more importantly, I want to keep supporting MobX 4 for those we have to support browsers without proxies. And calling this one 5 would hamper the ability to make breaking changes on the "4-branch"

Bnaya: hmhm. Major is the new minor

Bnaya: I mean i have no objection, it's just a bit mind bending ^^

mweststrate:
@Bnaya not sure how it related to this PR :) But that behavior is correct; `enforceActions: true` is the 'classic' (and recommended) behavior, 'enforceActions: "strict"`, doesn't allow modifying any observables at all without actions, (including constructors, which are just normal functions after all). (So make sure your classes are created inside actions in this mode).

For giving those a clearer name there is a separate issue already. For follow up questions, please open a fresh issue, doesn't really relate to this PR (I hope :-P)

mweststrate:
Haha yeah not entirely sure about it yet either

Op ma 14 mei 2018 om 15:32 schreef Bnaya Peretz <notifications@github.com>:

Bnaya:
My original comment was on the commit that introduced this flag change, and because the commit is  for some reason in this PR, even that is already in master(?) the comment appear also as part of the PR

mweststrate:
Argh yeah, github diff is with base, not master. Always confuses me

Op ma 14 mei 2018 15:39 schreef Bnaya Peretz <notifications@github.com>:

mweststrate:
MobX 5 can now be tried! 
`mobx@5.0.0-beta.6`

It is a release candidate and I don't expect much changes anymore (we are using it already internally), changelog needs to be polished, and docs need to be updated to reflect the changes (volunteer welcome!)

benjamingr:
I'll take a look when I'm back from Berlin after next week - it's a lot of code and I hope I can help :) 

I think there is a lot of value in doing a blog post about a beta release.

CHANGELOG.md
```diff
@@ -1,16 +1,88 @@
+# 5.0.0
+
+* The minimum runtime target is now ES2015, not ES5
```
Bnaya: Idea: what about using preset-env with well defined and published browserlist,
And advise users that need to support older browser to pass it thru babel again with different browserlist?

jeffberry:
@mweststrate in your todos it appears that you stated all that would need to be done is to upgrade the peer-deps in the sister projects, but otherwise it should be backwards-compatible. I gave that a quick shot, but realized that mobx-state-tree will need a few changes, specifically in regards to the `$mobx` Symbol change; I checked for branches/PRs over there, but didn't find any. Were you already working on getting that project updated? If not, would you mind some help? I'd be happy to pitch in over there.

jeffberry:
Thank @mweststrate, I've started the updates and there doesn't appear to be many.

Another one I've ran into, and I want to get your input on before I go one way or another: in mobx-state-tree, `addHiddenFinalProp` is used quite often to assign `$treenode`, `toString`, etc to ObservableArray; now that those are proxy objects, those `defineProperty` calls are throwing a `Defining properties on observable arrays is not supported, directly assign them instead` exception now.

I think we can solve this by either A) using the proxy trap to define the property directly on the real array `$mobx.values` instead of throwing, something like: 
```
defineProperty(target, name, value) {
  Object.defineProperty(_getAdministration(target).values, name, value)
}
```

Or, B) change all of the `addHiddenFinalProp` calls in MST to something like this: `addHiddenFinalProp(_getAdministration(array).values, "toString", arrayToString)`

mweststrate:
@jeffberry good catch, thanks! I think that trap is just overly protective and can be removed

Other things left:
* [ ] remove defineProperty trap for arrays
* [ ] investige jest test difference for symbols (#1567 )
* [ ] upgrade to TS 2.9

CHANGELOG.md
mweststrate: Not familiar with preset-env. How would that work / what does it do?

tkrotoff: @mweststrate check https://github.com/tkrotoff/babel-preset-env-example and https://github.com/babel/website/issues/1660

To sum up, babel-preset-env automagically only includes the polyfills and babel plugins needed for a given list of targeted browsers (`browsers: ['> 1% in FR']`)

Bnaya: you also have granularity of last 2 major versions of edge, chrome, safari, and so

mweststrate:
Not sure if it is relevant, I really don't care or know exactly which
versions of browsers are supported, browser doesn't really matter, only
thing that is interesting is that the engine supports Proxy. From that
point on the browser will most probably be supported for a very long time

Op ma 4 jun. 2018 om 16:46 schreef Bnaya Peretz <notifications@github.com>:

CHANGELOG.md
```diff
@@ -1,16 +1,88 @@
...
+* Undeprecated `ObservableMap`, fixes [#1496](https://github.com/mobxjs/mobx/issues/1496)
+* Observable arrays now support `Symbol.toStringTag` (if available / polyfilled). This allows libraries like Ramda to detect automatically that observable arrays are arrays. Fixes [#1490](https://github.com/mobxjs/mobx/issues/1490). Note that `Array.isArray` will keep returning false for the entire MobX 4 range.
+* Actions are now always `configurable` and `writable`, like in MobX 3. Fixes [#1477](https://github.com/mobxjs/mobx/issues/1477)
+* Merged several improvements to the flow typings. [#1501](https://github.com/mobxjs/mobx/pull/1501) by @quanganhtran
+* Fixed several accidental usages of the global `fail`, by @mtaran-google through [#1483](https://github.com/mobxjs/mobx/pull/1483) and [#1482](https://github.com/mobxjs/mobx/pull/1482)
+
+# 4.1.1
+
+* Import `default` from MobX will no longer throw, but only warn instead. This fixes some issues with tools that reflect on the `default` export of a module.
```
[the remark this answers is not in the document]
mweststrate: Fixed

.travis.yml
```diff
@@ -5,7 +5,7 @@ script: CI=true yarn test:travis
 after_success:
   - cat ./coverage/lcov.info|./node_modules/coveralls/bin/coveralls.js
 node_js:
-  - 6
+  - 9.3
```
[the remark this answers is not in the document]
mweststrate: Updated

package.json
```diff
@@ -71,10 +71,9 @@
     "ts-jest": "^22.0.0",
     "tslib": "^1.7.1",
     "typescript": "^2.1.4",
-    "uglify-js": "^2.6.1"
-  },
-  "dependencies": {
+    "uglify-es": "^3.3.9"
```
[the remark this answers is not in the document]
mweststrate: Fixed, should have been dev indeed

scripts/build.js
```diff
@@ -77,11 +77,11 @@ function generateMinified() {
     console.log("Generating mobx.min.js and mobx.umd.min.js")
     exec(`NODE_ENV=production ${binFolder}/envify lib/mobx.js > lib/mobx.prod.js`)
     exec(
-        `${binFolder}/uglifyjs -m sort,toplevel -c warnings=false --screw-ie8 --preamble "/** MobX - (c) Michel Weststrate 2015, 2016 - MIT Licensed */" --source-map lib/mobx.min.js.map -o lib/mobx.min.js lib/mobx.prod.js`
+        `${binFolder}/uglifyjs --toplevel -m -c warnings=false --preamble "/** MobX - (c) Michel Weststrate 2015, 2016 - MIT Licensed */" --source-map -o lib/mobx.min.js lib/mobx.prod.js`
```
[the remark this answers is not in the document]
mweststrate: Updated

mweststrate:
* [ ] Fix binding of computed properties

CHANGELOG.md
```diff
@@ -1,16 +1,118 @@
...
+* `.$mobx` property has been dropped from all observables and replaced by a Symbol. Instead of using `x.$mobx.name`, use `import { $mobx } from "mobx"; x[$mobx].name` etc.
+* In some cases, the order in which autoruns are fired could have changed due to some internal optimizations (note that MobX never had a guarantee about the order in which autoruns fired!)
+
+### New features
+
+* It is possible to pass the `proxy: false` argument to `observable.object` to disable proxying (theoretically slightly faster, but removes no dynamic key addition)
+
+### Known Issues
+
+* Jest `toEqual` might throw an error `allKeys[x].match is not a function` when trying to equal observable arrays. This is a bug in Jest [report](https://github.com/facebook/jest/issues/6391). The simple work around for now is to slice (or `toJS` if the problem is recursive) the array first.
```
benjamingr: This was merged into jest - so update to "older versions of jest"?

CHANGELOG.md
```diff
@@ -1,16 +1,118 @@
...
+* Jest `toEqual` matcher might no longer corretly equal your class instances, complaining about differences in the MobX adminstration. This is due to a bug with the processing of symbols: [report](https://github.com/facebook/jest/issues/6392). For now you might want to use a custom matcher if you are directly equalling observable objects. As a work around `toJS(object)` could be used before diffing.
```
benjamingr: Also fixed I think?

mweststrate: Same

mweststrate:
Released!

PR #1383, opened 2018-03-09
introduced canceellable flows

mweststrate:
```javascript
const asyncFlow = flow(function * () { 
    // do stuff
   try {
      yield Promise.resolve(3)
   } catch (e) {
       // cancel end's up here, e is "FLOW_CANCELLED"
      throw e // now 'promise' will reject with "FLOW_CANCELLED"
   }
})
const promise = asyncFlow()

promise.cancel() // rejects the next yield / return
```

Would you mind reviewing @benjamingr ? Does this approach make sense?

### DIFF
--- a/src/api/flow.ts
+++ b/src/api/flow.ts
@@ -117,53 +117,58 @@ export function createFlowGenerator(name: string, generator: Function) {
         const args = arguments
         const runId = ++generatorId
         const gen = action(`${name} - runid: ${runId} - init`, generator).apply(ctx, args)
-        let stepId = 0
-        let resolver: (value: any) => void
         let rejector: (error: any) => void
 
-        function onFulfilled(res: any) {
-            let ret
-            try {
-                ret = action(`${name} - runid: ${runId} - yield ${stepId++}`, gen.next).call(
-                    gen,
-                    res
-                )
-            } catch (e) {
-                return rejector(e)
-            }
-            next(ret)
-            return null
-        }
-
-        function onRejected(err: any) {
-            let ret
-            try {
-                ret = action(`${name} - runid: ${runId} - yield ${stepId++}`, gen.throw).call(
-                    gen,
-                    err
-                )
-            } catch (e) {
-                return rejector(e)
-            }
-            next(ret)
-        }
-
-        function next(ret: any) {
-            if (ret.done) return resolver(ret.value)
-            // TODO: support more type of values? See https://github.com/tj/co/blob/249bbdc72da24ae44076afd716349d2089b31c4c/index.js#L100
-            if (!ret.value || typeof ret.value.then !== "function")
-                return fail("Only promises can be yielded to asyncAction, got: " + ret)
-            return ret.value.then(onFulfilled, onRejected)
-        }
-
         const res = new Promise(function(resolve, reject) {
-            resolver = resolve
+            let stepId = 0
             rejector = reject
+
+            function onFulfilled(res: any) {
+                let ret
+                try {
+                    ret = action(`${name} - runid: ${runId} - yield ${stepId++}`, gen.next).call(
+                        gen,
+                        res
+                    )
+                } catch (e) {
+                    return reject(e)
+                }
+                next(ret)
+                return null
+            }
+
+            function onRejected(err: any) {
+                let ret
+                try {
+                    ret = action(`${name} - runid: ${runId} - yield ${stepId++}`, gen.throw).call(
+                        gen,
+                        err
+                    )
+                } catch (e) {
+                    return reject(e)
+                }
+                next(ret)
+            }
+
+            function next(ret: any) {
+                if (ret.done) return resolve(ret.value)
+                // TODO: support more type of values? See https://github.com/tj/co/blob/249bbdc72da24ae44076afd716349d2089b31c4c/index.js#L100
+                if (!ret.value || typeof ret.value.then !== "function")
+                    return fail("Only promises can be yielded to asyncAction, got: " + ret)
+                return ret.value.then(onFulfilled, onRejected)
+            }
+
             onFulfilled(undefined) // kick off the process
         }) as any
-        res.cancel = function() {
-            onRejected(new Error("FLOW_CANCELLED"))
-        }
+
+        res.cancel = action(`${name} - runid: ${runId} - cancel`, function() {
+            try {
+                gen.return()
+                rejector(new Error("FLOW_CANCELLED"))
+            } catch (e) {
+                rejector(e) // there could be a throwing finally block
+            }
+        })
         return res
     }
 }
--- a/src/api/flow.ts
+++ b/src/api/flow.ts
@@ -118,12 +118,14 @@ export function createFlowGenerator(name: string, generator: Function) {
         const runId = ++generatorId
         const gen = action(`${name} - runid: ${runId} - init`, generator).apply(ctx, args)
         let rejector: (error: any) => void
+        let pendingPromise: CancellablePromise<any> | undefined = undefined
 
         const res = new Promise(function(resolve, reject) {
             let stepId = 0
             rejector = reject
 
             function onFulfilled(res: any) {
+                pendingPromise = undefined
                 let ret
                 try {
                     ret = action(`${name} - runid: ${runId} - yield ${stepId++}`, gen.next).call(
@@ -138,6 +140,7 @@ export function createFlowGenerator(name: string, generator: Function) {
             }
 
             function onRejected(err: any) {
+                pendingPromise = undefined
                 let ret
                 try {
                     ret = action(`${name} - runid: ${runId} - yield ${stepId++}`, gen.throw).call(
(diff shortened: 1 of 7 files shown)


pavanpodila:
One issue I am seeing here is with the use of `@flow`, where the type signature is not recognizable by TypeScript, resulting in a type check failure:

```ts
class AuthStore {
@flow
    *login() {
        this.isLoggingIn = true;
        try {
            yield service.googleLogin();
            this.loginUser = new User('Pavan Podila', 'pavan@pixelingene.com');
        } finally {
            this.isLoggingIn = false;
        }
    }
}

// Usage

authStore.login().then(()=>{ console.log('logged in'); }) // FAILS type check
```

**Exception**
```
error TS2339: Property 'then' does not exist on type 'IterableIterator<Promise<{}>>'.
```
Is it a type definition issue ?

benjamingr:
I think this should probably be an async iterator so we can await inside it (rather than an iterator), also note that it should probably be a function (rather than a decorator) so it can change the types for TS

benjamingr:
Also, I would considering "disinterest semantics" which would mean calling `return` on the flow on cancellation (rather than `throw`), this lets cleanup run (`.finally` blocks) but it does _not_ call `catch` blocks. 

It's the "correct" approach to cancellation and it's what most modern'ish systems do (bluebird cancellation being one).

mweststrate:
Double check: that is a return with a Promise.reject("FLOW_CANCELLED") ?

Op zo 11 mrt. 2018 om 13:10 schreef Benjamin Gruenbaum <
notifications@github.com>:

benjamingr:
I don't think anyone is sure what `return`'s argument really does, I recall @jhusain regretting adding it. I would call `.return` without any arguments - `finally` is still run (so cleanup) but no further code runs.

This was the ES promise cancellation proposal before it was withdrawn and is what bluebird does.

Bnaya:
curious:
Why not similar to `AbortController`? or advocate the user to implement his own aborting using it?

benjamingr:
@Bnaya 
> Why not similar to AbortController? or advocate the user to implement his own aborting using it?

To be fair, `AbortController` was decided at a whatwg `fetch` meeting not due to merit of its design or architecture but because everyone involved wanted to make progress with it and cancellation was 2 years overdue when we waited for promise cancellation that didn't come natively.

Here it's not a promise, I don't see why we'd want to use an `AbortController` with a `signal` for cancellation, and semantics would be flow like (rather than a promise).

I do think it would make a lot more sense if this was an async iterator though :D

mweststrate:
@benjamingr updated. Care for re-review? `.cancel` now returns the generator (running finally blocks if needed), and then reject original promise with cancellation. if the finally block throws, the promise will be rejected with that error

src/api/flow.ts
```diff
@@ -117,53 +117,58 @@ export function createFlowGenerator(name: string, generator: Function) {
...
-            if (ret.done) return resolver(ret.value)
-            // TODO: support more type of values? See https://github.com/tj/co/blob/249bbdc72da24ae44076afd716349d2089b31c4c/index.js#L100
-            if (!ret.value || typeof ret.value.then !== "function")
-                return fail("Only promises can be yielded to asyncAction, got: " + ret)
-            return ret.value.then(onFulfilled, onRejected)
-        }
+            function next(ret: any) {
+                if (ret.done) return resolve(ret.value)
+                // TODO: support more type of values? See https://github.com/tj/co/blob/249bbdc72da24ae44076afd716349d2089b31c4c/index.js#L100
+                if (!ret.value || typeof ret.value.then !== "function")
```
benjamingr: Native async functions allow `await`ing arbitrary things, although I still think we shouldn't roll our own coroutines if we don't also get async generators just for cancellation.

src/api/flow.ts
```diff
@@ -117,53 +117,58 @@ export function createFlowGenerator(name: string, generator: Function) {
...
             onFulfilled(undefined) // kick off the process
         }) as any
-        res.cancel = function() {
-            onRejected(new Error("FLOW_CANCELLED"))
-        }
+
+        res.cancel = action(`${name} - runid: ${runId} - cancel`, function() {
+            try {
+                gen.return()
+                rejector(new Error("FLOW_CANCELLED"))
```
benjamingr: So externally we're rejecting but internally we're doing cancellation.

What happens if `.cancel` is called while a flow is being `yield`ed in a flow? Preferably, the internal flow should be cancelled as well (so that, if the current promise I'm waiting for has a `.cancel` and `.cancel` was called on me - call it on that promise).

test/base/flow.js
```diff
@@ -170,62 +170,72 @@ test("flows can be cancelled - 1 - uncatched cancellation", done => {
...
             expect("" + err).toBe("Error: FLOW_CANCELLED")
             done()
         }
     )
+    debugger
     promise.cancel()
 })
 
-test("flows can be cancelled - 2 - catch cancellation in generator", done => {
+test("flows can be cancelled - 2 - finally clauses are run", done => {
```
benjamingr: This behavior looks correct.

test/base/flow.js
```diff
@@ -236,14 +246,8 @@ test("flows can be cancelled - 4 - pending Promise will be ignored", done => {
...
         steps = 1
-        try {
-            yield Promise.reject("This won't be catched anywhere!") // cancel will resolve this flow before this one is throw, so this promise goes uncatched
-            steps = 2
-        } catch (e) {
-            expect(steps).toBe(1)
-            expect(e.toString()).toBe("Error: FLOW_CANCELLED")
-            throw e
-        }
+        yield Promise.reject("This won't be catched anywhere!") // cancel will resolve this flow before this one is throw, so this promise goes uncatched
```
benjamingr: So, this is a little risky (what happens if a cancelled function throws) but I think this is the more reasonable compromise.

src/api/flow.ts
mweststrate: good point, will call `cancel` on any pending promise if possible

src/api/flow.ts
```diff
@@ -210,10 +154,22 @@ export function createFlowGenerator(name: string, generator: Function) {
...
+                    return fail("Only promises can be yielded to asyncAction, got: " + ret)
                 return ret.value.then(onFulfilled, onRejected)
             }
+
+            onFulfilled(undefined) // kick off the process
+        }) as any
+
+        res.cancel = action(`${name} - runid: ${runId} - cancel`, function() {
+            try {
+                gen.return()
```
benjamingr: Check if `return` exists on the generator maybe? If a non-generator iterator is available, it's probably best to skip the call. This is a really edgy edge case though.

src/api/flow.ts
```diff
@@ -210,10 +154,22 @@ export function createFlowGenerator(name: string, generator: Function) {
                 if (ret.done) return resolve(ret.value)
                 // TODO: support more type of values? See https://github.com/tj/co/blob/249bbdc72da24ae44076afd716349d2089b31c4c/index.js#L100
                 if (!ret.value || typeof ret.value.then !== "function")
-                    fail("Only promises can be yielded to asyncAction, got: " + ret)
+                    return fail("Only promises can be yielded to asyncAction, got: " + ret)
```
benjamingr: This is different from the default async/await behavior

src/api/flow.ts
```diff
@@ -173,11 +115,13 @@ export function createFlowGenerator(name: string, generator: Function) {
     return function() {
         const ctx = this
         const args = arguments
-        return new Promise(function(resolve, reject) {
-            const runId = ++generatorId
+        const runId = ++generatorId
+        const gen = action(`${name} - runid: ${runId} - init`, generator).apply(ctx, args)
+        let rejector: (error: any) => void
+
+        const res = new Promise(function(resolve, reject) {
```
benjamingr: I think we should check if `gen` returned a promise (which is the case if it's an async generator) and then `then` it with `next` rather than passing `gen` to next.

That way, you can use async generators and yield for state changes and avoid accidental changes.

I still think it would be really awesome if you could `await` for async stuff but `yield` for state updates of an observable.

src/api/flow.ts
mweststrate: Fixed, using `Promise.resolve(ret.value)` now

src/api/flow.ts
mweststrate: > I still think it would be really awesome if you could await for async stuff but yield for state updates of an observable. 

Me2, but different story I think :)

src/api/flow.ts
benjamingr: That sounds good, though it's a matter of choice mostly

src/api/flow.ts
benjamingr: That would just require checking if `gen` returns a promise when you call `.next` (which would make it an async iterator) and then wait for it - that's all it takes given the implementation since `await`s would work anyway and `yield` would run in an action because of the code above. Cancellation would also work for sub actions just the same.

mweststrate: Added support for async iterators

benjamingr:
Awesome, although I think it's important to be able to yield into an observable - so that yielded values update a check too.

benjamingr:
So basically, `flow` as we have it now, but whenever a value is `yielded` we also update a backing observable. It should be a _very_ small change at this point and it'll let me write a ton of cool demos with it :D

mweststrate:
@benjamingr interested in creating a sandbox or PR showing that of? I think it is a super interesting direction to further pursue, but most probably not breaking so we could add it in 4.1 for example

benjamingr:
I was afraid you were going to say that :D I guess I'll have to dig into the source and add it.

PR #1384, opened 2018-03-09
add removeAll support to ObservableArray

### LINKED ISSUE #1248: array remove by predicate should be implemented

benbraou:
Fixes #1248

PR checklist:

* [x] Added unit tests
* [x] Updated changelog
* [x] Updated docs => https://github.com/mobxjs/mobx/pull/1385

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 
+# 3.6.2
+
+* Fixed [#1248](https://github.com/mobxjs/mobx/issues/1248): removeAll support for observable arrays. By [benbraou](https://github.com/benbraou) through [#1384](https://github.com/mobxjs/mobx/pull/1384)
+
 # 3.6.1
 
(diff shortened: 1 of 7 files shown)


benbraou:
@mweststrate The size limit 20KB is exceeded (the size is now 20.02 KB). Is this acceptable? Should I just update the `size` script in `package.json` ? In that case, what do you recommend as a new limit?

mweststrate:
Yeah feel free to update the size thing a little bit. Thanks for the neat pr, like it should be done!

I'm thinking though on putting this PR on hold, as it kinda bites the direction where mobx is going; making arrays as native as possible. So for example `find` / `findIndex` and `move` will also be dropped in MobX 5 (relying on the browser versions if necessary).

As soon as arrays are proxied, these utilties are less needed, as one could just use the lodash counterpart, and array convenience utilities aren't really the scope of this package

mweststrate:
@benbraou thanks for the effort, and apologies of bringing that not up earlier and wasting your time! I'll close it for now, we can always reconsider if there is lot's of interest.

PR #1386, opened 2018-03-10
WIP: MobX 4 doc updates

mweststrate:


docs/refguide/autorun.md
```diff
@@ -29,33 +29,40 @@ numbers.push(5);
...
-## Error handling
+## The `delay` option
+
+```javascript
+autorun(() => {
+	// Assuming that profile.asJson returns an observable Json representation of profile,
+	// send it to the server each time it is changed, but await at least 300 milliseconds before sending it.
+	// When sent, the latest value of profile.asJson will be used.
+	sendProfileToServer(profile.asJson);
+}, { delay: 300 });
```
pavanpodila: I think it's worth mentioning that this behavior is like **`debounce()`** from libraries like lodash or underscore where updates during the `delay` period are ignored.

PR #1394, opened 2018-03-12
Added something missing from the CHANGELOG

jjinux:
While upgrading from 2.6.0 to 2.7.0, I figured out that if people used `observable({ @computed get someProp()` in their code, reactions wouldn't happen leading to silent test failures and incredibly difficult to debug code. I added some notes to the CHANGELOG about this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -613,6 +613,18 @@ observable({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks!

jjinux:
Yay!

PR #1399, opened 2018-03-13
Minor fix for flow typing

ismailhabib:
The flow typing which accept generator with one and two parameters are returning `Promise` instead of `CancellablePromise`.


mweststrate:
Yay! Thanks

mweststrate:
...but but, are you using flow :-P?

PR #1451, opened 2018-03-23
Flow-typed: add missing runInAction declaration

AMilassin:
Updated the flow typings with the missing runInAction declaration.


mweststrate:
Thanks!

PR #1452, opened 2018-03-23
Expanded and updated docs on configure

### LINKED ISSUE #1420: Docs: add a page on configuring?
I have an:
3. [x] Idea:
  * [x] What problem would it solve for you?
  * [x] Do you think others will benefit from this change as well and it should in core package (see also mobx-utils)?
  * [x] Are you willing to (attempt) a PR yourself?

I can't find any documentation on the `configure` call. I only know of it because `useStrict` is no longer available and was mentioned in the changelog.

```typescript
import { configure } from "mobx";

configure({
	enforceActions: true,
});
```

I can submit a doc page for this. It should probably be under section 6. Utility Functions?

JoshuaKGoldberg:
Fixes #1420

### DIFF
--- a/docs/refguide/api.md
+++ b/docs/refguide/api.md
@@ -460,13 +460,75 @@ export class City {
 
 ### `configure`
 Usage: `configure(options)`.
+Sets global behavior settings on the active MobX instance.
+Use this to change how MobX behaves as a whole.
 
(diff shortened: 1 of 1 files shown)


mweststrate:
Great, thanks!

Issue #1459, opened 2018-03-27
The problem removing spy listener from spy handler


farwayer:
2. [x] Issue:
  * [x] **Provide error messages including stacktrace**

```js
node_modules/mobx/lib/mobx.js:160
        listeners[i](event);
                    ^

TypeError: listeners[i] is not a function
    at spyReport (node_modules/mobx/lib/mobx.js:160:21)
    at spyReportStart (node_modules/mobx/lib/mobx.js:164:5)
    at notifyPropertyAddition (node_modules/mobx/lib/mobx.js:1243:9)
    at defineObservableProperty (node_modules/mobx/lib/mobx.js:1183:5)
    at Object.propertyCreator (node_modules/mobx/lib/mobx.js:1264:9)
    at initializeInstance (node_modules/mobx/lib/mobx.js:105:15)
    at Todo.set [as title] (node_modules/mobx/lib/mobx.js:92:17)
    at Object.<anonymous> (test.js:12:12)
    at Module._compile (module.js:649:30)
    at Object.Module._extensions..js (module.js:660:10)
```

  * [x] Provide a **minimal** sample reproduction. **Create a reproduction based on this [sandbox](https://codesandbox.io/s/v3v0my2370)**

https://codesandbox.io/s/qqz1nz9144

```js
import { spy, decorate, observable } from 'mobx';

const stopSpy = spy(() => stopSpy());
spy(() => { });

class Todo { }
decorate(Todo, {
  title: observable,
});

const todo = new Todo();
todo.title = "";
```

  * [x] Did you check this issue wasn't filed before?
  * [x] Elaborate on your issue. What behavior did you expect?

The problem is mobx do not copy listeners before enumerate it.

https://github.com/mobxjs/mobx/blob/4.1.0/src/core/spy.ts#L11
https://github.com/mobxjs/mobx/blob/4.1.0/src/core/spy.ts#L30

  * [x] State the versions of MobX and relevant libraries. Which browser / node / ... version?

**mobx 4.1.0**

PR #1460, opened 2018-03-27
Spy: fix stop listen from handler

farwayer:
Fix https://github.com/mobxjs/mobx/issues/1459

Making copy of `listeners` in `spyReport` can have performance impact so I changed unsubscribe action because it rarer event. Also all same listeners is removed with this change (not only first).


mweststrate:
Thanks!

PR #1483, opened 2018-04-07
add another missing import of `fail`

mtaran-google:
this makes strict compilation error out


mtaran-google:
maybe I spoke too soon. I'll just close my big mouth and investigate more before commenting :)

mweststrate:
Both PRs are released as part of 4.2.0. Thanks for the PR!

PR #1499, opened 2018-04-13
avoid the unnecessary reassign with the equivalent comparison result

kuitos:
Now the computed value will be reassign although the assignment was equivalent by the comparer.

@mweststrate 
Here is the [jsfiddle demo](https://jsfiddle.net/Kuitos/ycv4bb2a/15/) which describes the unexpected behavior, or u can check the added unit test for details🙂

* [x] Added unit tests
* [ ] Updated changelog


mweststrate:
Thanks! This makes a lot of sense :-). Thanks for catching and fixing

mweststrate:
Released as 4.2.0

Issue #1500, opened 2018-04-14
Flow typing of observable.map is not updated for MobX 4


quanganhtran:
I have a:

2. [x] Issue:
  * [x] Did you check this issue wasn't filed before?
  * [x] State the versions of MobX and relevant libraries: MobX 4

Since MobX 4 uses ES6 Map, all typings related to `Map<T>` has been updated to `Map<K, V>` except `observable.map`. This introduces Flow error with Flow 0.69

I have prepared a PR to update the definition.

mweststrate:
Merged! Will release soon

PR #1501, opened 2018-04-14
Fix Flow typing for observable.map

quanganhtran:
It is a PR to address #1500 

Flow typing of `observable.map` is changed from generic of `<T>` to `<K, V>`


mweststrate:
Looking good, thanks!

mweststrate:
Released as part of 4.2.0. Thanks!

Issue #1454, opened 2018-03-23
Should putting `@observable` on a getter produce an error?


JoshuaKGoldberg:
I have a:

1. [x] Question: Feel free to just state your question. For a quick answer, there are usually people online at our [Gitter](https://gitter.im/mobxjs/mobx) channel.

I did a silly:

```typescript
class MyStore {
    @observable
    public value: number;

    @observable // should have been @computed
    public get double() { return this.value * 2; }
}
```

Should this produce an error? In practice, `MyStore.prototype.timeSinceLastUpdated` is `undefined`, as is all its instances.

mweststrate:
It is not a supported case indeed, so a PR for runtime checks is welcome

quanganhtran:
Released in 4.2.1

PR #1511, opened 2018-04-21
Throw error in development when @observable is used on getters

quanganhtran:
This PR is to address #1454

The variant check is added to the property creator function. The caveat is that it throws when the target is initialized (even if the property being accessed is not the erroneous one).

Fixed random typos in comments that I came across.

* [x] Added unit tests
* [ ] Updated changelog


mweststrate:
Thanks!

PR #1513, opened 2018-04-22
Docs: Remove nonexistent options from autorun docs

Retsam:
It looks like the options section of the `autorun` docs were copied from the docs for `reaction`, and so include options that don't apply to `autorun`.


mweststrate:
Dôh, good catch!

PR #1515, opened 2018-04-22
Fix invariant check for ComputedValue constructor

quanganhtran:
The invariant check currently is checking *only* in production. That was not intended, so we change it to check when *not* in production.

* [ ] Added unit tests
* [ ] Updated changelog


mweststrate:
Thanks!

wbercx:
@mweststrate Would it be possible to release a version that includes this fix?

PR #1518, opened 2018-04-23
toJS: support objectKeysFn options

kindy:

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/src/api/tojs.ts
+++ b/src/api/tojs.ts
@@ -8,6 +8,7 @@ import { keys } from "./object-api"
 export type ToJSOptions = {
     detectCycles?: boolean
     exportMapsAsObjects?: boolean
+    objectKeysFn?: (object) => string[]
 }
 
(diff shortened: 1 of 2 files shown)


mweststrate:
I rather like to prevent customizing behavior of `toJS`, as otherwise the number of customizations get pretty  endless; you now allow to customize the set of properties to serialize, but similarly one might want to customize the serialization for certain fields completely. `toJS` is just an default implementation to un-observable observable objects, but for custom behavior on how stuff should be serialized one should better take a look at utilties that are designed for this. For example lodash.deepClone or the serializr package. Otherwise the amount of scenarios that `toJS` would have to support are pretty endless, and in general writing your own recursive serialization function, optimized for problem domain is not that hard.

kindy:
@mweststrate well, got it.

Issue #1496, opened 2018-04-13
Consider exporting ObservableMap once again


nathanstitt:
It was exported in 3.x and I don't see any discussion on why it's no longer exported. I do appreciate the deprecation message though, since that at least told me it was intentional 😁 

I'm guessing there were good reasons for it's removal, but I thought I'd open this issue in case it was just "we thought no-one used it"

There's at least a few people who found it useful https://github.com/mobxjs/mobx/issues/863

My use-case for it is that I have custom store-like objects that add filtering methods to maps.  

There are multiple alternative ways these could be built, such as a factory that builds custom maps with the methods added, or possibly interceptors in some cases.  I think my method is the clearest though, and I suspect that a they're more easily optimized by JS engines.

Of course we could still get the class by `observable.map().constructor` 😉 .  Therefore I don't think you're gaining anything by not exporting it.

I have a:

1. [x] Question: Feel free to just state your question. For a quick answer, there are usually people online at our [Gitter](https://gitter.im/mobxjs/mobx) channel
2. [ ] Issue:
  * [ ] **Provide error messages including stacktrace**
  * [ ] Provide a **minimal** sample reproduction. **Create a reproduction based on this [sandbox](https://codesandbox.io/s/v3v0my2370)**
  * [ ] Did you check this issue wasn't filed before?
  * [ ] Elaborate on your issue. What behavior did you expect?
  * [ ] State the versions of MobX and relevant libraries. Which browser / node / ... version?
3. [x] Idea:
  * [ ] What problem would it solve for you?
  * [ ] Do you think others will benefit from this change as well and it should in core package (see also mobx-utils)?
  * [x] Are you willing to (attempt) a PR yourself?

mweststrate:
The reason it was original deprecated was that it might disappear in MobX 5. But it will most probably stay, so will undeprecated it :)

nathanstitt:
ok, I'll make a PR to do so then, it looks fairly straightforward to do so. Thanks @mweststrate !

mweststrate:
Released as 4.2.0

mweststrate:
Sorry, just beat you to it!

Op ma 16 apr. 2018 om 20:06 schreef Nathan Stitt <notifications@github.com>:

mweststrate:
Thanks for the effort though :)

Op ma 16 apr. 2018 om 21:13 schreef Michel Weststrate <mweststrate@gmail.com

PR #1525, opened 2018-04-30
Extending ObservableMap with computed properties fails

nathanstitt:
I've been updating to mobx4 and discovered an interesting bug

As we discussed in https://github.com/mobxjs/mobx/issues/1496 maps weren't exported at first, but you kindly re-exported them.  However I've now discovered that extending a map works fine except if you add a @computed decorator to a getter.

That causes a `Cannot set property '<property>' of undefined`, because the $mobx property on maps isn't a "real" instance of ObservableObjectAdministration, and lacks the `values` property (stacktrace below).  I attempted various things to fix that (commented in this PR), but that broke several specs.   

If you can provide some guidance on what you think the best approach would be I'll attempt to do that, but I'm out of ideas for now.

The backtrace of the error:
```
    TypeError: Cannot set property 'mult' of undefined

      286 |     options.name = `${adm.name}.${propName}`
      287 |     options.context = target
    > 288 |     adm.values[propName] = new ComputedValue(options)
      289 |     Object.defineProperty(target, propName, generateComputedPropConfig(propName))
      290 | }
      291 |

      at Object.defineComputedProperty (src/types/observableobject.ts:288:25)
      at Object.propertyCreator (src/api/computed.ts:29:9)
      at initializeInstance (src/utils/decorators2.ts:59:15)
      at MyMap.get (src/utils/decorators2.ts:20:13)
      at Object.<anonymous> (test/base/map.js:717:18)
```

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -1,5 +1,6 @@
 import { IEnhancer, deepEnhancer } from "./modifiers"
 import { untracked } from "../core/derivation"
+import { asObservableObject } from "./observableobject"
 import { IObservableArray, ObservableArray } from "./observablearray"
 import { ObservableValue, UNCHANGED } from "./observablevalue"
 import {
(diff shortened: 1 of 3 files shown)


mweststrate:
extendObservable is only applicable to observable objects, not maps or arrays (as you discovered yourself your patch isn't sufficient and the impact would be actually much bigger). 

I updated the docs to make that more clear that this pattern is not supported

PR #1536, opened 2018-05-08
add entries method to observable object api [WIP]

samjacobclift:

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

Still a WIP, looking at the best place to add the unit tests

### DIFF
--- a/src/api/object-api.ts
+++ b/src/api/object-api.ts
@@ -45,6 +45,25 @@ export function values(obj: any): string[] {
     )
 }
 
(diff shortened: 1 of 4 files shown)


mweststrate:
The typings are not correct (entries returns `[[K, V]]`) but I'll update them. Thanks for the PR!

samjacobclift:
Thanks for the merge, I've just recently started using mobx its a fantastic tool 👍

Issue #1534, opened 2018-05-08
Use @computed({keepAlive: true}) without autorun


hector:
I would like this to work:

```js
class Foo {
  @observable stuff = 'acme';

  @computed get coolStuff() {
    return expensiveCalculationWithStuff();
  }

}

const foo = Foo();
console.log(coolStuff); // expensiveCalculationWithStuff() is called
console.log(coolStuff); // expensiveCalculationWithStuff() is NOT called: use cached value
```

I have learned from other issues (https://github.com/mobxjs/mobx/issues/1093) that the cached value of a computed is only kept while something is observing it.

The suggested solution is to use `@computed({keepAlive: true})`. This works but it calculates the computed value on object creation, not when trying to read the computed value for the first time. This is because `keepAlive: true` creates an `autorun`:
https://github.com/mobxjs/mobx/blob/8fb2861d551e254ebb3c6f167f3327dfdcbfa7bb/src/core/computedvalue.ts#L126

This seems wrong since it breaks the description of a computed:

> 1. First time it's being accessed it will compute and remember result

Trying solutions I have found that using `observe` instead of `autorun` works fine: it caches the value but it does not calculate it before it is needed. Should this change be done?

The only concern I have is if `observe` will get disposed when `foo` is garbage collected.

urugator:
Eager behavior is not very practical indeed. (btw it was already [mentioned here](https://github.com/mobxjs/mobx-utils/issues/51#issuecomment-299345024))
I quess it wasn't implemented, because it would require to change the implementation of `computed` itself and added complexity wouldn't worth the limited usage of keelAlive? dunno...    

> I have found that using observe instead of autorun works fine: it caches the value but it does not calculate it before it is needed.

You sure? Check the console https://jsfiddle.net/nLev2ozo/1/

As a workaround you can create autorun lazily by yourself

```javascript
class Foo {
  @observable stuff = 'acme';  
    
  @computed get _coolStuff() {
    /* expensive computation */
  }  
  
  get coolStuff() {
    if (!this._coolStuffDisposer) {
      this._coolStuffDisposer = autorun(() => this._coolStuff);      
    }
    return this._coolStuff;
  }
  
  // Optionally expose disposer
  disposeCoolStuff() {
    this._coolStuffDisposer();
    this._coolStuffDisposer = null; 
  }
}
```

hector:
> You sure? Check the console https://jsfiddle.net/nLev2ozo/1/

[...]

PR #1560, opened 2018-05-25
@computed({keepAlive: true}) no long calculates before being accessed

hector:
Fixes https://github.com/mobxjs/mobx/issues/1534

PR checklist:

* [x] Added unit tests
* [x] Updated changelog
* [x] Updated docs. [docs not changed, in my opinion, with this change we respect what the docs already say]
* [x] Added typescript typings. [not necessary]
* [x] Verified that there is no significant performance drop (`npm run perf`) 16.697 -> 17.049


mweststrate:
Thanks! Will ship with next minor as this is a subtle semantic change

hector:
Great :)

PR #1562, opened 2018-05-26
adds Browser support section to the main readme

capaj:
as this might be a common question for mobx 5 and up


mweststrate:
Thanks!

PR #1564, opened 2018-05-27
Remove redundant code

JSlife23:
I found that there is a tool function hasOwnProperty under utils/utils.ts that is not used to be redundant and that there is the same function in utils/eq.ts, has, should hasOwnProperty be deleted?

utils/utils.ts
const prototypeHasOwnProperty = Object.prototype.hasOwnProperty
export function hasOwnProperty(object: Object, propName: string) {
    return prototypeHasOwnProperty.call(object, propName)
}
utils/eq.ts
function has(a: any, key: string) {
    return Object.prototype.hasOwnProperty.call(a, key)
}

### DIFF
--- a/src/utils/utils.ts
+++ b/src/utils/utils.ts
@@ -58,14 +58,14 @@ export function deprecated(msg: string, thing?: string): boolean {
  */
 export function once(func: Lambda): Lambda {
     let invoked = false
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for spotting!

Issue #1568, opened 2018-05-31
Can't initialize properties when using `strict` mode


newraina:
I'm having an issue with MobX 4.3.0, properties initializing are causing errors if I use enforceActions: 'strict'：

```js
import { observable, configure } from 'mobx'

configure({
  enforceActions: 'strict'
})

const store = observable({
  names: [1, 2]
})

console.log(store.names)
```

Causing error:

```
Error: [mobx] Since strict-mode is enabled, changing observed observable values outside actions is not allowed. Please wrap the code in an `action` if this change is intended. Tried to modify: ObservableObject@18.names
```

Reproduction link:
https://codesandbox.io/s/l5l3l9v8ll

Does this is intended? `store.names` just has a initialized value, I never tried to modify.

PR #1573, opened 2018-06-05
bugfix: allow changing while observable array initialization

kuitos:
fixed: https://github.com/mobxjs/mobx/issues/1568

* [x] Added unit tests

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -23,6 +23,7 @@ import { IListenable, registerListener, hasListeners, notifyListeners } from "./
 import { isSpyEnabled, spyReportStart, spyReportEnd } from "../core/spy"
 import { declareIterator, makeIterable } from "../utils/iterable"
 import { IEnhancer } from "./modifiers"
+import { allowStateChangesEnd, allowStateChangesStart } from "../core/action"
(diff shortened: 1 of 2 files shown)


mweststrate:
Thanks!

jspopisno1:
@mweststrate 

Our project is now switching to 4.3.1 and we also found bug #1568 .
Was glad to see this bugfix, but wondering why it's not in 4.3.1 (as it's reported in 4.3.0) ?
Found code in 
 4.3.1 as https://github.com/mobxjs/mobx/blob/2c82242bb2ff6582c42e3963b5f83f98475541a9/src/types/observablearray.ts#L338

Are we using it incorrectly or is it really a bug in latest 4.3.x ?

Hope for further info. and heaps of thanks!

mweststrate:
Please open a new issues when you run into something. Commenting on a closed item gives only a 20÷ chance something sees your comment. 

Please open a new issue,  refer to this one and we'll look into it

jspopisno1:
@mweststrate Just drop here to confirm when this bugfix is going to get released, and no problem now (I saw 4.3.2) !

thanks very much !!! ;)

PR #1581, opened 2018-06-08
docs: fix minor typo in readme

Loilo:
Changed

> the **later** uses Proxies

to

> the **latter** uses Proxies

I'm not a native speaker, so probably someone should take a look at this — but as far as I can tell this is the correct spelling. 🙃


mattruby:
Thanks!

Loilo:
My pleasure.

PR #1588, opened 2018-06-12
Update docs `mobx.expr` to `mobxUtils.expr`

alexlouden:
Now it matches https://github.com/mobxjs/mobx-utils


mweststrate:
thanks!

PR #1591, opened 2018-06-13
improvement: use allowStateChanges instead of manual control

kuitos:
A little bit code improvement


mweststrate:
Closing for now, code is indeed cleaner, but this is a quite performance critical path (called a lot, and this way of writing takes an additional closure and a try/finally block in `allowStateChanges`, which kills some optimizations on older JS engines), so leaving this as-is for now to avoid risking a performance drop in certain cases.

PR #1593, opened 2018-06-15
Do not reassign Symbol.asyncIterator in ts tests

dmitrage:
Currently 2 tests are failing with TypeError on node 10 because Symbol.asyncIterator is defined and not writable. Sorry for removing PR template but it looked irrelevant to this small change.


mweststrate:
Thanks!

PR #1601, opened 2018-06-20
Add support for obs. arrays on `keys`. Fixes #1600

### LINKED ISSUE #1600: [Feature request] Object API `keys` should also support observable arrays
All other object API functions support objects, arrays, and maps. I find it arbitrary that only the `keys` function does not support arrays. `array.keys()` is a thing; as well as `Object.keys(array)`.

I agree that it might seem like a useless functionality, but I believe it's worth it for API consistency: all functions should work on all types of collections.

I, in fact, have a use for `keys` where I'd like not to worry about the type of the underlying collection (thought I admit it is not a common use-case).

I'd be happy to provide a PR if you agree with my suggestion.

nunocastromartins:
I went ahead and just did it, feel free to refute it if you believe that the change shouldn't be made.
If you're willing to accept it, should I add an entry to the changelog?

PR checklist:
* [x] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/src/api/object-api.ts
+++ b/src/api/object-api.ts
@@ -15,6 +15,7 @@ import {
 } from "../internal"
 
 export function keys<K>(map: ObservableMap<K, any>): ReadonlyArray<K>
+export function keys<T>(ar: IObservableArray<T>): ReadonlyArray<number>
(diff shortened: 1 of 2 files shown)


mweststrate:
Looking great, thanks! Sorry for the slow follow up :)

Issue #1646, opened 2018-07-21
Suspense API support


nin-jin:
I have a Idea:
  * [x] What problem would it solve for you?
  * [x] Do you think others will benefit from this change as well and it should in core package (see also mobx-utils)?

# API

- Calculation can be stopped by throwing `Promise` instance.
- After promise is `resolved` or `rejected` calculation should be restarted automatically.

# Profit

- Synchronous code without async/await everywhere.
- Intercommunication with other libraries.
- Automatic tracking of `waiting` state (when we can't render until data loaded, we should render placeholder).

# Changes

1. Catched promises shouldn't be logged as errors.
2. First computed, which have caught promise should subscribe on promise rejection and resolving and initiate actualization after.

# Other libraries with Suspense API support:

- [ReactJS](https://medium.com/@baphemot/understanding-react-suspense-1c73b4b0b1e6) (slave)
- [$mol_fiber](https://github.com/eigenmethod/mol/tree/master/fiber) (master/slave)

nin-jin:
```
class User {
    firstName: string
    lastName: string
}
```

```
function lazy< Entity >( entity : Entity , fetch: ( field : PropertyKey )=> Promise<any> ) {
    return new Proxy( new Entity , {
        get : ( entity , field )=> {
            if( entity[ field ] === undefined ) throw fetch( field )
            return entity[ field ]
        } ,
    } )
}
```

```
const user = new User
const lazyUser = lazy( user , async ( field )=> {
    const json = await ( await fetch( '/user' ) ).json()
    Object.assign( user , json.data.user )
} )
```

```
function UserName( user : User ) {
    return <div>{ user.firstName } { user.lastName }</div>
}
```

function PageTitle( config : Config , user : User , page : Page ) {
    return <title>{ config.showName && `[${ user.firstName  } ${ user.lastName }] ` } { page.title }</title>
}

nin-jin:
Simple example (with mobx usage in master) in this test: https://github.com/mobxjs/mobx/pull/1647/commits/1220eee8b74ccc1552fd55af1fc8c4d09a9606c0#diff-19c935498f7daabdc3c5f9a02361e212
But this issue is about masters without mobx like example below.

mayorovp:
This behavior can be achieved by current mobx easily:

```js
function lazy< Entity >( entity : Entity , fetch: ( field : PropertyKey )=> Promise<any> ) {
    return new Proxy( entity , {
        get : ( entity , field )=> {
[...]

PR #1647, opened 2018-07-21
WIP: Don't log promises as errors

nin-jin:
#1646 

* [+] Added unit tests
* [ ] Updated changelog
* [+] Verified that there is no significant performance drop (`npm run perf`)


mayorovp:
Why you cannot set an errorHandler instead?

nin-jin:
Because it's not an error.

PR #1652, opened 2018-07-25
decorate - compose decorators for a single prop

ramybenaroya:
PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

## Changes
- Reduce multiple decorators to a single decorator
- Added a test for multiple decorators (@action + custom) on a function property
- Added a test for multiple decorators (@observable + @serializable) on
  a regular property
- Added a usage example (+ caveat warning) in readme under `decorate` section

Usage Example: 
```javascript
import { decorate, observable } from "mobx"
import { serializable, primitive } from "serializr"
import persist from 'mobx-persist';

class Todo {
    id = Math.random();
    title = "";
    finished = false;
}
decorate(Todo, {
    title: [serializable(primitive), persist('object'), observable],
    finished: observable
})
```


package.json
```diff
@@ -67,6 +67,7 @@
     "rollup": "^0.41.6",
     "rollup-plugin-filesize": "^1.3.2",
     "rollup-plugin-node-resolve": "^3.0.0",
+    "serializr": "^1.3.0",
```
ramybenaroya: I added `serializr` as a dev dependency just for this [test](https://github.com/mobxjs/mobx/pull/1652/files#diff-2492cdcb925c1d209eec08acc8ff1320R428), so I can compose `@observable` with another decorator `@serializable(primitive())`.

README.md
```diff
@@ -107,6 +107,24 @@ decorate(Todo, {
...
+    id = Math.random();
+    title = "";
+    finished = false;
+}
+decorate(Todo, {
+    title: [serializable(primitive), persist('object'), observable],
+    finished: [serializable(primitive), observable]
+})
+```
+Note: Composing decorators can sometimes lead to strange things when one decorator modifies the [property descriptor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) in a way that is unexpected for the other decorators.
```
ramybenaroya: Not so sure that my explanation is clear enough. Would love to get a feedback about it.

package.json
mweststrate: N.P.

README.md
mweststrate: Note: not all decorators can be composed together, and this functionality is just best-effort. Some decorators affect the instance directly and can 'hide' the effect of other decorators that only change the prototype

README.md
```diff
@@ -107,6 +107,24 @@ decorate(Todo, {
 })
 ```
 
+For applying multiple decorators on a single property, you can pass an array of decorators.
```
mweststrate: Please mension what the application order is. I assume serializable is applied first, then persist and finally observable?

mweststrate:
Some small comments, further looking great! 
Note that if you need to support this feature for MobX, a second PR targeting `mobx4-master` should be opened.

Thanks!

README.md
ramybenaroya: correct

ramybenaroya:
Thanks for the feedback @mweststrate.
> if you need to support this feature for MobX...     

I didn't get what you meant by that. Can you elaborate about "support for MobX"?

mweststrate:
@ramybenaroya sorry, meant MobX _4_. In other words; master represents 5, but 4 is still maintained as well so it might be interesting to backport this one. (if this branch is merged with squash, it can probably just be cherry-picked)

PR #1656, opened 2018-07-26
Update typings to allow flow to take async generators - async function*

dannsam:
`flow` already supports async generators yielding Promises, just changing typings to allow it.
Also changing from `IterableIterator` to `Iterator`:
`IterableIterator` implements `Iterator` with the addition of `[Symbol.iterator]` property.
Because `mobx` does not use `[Symbol.iterator]` property, it's not necessary to use `IterableIterator`.

PR to support it in Mobx 4 - https://github.com/mobxjs/mobx/pull/1655


mweststrate:
Thanks!

PR #1659, opened 2018-07-27
Rename decorators2.ts to decorators.ts

qcam:
Hello, thanks for making this awesome library!

As the title might imply, this PR renames `decorators2.ts` file to `decorators.ts`. I think this would make it less confusing when reading the code.

Or are there any special reasons to name it `decorators2`? Thank you.


mweststrate:
Nope! This is just a left over from the refactoring from mobx 3 to 4. Thanks!

PR #1686, opened 2018-08-15
Docs: decorate - compose decorators for a single prop

ramybenaroya:
Add usage example `decorate` util for new functionality introduced in https://github.com/mobxjs/mobx/pull/1652


mweststrate:
Thanks!

PR #1699, opened 2018-08-27
Refactor toJS

wangyiz4262:
PR checklist:

* [x] Added unit tests
* [x] Updated change log
-----
`toJS` will recursively convert an object containing Mobx observable value into a plain object

### DIFF
--- a/src/api/tojs.ts
+++ b/src/api/tojs.ts
@@ -23,20 +23,17 @@ function cache<K, V>(map: Map<any, any>, key: K, value: V, options: ToJSOptions)
 }
 
 function toJSHelper(source, options: ToJSOptions, __alreadySeen: Map<any, any>) {
-    if (!isObservable(source)) return source
+    if (typeof source !== "object") {
+        return source
+    }
 
     const detectCycles = options.detectCycles === true
 
-    if (
-        detectCycles &&
-        source !== null &&
-        typeof source === "object" &&
-        __alreadySeen.has(source)
-    ) {
+    if (detectCycles && source !== null && __alreadySeen.has(source)) {
         return __alreadySeen.get(source)
     }
 
-    if (isObservableArray(source)) {
+    if (isObservableArray(source) || Object.getPrototypeOf(source) === Array.prototype) {
         const res = cache(__alreadySeen, source, [] as any, options)
         const toAdd = source.map(value => toJSHelper(value, options!, __alreadySeen))
         res.length = toAdd.length
@@ -44,16 +41,7 @@ function toJSHelper(source, options: ToJSOptions, __alreadySeen: Map<any, any>) 
         return res
     }
 
-    if (isObservableObject(source)) {
-        const res = cache(__alreadySeen, source, {}, options)
-        keys(source) // make sure we track the keys of the object
-        for (let key in source) {
-            res[key] = toJSHelper(source[key], options!, __alreadySeen)
-        }
-        return res
-    }
-
-    if (isObservableMap(source)) {
+    if (isObservableMap(source) || Object.getPrototypeOf(source) === Map.prototype) {
         if (options.exportMapsAsObjects === false) {
             const res = cache(__alreadySeen, source, new Map(), options)
             source.forEach((value, key) => {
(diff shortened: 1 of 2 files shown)


wangyiz4262:
@mweststrate, thanks for the advice, will do!

wangyiz4262:
Hi @mweststrate, I have added an optional flag as you suggested. Please check it out and merge if it is OK. Thanks!

src/api/tojs.ts
```diff
@@ -1,15 +1,13 @@
-import {
-    isObservable,
-    isObservableArray,
-    isObservableMap,
-    isObservableObject,
-    isObservableValue,
-    keys
-} from "../internal"
+import { keys } from "./object-api"
```
mweststrate: Imports should be from `internal`

src/api/tojs.ts
```diff
@@ -23,37 +21,35 @@ function cache<K, V>(map: Map<any, any>, key: K, value: V, options: ToJSOptions)
...
-        source !== null &&
-        typeof source === "object" &&
-        __alreadySeen.has(source)
-    ) {
+    if (detectCycles && source !== null && __alreadySeen.has(source)) {
         return __alreadySeen.get(source)
     }
 
-    if (isObservableArray(source)) {
+    if (isObservableArray(source) || Array.isArray(source)) {
```
mweststrate: why was the second condition dropped here?

mweststrate: It is now cloned by the fallback below it seems, but then it is cloned into an object instead of array?

mweststrate: Best verify this with a unit test

test/base/tojs.js
```diff
@@ -298,9 +298,9 @@ test("verify #566 solution", () => {
     const b = mobx.observable({ x: 3 })
     const c = mobx.observable({ a: a, b: b })
 
-    expect(mobx.toJS(c).a === a).toBeTruthy() // true
-    expect(mobx.toJS(c).b !== b).toBeTruthy() // false, cloned
-    expect(mobx.toJS(c).b.x === b.x).toBeTruthy() // true, both 3
+    expect(mobx.toJS(c).a).toBe(a)
+    expect(mobx.toJS(c).b).toEqual(mobx.toJS(c.b))
```
mweststrate: for completeness sake, restore the previous:

`expect(mobx.toJS(c).b).not.toEqual(b)`

src/api/tojs.ts
```diff
@@ -1,15 +1,13 @@
...
+import { keys } from "./object-api"
+import { isObservable } from "./isobservable"
+import { isObservableArray } from "../types/observablearray"
+import { isObservableValue } from "../types/observablevalue"
+import { isObservableMap } from "../types/observablemap"
 
 export type ToJSOptions = {
     detectCycles?: boolean
     exportMapsAsObjects?: boolean
+    recurseEverything?: boolean
```
mweststrate: I think it would be benificial to give a warning (or simply force) to set `detectCycles` when `recurseEverything` is used? Not sure

test/base/tojs.js
wangyiz4262: Reverted

src/api/tojs.ts
wangyiz4262: Thanks for raising it. However, if the source is either a plain array or observable array, it will be converted to plain array with the flag set to be true. I've added tests for this scenario.

src/api/tojs.ts
wangyiz4262: I would prefer to force `detectCycles` to be true with presence of `recurseEverything` since it for sure will traverse deeply into all the levels, and it would be meaningful to stop when cycles are encountered. I will incorporate the changes in next commit. Thanks for idea!

wangyiz4262:
Hi @mweststrate All comments have been addressed. And changes are also available for mobx4 in PR [#1733](https://github.com/mobxjs/mobx/pull/1733)

mweststrate:
Thanks, merged! Will be part of next release

mweststrate:
Released as 5.5.1. Sorry for the late release! Missed that a PR was pending for release

stnwk:
Thanks! :)

PR #1706, opened 2018-08-30
add allowStateChangesInsideComputed to allow changes inside a computed value

xaviergonz:
in order to fix MST https://github.com/mobxjs/mobx-state-tree/issues/967


xaviergonz:
I changed the implementation so it only allows it for the current computed level, this is
computed -> changes state inside allowStateChangeInsideAction - ok
computed -> changes state inside allowStateChangeInsideAction -> computed -> changes state without allowStateChangeInsideAction - error

mweststrate:
@xaviergonz I just pushed an alternative solution, not sure whether I do like it better (I do like it better from api perspective, but not sure I like it better from the constraint that is being given up)

xaviergonz:
@mweststrate Thanks for taking a look!
TBH I'd rather not give up on the checks (that are useful for all users) altogether just for this feature (that's an internal one most probably only to be used by MST) TBH

src/core/action.ts
```diff
@@ -104,3 +104,15 @@ export function allowStateChangesStart(allowStateChanges: boolean) {
...
 }
+
+export function allowStateChangesInsideComputed<T>(func: () => T): T {
+    const prev = globalState.computationDepth
+    globalState.computationDepth = 0
+    let res: T
+    try {
+        res = func()
+    } finally {
+        globalState.computationDepth += prev
```
xaviergonz: @mweststrate is there any chance where computationDepth might not be 0 here? I didn't know if I should put += or = here

xaviergonz:
@mweststrate should it go back to the original solution (the one that doesn't give up the constraint) in order to fix mobx-state-tree afterCreate?

mweststrate:
Sorry, yeah that is anyway better than the current one. Still don't really
like it, nasty side effects of laziness... 😋.

Op ma 24 sep. 2018 19:14 schreef Javier Gonzalez <notifications@github.com>:

xaviergonz:
Well, it is only to fix that very particular use case :) Don't think people will (or even should) use it too much.
Reverted

xaviergonz:
Improved the unit test to show that:
computed -> afterCreate (change) will be ok
computed -> afterCreate (change) -> computed (other change) will throw

xaviergonz:
I was wondering, does this need to be retrofit to mobx 4 master as well?

mweststrate:
@xaviergonz yes, but no worries, already on it :)

xaviergonz:
ah thanks!

mattruby:
Thanks for keeping on this one!

PR #1724, opened 2018-09-14
Prevent rollup from breaking on `process.env.NODE_ENV`

Lipus86:
update for prevent error when rollup-plugin-commonjs is used

### DIFF
--- a/src/mobx.ts
+++ b/src/mobx.ts
@@ -41,7 +41,7 @@ try {
         process.env.NODE_ENV !== "production"
     ) {
         console.warn(
-            "[mobx] you are running a minified build, but 'process.env.NODE_ENV' was not set to 'production' in your bundler. This results in an unnecessarily large and slow bundle"
+            '[mobx] you are running a minified build, but "process.env.NODE_ENV" was not set to "production" in your bundler. This results in an unnecessarily large and slow bundle'
         )
     }
 })()
--- a/src/mobx.ts
+++ b/src/mobx.ts
@@ -41,7 +41,7 @@ try {
         process.env.NODE_ENV !== "production"
     ) {
         console.warn(
-            '[mobx] you are running a minified build, but "process.env.NODE_ENV" was not set to "production" in your bundler. This results in an unnecessarily large and slow bundle'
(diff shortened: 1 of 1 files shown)


mweststrate:
Ouch, looks like a nasty problem!

How does this fail? Does rollup still replace the thing inside the string? And that won't happen with the quotes reversed? I would make sure to file a bug report to rollup as well and crosslink it here. 

Alternatively your solution could dom something like `["process", "env", "NODE_ENV"].join(".")`, I think that would be more robust? As the quote change is quite implicit and might break in the future again, or with other tools. (or prettier or a linter might just change the quotes back). Also make sure to leave a comment what is happening here, to prevent someone else to "clean it up" in the future :-)

Lipus86:
Thanks, changed by your comment
and created the issue for [rollup-plugin-commonjs ](https://github.com/rollup/rollup-plugin-commonjs/issues/344)

mweststrate:
Could you add a comment to the code as well, linking to the rollup issue? So we know why we do that weird stuff string joining? Not necessary, but using backtick strings would make this code look nicer :)

mweststrate:
Stupid doublecheck: you already checked if this solves the issue for you?

mweststrate:
Beyond that approved btw.

Lipus86:
yes, it solves my issue

mweststrate:
Ok, thanks!

mweststrate:
Released as: 5.1.2 / 4.4.2

thdk:
I have mobx 5.6 and i see the following line in mobx.module.js

```
if (testCodeMinification.name !== "testCodeMinification" &&
        process.env.NODE_ENV !== "production") {
        console.warn(
        // Template literal(backtick) is used for fix issue with rollup-plugin-commonjs https://github.com/rollup/rollup-plugin-commonjs/issues/344
        "[mobx] you are running a minified build, but 'process.env.NODE_ENV' was not set to 'production' in your bundler. This results in an unnecessarily large and slow bundle");
    }
```

The comment regarding the template literal is there, but the line below still uses the double quotes instead of backticks.

Did this fix get lost somewhere?

Note: The string **process.env.NODE_ENV** is being replaced with **'development'** or **'production'** a plugin rollup-plugin-replace.

mweststrate:
Please open new issues rather than commenting on closed issues / PRs

Op di 20 nov. 2018 om 08:33 schreef Thomas Dekiere <notifications@github.com

Issue #1650, opened 2018-07-23
object with 'toString' key is ignored by ObservableArray


Ixanin:
Hi,  I can't  assign object with 'toString' key to observable array.

data from server
![data from server](https://user-images.githubusercontent.com/9035056/43067198-efd2cff2-8e66-11e8-9e3c-8228352a7531.png)

observable array
![observable array](https://user-images.githubusercontent.com/9035056/43067201-f36ca1d8-8e66-11e8-9184-cfeabe3987ca.png)

**demo**: https://codesandbox.io/s/1yxyzwjz14
I am using: "mobx": "^4.3.1", "mobx-react": "5.2.2",

urugator:
Minimal repro: https://codesandbox.io/s/pppj0njj8q
`toString` is defined on target object as observable prop, but it has "default descriptor" so I quess there is a call like: `Object.defineProperty(target, "toString", undefined)` somewhere...

mayorovp:
The bug lives here: [generateObservablePropConfig](https://github.com/mobxjs/mobx/blob/mobx5/src/types/observableobject.ts#L345)

mweststrate:
Confirmed as bug (sorry for the late follow up!), investigating a solution

mweststrate:
Fix has been released as: 5.1.2 / 4.4.2

PR #1725, opened 2018-09-14
Change property descriptor caches to null-prototype objects

mayorovp:
probably fixes issue #1650


mweststrate:
@mayorovp awesome, nice find!

mweststrate:
Confirmed solution, released as: 5.1.2 / 4.4.2

PR #1728, opened 2018-09-17
Update installation.md

akraines:
typo: mobx 15 => 5


mweststrate:
Thanks!

Issue #1526, opened 2018-04-30
[Mobx >= 4] Object.hasOwnProperty fails


rennerDa:
I don't know if this is a bug in mobx or babel or anything else.
I tested this with MobX 3.6 and earlier versions and it worked. It breaks with Mobx >= v4.

I uses this repository (https://github.com/mobxjs/mobx-react-boilerplate) and changed the following files:
.babelrc
```
{
  "presets": [
    "es2015",
    "react",
    "stage-0"
  ],
  "plugins": ["transform-decorators-legacy",
      "transform-class-properties",
      "transform-runtime"]
}
```

package.json
```
{
  "name": "mobx-react-boilerplate",
  "version": "1.0.0",
  "description": "Boilerplate for MobX + React project with JSX, ES6 and decorator compilation",
  "scripts": {
    "start": "webpack-dev-server --hot --open",
    "lint": "eslint src"
  },
  "repository": {
    "type": "git",
    "url": "https://github.com/mobxjs/mobx-react-boilerplate.git"
  },
  "keywords": [
    "react",
    "reactjs",
    "boilerplate",
    "mobx",
    "starter-kit"
  ],
  "author": "Michel Weststrate <mweststrate@gmail.com> (http://github.com/mweststrate)",
  "license": "MIT",
  "bugs": {
    "url": "https://github.com/mobxjs/mobx/issues"
  },
  "homepage": "http://mobxjs.github.com/mobx",
  "devDependencies": {
    "babel": "^6.23.0",
    "babel-core": "^6.26.3",
    "babel-eslint": "^8.2.1",
    "babel-jest": "^22.4.3",
    "babel-loader": "^7.1.4",
    "babel-plugin-react-intl": "^2.4.0",
    "babel-plugin-transform-class-properties": "^6.24.1",
    "babel-plugin-transform-decorators-legacy": "^1.3.4",
    "babel-plugin-transform-runtime": "^6.23.0",
    "babel-polyfill": "^6.26.0",
    "babel-preset-es2015": "^6.24.1",
    "babel-preset-react": "^6.24.1",
    "babel-preset-stage-0": "^6.24.1",
    "babel-preset-stage-3": "^6.24.1",
    "babel-runtime": "^6.26.0",
    "webpack": "^3.5.5",
    "webpack-dev-server": "^2.7.1",
    "mobx-react-devtools": "^5.0.1"
  },
  "dependencies": {
    "mobx": "^4.2.0",
[...]

Issue #1352, opened 2018-02-16
Support the new Babel 7 decorator implementation


mweststrate:
Example config for babel 7, in decorators legacy mode: 

```json
//.babelrc
{
    "presets": ["@babel/preset-env"],
    "plugins": [
        ["@babel/plugin-proposal-decorators", { "legacy": true }],
        ["@babel/plugin-proposal-class-properties", { "loose": true }]
    ]
}
```

Please note that plugin ordering is important, and `plugin-proposal-decorators` should be the _first_ plugin in your plugin list

```package.json
"devDependencies": {
    "@babel/core": "^7.1.0",
    "@babel/plugin-proposal-class-properties": "^7.1.0",
    "@babel/plugin-proposal-decorators": "^7.1.0",
    "@babel/preset-env": "^7.1.0"
}
```

Non-legacy mode decorators (stage 2) is work in progress, see #1732 

_Edit: updated config to show the non-beta configuration for babel 7_

brandonhall:
Just wanted to share my requirements for setting up `mobx` with `babel 7.0.0-beta.40`. Mainly, I wanted to highlight the necessary inclusion of `babel-plugin-transform-decorators-legacy`:

```
// package.json
  "devDependencies": {
    "@babel/core": "^7.0.0-beta.40",
    "@babel/preset-env": "^7.0.0-beta.40",
    "@babel/preset-react": "^7.0.0-beta.40",
    "@babel/preset-stage-1": "^7.0.0-beta.40",
    "babel-plugin-transform-class-properties": "^6.24.1",
    "babel-plugin-transform-decorators-legacy": "^1.3.4",
     ....
  }

// using webpack.config.js instead of babel.rc
presets: [
  '@babel/preset-react',
  '@babel/preset-env',
  '@babel/preset-stage-1',
],
plugins: [
  'react-hot-loader/babel',
  'transform-decorators-legacy',
  [
    'transform-class-properties', { loose: true, },
  ],
],
```

Kielan:
I really like decorators, any way to keep it?

und3fined:
My config

```
// package.json
"@babel/core": "^7.0.0-beta.42",
    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.42",
    "@babel/plugin-proposal-decorators": "^7.0.0-beta.42",
    "@babel/plugin-transform-runtime": "^7.0.0-beta.42",
    "@babel/preset-env": "^7.0.0-beta.42",
[...]

Issue #1273, opened 2017-12-13
Can't make @observable parameter properties


JoshuaKGoldberg:
I have an:

2. [x] Issue:
  * [x] **Provide error messages including stacktrace**
  * [x] Provide a **minimal** sample reproduction.
  * [x] Did you check this issue wasn't filed before?
  * [x] Elaborate on your issue. What behavior did you expect?
  * [x] State the versions of MobX and relevant libraries. Which browser / node / ... version?

MobX version: 3.3.1
TypeScript version: 2.7.0-dev.20171126

```typescript
import { observable } from "mobx";

type Model = { };

export class ViewModel {
    public constructor(
        // Argument of type 'number' is not assignable to parameter of type 'PropertyDescriptor | undefined'.
        @observable
        public readonly model: Model,
    ) { }
}
```

On the other hand, works:

```typescript
export class ViewModel {
    @observable
    public readonly model: Model;

    public constructor(model: Model) {
        this.model = model;
    }
}
```

JoshuaKGoldberg:
Not sure if https://github.com/Microsoft/TypeScript/issues/10777 is relevant here.

mweststrate:
We might implement this ones decorators are standardized. For now blocked until then as there is an easy workaround.

mweststrate:
Closing for now, will be reopened once #1928 happens.

PR #1732, opened 2018-09-20
[on hold] WIP: Support decorators stage 2 proposal, implements #1352

mweststrate:
* [x] #1731 should be merged first
* [x] make all tests pass
* [x] make sure decorate / extend observable prefer stage2 decorator implementation
* [ ] add additional assertions to the stage2 decorators
* [x] document change in behavior: computed properties now become visible as _own_ property
* [x] document change in behavior: some non-observable fields will unexpectedly non-enumerable, see babel bug
* [X] ~document bug: `isObservable` / `isObservableProp` etc will yield incorrect results for members that were never visited (or create workaround via `finisher`?)~ Found workaround
* [ ] make sure that tests also succeed in legacy mode (swap out .babelrc or different .babelrc in sub dir?)
* [ ] backport to MobX 4
* [x] follow up on late initialization issues: https://github.com/tc39/proposal-decorators/issues/153
* [x] Investigate using `["@babel/plugin-proposal-class-properties", { "loose": false }]`, or document that `true` should be used! as `loose: false` will wipe the properties and redefine them (when using `decorate` 
 instead of decorators)
* [x] ~await fix for non-decorated field enumerability, see: https://github.com/babel/babel/issues/8760~ Fixed, released with next babel release
* [x] Document that class-properties proposal should be used in loose mode if the `decorate` utility is used
* [ ] update gh-pages with syntax settings
* [ ] verify against mobx-state-tree
* [x] check performance impact. Before -> After: (babel create update / TS create update) `200 / 6300 / 3700 / 5700` => `1800 / 5000 / 3800 / 6100`
* [ ] Implement `flow` and `flow.bound` as decorator: #1769 / #1767 
* Follow up on 
  * [ ] #1526 
  * [ ] #1352 
  * [ ] #1273 
* [ ] create follow-up issue, in MobX 6 legacy decorators will not be supported (4 will stay as is supporting both). But not before TS followed suit
* [x] wait until key-less initializers are implemented in babel: https://github.com/tc39/proposal-decorators/pull/165, waiting for: https://github.com/babel/babel/issues/9260 / 
* [ ] wait until initializers have been renamed to hooks: https://github.com/babel/babel/pull/9360/files (see https://github.com/tc39/proposal-decorators/pull/199 for the naming changes)


CHANGELOG.md
```diff
@@ -1,3 +1,37 @@
...
+* By supporting the stage2 implementation, it is now possible to use `@babel/plugin-proposal-class-properties` in spec mode (that is, with option: `{ "loose": false }`). However, the `spec` mode breaks using the `decorate` utility on class instances, so for now it is recommended to keep using `{ "loose": true }` in your babel configuration
+* The value of te `decoratorsBeforeExport` setting of `plugin-proposal-decorators` is technically indifferent to MobX, but it is recommended to set it to `true`, this is in line with the legacy syntax and all current MobX documentation. With `true`, one can write: `@observer export class MyComponent extends React.Component {...`, with `false`, this needs to be rewritten as: `export @observer class MyComponent extends React.Component {...`. ([background](https://github.com/tc39/proposal-decorators/issues/69))
```
nicolo-ribaudo: Maybe instead of recommending `decoratorsBeforeExport: true` you could just say that it is similar to the old behavior but that it could change? Since TC39 hasn't decided yet it would be better not to recommend neither `true` not `false`.

src/utils/stage2decorators.ts
```diff
@@ -0,0 +1,13 @@
...
+    key: string
+    placement: "static" | "prototype" | "own"
+    descriptor: PropertyDescriptor
+    initializer?: () => any
+    finisher?: (klass) => void
+    extras?: Stage2Decorator[]
+}
+
+export function quacksLikeAStage2Decorator(args: IArguments): boolean {
+    return args.length === 1 && args[0] && (args[0].kind === "field" || args[0].kind === "method")
```
nicolo-ribaudo: I think that it would be more future-proof to also check if `args[0].kind === "class"`, just in case someone tries to use this helper with a class decorator.

test/base/babel-tests.js
```diff
@@ -663,11 +670,32 @@ test("enumerability - workaround", () => {
...
+    class Todo2 {
+        id = 1
+        @dec x
+    }
+
+    const t1 = new Todo1()
+    expect(Object.getOwnPropertyDescriptor(t1, "id").enumerable).toBe(true)
+
+    const t2 = new Todo2()
+    expect(Object.getOwnPropertyDescriptor(t2, "id").enumerable).toBe(true) // fails
```
nicolo-ribaudo: I believe that this is fixed in Babel 7.1.3

mweststrate:
Current main blocker: wait until https://github.com/tc39/proposal-decorators/pull/165 is implemented in babel.

nicolo-ribaudo:
I will implement it as soon as that PR is merged in the proposal repo.

mweststrate:
@nicolo-ribaudo awesome, thanks!

IngwiePhoenix:
Is the decorator-plugin so much in WIP that you can not use it without legacy being turned on? I am asking because I am - sort of desparately? - trying to get class-level decorators to work, but have had little to no luck. I managed to get method decorators to work, but that's it. I can not seem to be able to attach a field to a class through a decorator.

"legacy" sounds like "old", or rather "do not use for long-term support". I would like to get started on decorators with sort of an LTS in sight, so that I can stay on that implementation for at least a while.

This is the test-file I am trying to compile:
```js
function classDecorator(value) {
  //console.log("class", value, arguments.length, arguments)
  return function decorator(target) {
    let obj = {
      kind: "field",
      key: "isDecorated",
      placement: "static",
      descriptor: {
        configurable: false,
        enumerable: false,
        writeable: false,
      },
      method: () => {console.log("class rt mtd"); return true},
      value: true
    }
    target.elements.push(obj)
    console.log(target)
    return target
  }
}

function methodDecorator(value) {
  //console.log("method", value, arguments.length, arguments)
  return (target) => {
    //console.log("method rt", arguments)
    target.descriptor.value.isDecorated = true
    return target
  }
}

@classDecorator("class val")
class Bar {
  @methodDecorator("mtd val")
  baz() {
    console.log("foo")
  }
}

console.log(Bar, Bar.isDecorated, Bar.prototype.isDecorated)
// Expected to see the class and "true".
```

And this is the `.babelrc`:
```json
{
  "presets": [
    ["@babel/preset-env", {
      "targets": { "node": "current" }
    }]
  ],
  "plugins": [
    ["@babel/plugin-proposal-decorators", {
      "decoratorsBeforeExport": true
    }]
  ]
}
```

mweststrate:
Eh one step back, what is the relation with MobX?

Op ma 21 jan. 2019 18:53 schreef Ingwie Phoenix <notifications@github.com>:

test/base/babel-tests.js
mweststrate: Confirmed

mweststrate:
Latest status: initializers haven't landed yet in babel, since the proposal was slightly changed. Waiting for https://github.com/babel/babel/pull/9360/files to land.

mweststrate:
Update: the proposal is undergoing significant changes again. Waiting on that: https://github.com/tc39/proposal-decorators/pull/250

mweststrate:
Closing for now, will be reopened once #1928 happens.

PR #1735, opened 2018-09-21
Fix windows compilation + remove nscript due to fibers dep

xaviergonz:
This PR:
- removes nscript due to a dependency with fibers, a native node module that was kinda hard to compile (and actually impossible with new node versions due to compilation errors)
- fixes the publish build script so it doesn't use such dependency and uses shelljs instead
- fixes publish and build scripts so they work in windows as well

I think this should help new contributors get up and running faster


xaviergonz:
yeah, prettier run over it

xaviergonz:
master conflicts fixed

mweststrate:
Thanks! Will try to check in couple of days, currently focused on getting
Babel 7 support as far as possible :-).

Op ma 24 sep. 2018 om 19:13 schreef Javier Gonzalez <
notifications@github.com>:

mweststrate:
@xaviergonz could you undo the package.json formatting? Otherwise it is pretty hard to review the changes :-) (and it will get reformatted by yarn / npm on any future dependency install)

xaviergonz:
@mweststrate done, also added package.json to prettierignore

mweststrate:
Thanks @xaviergonz!

PR #1771, opened 2018-10-11
Change onBecome(Un)Observed implementation to work with multiple listeners

### LINKED ISSUE #1537: onBecome(Un)Observed intercept hooks can remove each other
The way `onBecomeObserved` and `onBecomeUnobserved` are implemented means they can inadvertently cancel each other. 

Source of the issue is here:
https://github.com/mobxjs/mobx/blob/master/src/api/become-observed.ts#L44

If you add a hook, the behaviour is 'call the original hook, then the new one'. The disposer function that is returned has the behavior of 'restore the original hook'. 

This means that if a hook is disposed (with the return value), any subsequent hooks added since that was added will also be removed. Example here:
https://codesandbox.io/s/k0y70x7vy7

(You'd expect one of the two callbacks be called, but it isn't because the first was removed).

I started a fix for this, but the solution wasn't as immediately straightforward as I thought and since I don't know when I'll have time to give it another go I wanted to record this for now.

michalwarda:
This PR fixes #1537 

PR checklist:
* [x] Added unit tests
* [ ] Updated changelog

In this PR I've changed the implementation of onBecomeObserved and onBecomeUnobserved to allow for unsubscribing previous subscription. 
In the new implementation I keep the listeners in Sets rather than create a chain of calls as previously.
Also for performance reasons I've created custom getters that lazily initialise the Sets.

**This might break external libraries that implement on IObservable.** I thought about instead of changing the interface of IObservable, add a custom `getter` and `setter` to `onBecomeObserved` and `onBecomeUnobserved` to keep the previous interface but I decided to wait for a hint from you to decide on that.

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -27,6 +27,18 @@ export class Atom implements IAtom {
     diffValue = 0
     lastAccessedBy = 0
     lowestObserverState = IDerivationState.NOT_TRACKING
+
+    private observedListeners?: Set<Function>
+    private unobservedListeners?: Set<Function>
+
+    get onBecomeObservedListeners(): Set<Function> {
+        if (this.observedListeners === undefined) this.observedListeners = new Set()
+        return this.observedListeners
+    }
+    get onBecomeUnobservedListeners(): Set<Function> {
+        if (this.unobservedListeners === undefined) this.unobservedListeners = new Set()
+        return this.unobservedListeners
+    }
     /**
      * Create a new atom. For debugging purposes it is recommended to give it a name.
      * The onBecomeObserved and onBecomeUnobserved callbacks can be used for resource management.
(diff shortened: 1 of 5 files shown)


src/core/atom.ts
```diff
@@ -27,18 +27,30 @@ export class Atom implements IAtom {
     diffValue = 0
     lastAccessedBy = 0
     lowestObserverState = IDerivationState.NOT_TRACKING
+
+    private observedListeners?: Set<Function>
+    private unobservedListeners?: Set<Function>
+
+    get onBecomeObservedListeners(): Set<Function> {
+        if (this.observedListeners === undefined) this.observedListeners = new Set()
```
mweststrate: The getter causes the set to be initializaed lazily, but effectively for every atom, correct (as MobX core will often inspect this getter)? So this means that every atom now will have an additional `Set` to allocate. 

I would prefer to see a solution that doesn't allocate the Set (or just a shared, empty, frozen one), until the first listeners is established, and only then instantiate the local Set of listeners

mweststrate:
Closed in favor of #1833

PR #1772, opened 2018-10-11
Add  "never", "always" and "observed" to enforceActions

alexandersorokin:
Add  "never", "always" and "observed" possible values to enforceActions configuration parameter.
It's because true, false and "strict" now deprecated.


mweststrate:
Good catch, thanks!

PR #1783, opened 2018-10-22
Add check for nullable source

wangyiz4262:

PR checklist:

* [x] Added unit tests

### DIFF
--- a/src/api/tojs.ts
+++ b/src/api/tojs.ts
@@ -27,6 +27,9 @@ function toJSHelper(source, options: ToJSOptions, __alreadySeen: Map<any, any>) 
     if (!options.recurseEverything && !isObservable(source)) return source
 
     if (typeof source !== "object") return source
+    
+    // Directly return null if source is null
+    if (source === null) return null
 
     // Directly return the Date object itself if contained in the observable
     if (source instanceof Date) return source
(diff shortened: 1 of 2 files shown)


src/api/tojs.ts
```diff
@@ -27,16 +27,17 @@ function toJSHelper(source, options: ToJSOptions, __alreadySeen: Map<any, any>)
     if (!options.recurseEverything && !isObservable(source)) return source
 
     if (typeof source !== "object") return source
+    
+    // Directly return null if source is null
+    if (source === null) return null
```
ItamarShDev: won't it return null already in line#29?

src/api/tojs.ts
```diff
@@ -27,16 +27,17 @@ function toJSHelper(source, options: ToJSOptions, __alreadySeen: Map<any, any>)
...
     // Directly return the Date object itself if contained in the observable
     if (source instanceof Date) return source
 
     if (isObservableValue(source)) return toJSHelper(source.get(), options!, __alreadySeen)
 
     // make sure we track the keys of the object
-    if (isObservable(source)) {
-        keys(source)
-    }
+    if (isObservable(source)) keys(source)
```
ItamarShDev: removing brackets in if statements if not a healthy idea

wangyiz4262: @ItamarShDev I know, but it seems that this project is following this pattern...

src/api/tojs.ts
wangyiz4262: @ItamarShDev No, if you run `typeof null` in the console, you will see null is also a unique type of object. I didn't foresee this edge case in my previous PR.

src/api/tojs.ts
ItamarShDev: OK. deleting the comment :)

src/api/tojs.ts
ItamarShDev: thank you very much for the explanation!

PR #1791, opened 2018-10-27
@computed({keepAlive: true}) should calculate only when accessed

hector:
First, I split the existing tests trying to keep 1 test <=> 1 requirement.
Then I added the tests that check that no calculations are performed before accessing the computed.
We need somebody to implement the solution.

References issue #1534

* [x] Added unit tests
* [ ] Updated changelog


mweststrate:
Thanks for the test cases @hector!

mweststrate:
Thanks @hector, the tests are merged into #1799 which will be released soon

PR #1793, opened 2018-10-29
support setting arbitrary array properties

nathanstitt:
Since a proxy should behave like the original array it should also support setting properties just like a array will.

Without this the below code raises a `proxy: trap returned falsish for property "foo"`
```js
const ary = observable.array([])
Object.defineProperty(ary, 'foo', {
  set() { 
    // do something wonderful
  }
})
ary.foo = 'bar'
```

I agree it's pretty ugly to define arbitrary properties on built-in objects, but it works fine for "real" arrays so it seems like it should also be supported on Proxys.

Note that setting without a defined property such as the below example will still throw the exception but won't with a real array.   If we want to support that I can remove the 'hasOwnProperty` check and just set the property, then return true.
```js
const ary = observable.array([])
ary.foo = 'bar' // throws proxy: trap returned falsish for property "foo"
```

----

PR checklist:

* [x] Added unit tests
~* [ ] Updated changelog~ (I don't think this should be publicized)
~* [ ] Updated docs (either in the description of this PR as markdown, or as separate PR on the `gh-pages` branch. Please refer to this PR). For new functionality, at least [API.md](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md) should be updated~
* [x] Added typescript typings (not needed)

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -104,6 +104,10 @@ const arrayTraps = {
             arrayExtensions.set.call(target, parseInt(name), value)
             return true
         }
+        if (target.hasOwnProperty(name)) {
+            target[name] = value
+            return true
+        }
         return false
     },
     preventExtensions(target) {
@@ -166,15 +170,17 @@ class ObservableArrayAdministration
         fireImmediately = false
     ): Lambda {
         if (fireImmediately) {
-            listener(<IArraySplice<any>>{
-                object: this.proxy as any,
-                type: "splice",
-                index: 0,
-                added: this.values.slice(),
-                addedCount: this.values.length,
-                removed: [],
-                removedCount: 0
-            })
+            listener(
+                <IArraySplice<any>>{
+                    object: this.proxy as any,
+                    type: "splice",
+                    index: 0,
+                    added: this.values.slice(),
+                    addedCount: this.values.length,
+                    removed: [],
+                    removedCount: 0
+                }
+            )
         }
         return registerListener(this, listener)
     }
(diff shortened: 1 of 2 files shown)


mweststrate:
@nathanstitt thanks for the suggestion. However, I would prefer to keep arrays conceptually simple, and using an array with arbitrarily keys sounds like to me like a code smell anyway.

More importantly, allowing this raises a bunch of new questions, like
* How is the same achieved in MobX 4
* What does it mean for built in utilities like `entries`, `values`, `has`, `set` etc
* how do modifiers affect newly introduced properties etc.

In other words, I think that this is technically possible, but raises more questions, edge cases and issues than actually problem it solves, so closing this PR for that reason.

nathanstitt:
Thanks for your consideration @mweststrate, I think I didn't express clearly enough why this was an issue and how it changes the existing behavior of arrays.

In my case I hit this because my code was marking objects as processed by a renderer by setting a readonly `$wasProccessed` property on them.  That's worked fine through mobx 3 & 4 , but suddenly started throwing an exception with mobx5 because the proxy was returning false when the property was accessed.

I agree that there's a whiff of code smell with this approach and an alternative would be to store the object as a key in a weak map.

However, I also I think there's a good case to be made that mobx5 proxy's should behave as closely as possible to the underlying object, which in this case they do not.  In this case the error is quite confusing because you can create and set the property without errors,  and the exception will occur later when you access it.

To answer you bullets directly:
 * How is the same achieved in MobX 4
    * In mobx4 you can just set the property directly i.e. `arrayOfFoo.$wasProccessed = true`, however my code was using Object.defineProperty
 * What does it mean for built in utilities like entries, values, has, set etc
    * They won't know anything about the newly created property, and the property won't be observable, but I think that's fine.  I don't think we need to support observability, we just shouldn't cause exceptions for otherwise valid code.
 * how do modifiers affect newly introduced properties etc.
    * I'm pretty sure modifiers will cause errors, but as above I think that's fine.  All I'm after is not causing exceptions for otherwise valid code 😁 

Again, thanks for taking the time to ponder all this.  If you still disagree that's completely fine, but I think you may get some strange bug reports about `proxy: trap returned falsish for property ...`.

Issue #1805, opened 2018-11-09
Computed Properties Are Evaluated Unnecessarily


mjewell:
In certain cases, computed properties are re-evaluated despite that their values will never be used. I have only been able to reproduce it in the context of `observer`, but it may be possible using only plain mobx. While this is generally not an issue because computed properties should not have side effects, in my case it occurred when using mobx-state-tree, and that re-evaluation caused a read of a property in a mobx-state-tree node that was no longer alive, causing an error. This code sandbox demonstrates the example both with and without mobx-state-tree: https://codesandbox.io/s/yqrl5vmnw1

In the sandbox, the important thing to notice is the console logs look like this:

```
render app 
render child 
compute y 
stopped watching 
compute y 
render app
```

In both examples, we render the child component only once, and the child component is the only place that the computed value is referenced. Despite that, we trigger `compute y` twice. Stopping watching causes the app component to not render the child, but `compute y` still happens after that. In the MST example, you can see that this problem causes this error to occur:

`[mobx-state-tree][error] You are trying to read or write to an object that is no longer part of a state tree. (Object type was 'Child'). Either detach nodes first, or don't use objects after removing / replacing them in the tree. Use setLivelynessChecking("error") to simplify debugging this error.`

It appears as though most of the time mobx is intelligent enough to determine whether these properties need to be re-evaluated, but in this case it fails. Would it be possible to delay the evaluation of a computed property until it is actually accessed for the first time after one of its dependent observables changes instead of immediately after noticing the change?

urugator:
Firstly, provided reproduction requires a little tweek:
```javascript
// render is not sync, invoke action in callback
ReactDOM.render(<App />, rootElement, () => {
  counter.incrementX();
});  
```
For the actual problem:
Basically your assumption is that the `y` doesn't have to be recalculated, because the child reaction (depending on `y`) is disposed as a result of re-running the app (parent) reaction (depending on `watch`).
Meaning that one reaction depeds on the effect of another, however such behavior is not compatible with batching mechanism.
Both reactions (app and child) are invalidated before they have a chance to run. To find out whether child reaction should be invalidated, `y` must be recalculated first.
Both components are therefore scheduled for update, but react makes sure the parent component runs first and thefore prevets re-rendering of child - event though it's update is scheduled first.  

> instead of immediately after noticing the change

It actually isn't immediately (it happens after `stopped watching` even though `a` is modified before), but it's before any reaction effect (rendering) is fired.

[...]

PR #1811, opened 2018-11-14
idea: new reaction with delayed dependency check

mayorovp:
see issue #1805

The problem of unnecessary evaluations is early shouldCompute call. While reaction can be delayed via scheduler or onInvalidate - shouldCompute is always called immediatelly. But what if we delay this call?

New reaction class can be used in React observers this way:

```js
this[mobxReaction] = new ReactionEx(`${initialName}#${rootNodeID}.render()`, () => this.setState({}));

// ...

shouldComponentUpdate: function(nextProps, nextState) {
    return !shallowEqual(this.state, nextState) 
        || !shallowEqual(this.props, nextProps)
        || this[mobxReaction].shouldCompute()
}
```

### DIFF
--- a/src/core/globalstate.ts
+++ b/src/core/globalstate.ts
@@ -1,4 +1,4 @@
-import { IDerivation, IObservable, Reaction, fail } from "../internal"
+import { IDerivation, IObservable, IScheduledReaction, fail } from "../internal"
 
 /**
  * These values will persist if global state is reset
@@ -59,7 +59,7 @@ export class MobXGlobals {
     /**
      * List of scheduled, not yet executed, reactions.
      */
-    pendingReactions: Reaction[] = []
+    pendingReactions: IScheduledReaction[] = []
 
     /**
      * Are we currently processing reactions?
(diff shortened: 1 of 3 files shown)


urugator:
Like it.

Could `shouldComponentUpdate` be delayed so much that it would somehow mess with mobx synchronous nature?

Can we somehow skip prop/state check, when the component is invalidated via Mobx (specific state field perhaps)?

mayorovp:
> Could shouldComponentUpdate be delayed so much that it would somehow mess with mobx synchronous nature?

Computeds already can delay shouldCompute call forever.

> Can we somehow skip prop/state check, when the component is invalidated via Mobx (specific state field perhaps)?

We can move `this[mobxReaction].shouldCompute()` call before this checks.

mweststrate:
@mayorovp tip: if you fix the single test failing `api.js`, it will make this PR look more solid :). (PR's that have failing builds tend to be ignored, as they are assumed WIP)

mayorovp:
@mweststrate but this PR **is** WIP

ItamarShDev:
@mayorovp is it still WIP?

mweststrate:
Closing (can be further discussed in #1805).

PR #1817, opened 2018-11-19
Clarify autorun disposer

jasonk:
The current documentation says that the function passed to autorun will receive a disposer function as it's only argument, but the examples show the disposer being returned by autorun.  Experimentation has shown the examples to be correct, so this fixes the docs to match.


mattruby:
Thanks!

Issue #1816, opened 2018-11-19
Upgrade to TS 3 and improve flow() typing


ismailhabib:
The flow() typing is suboptimal since you we have to choose to either provide the type of the output (and having to manually provide all the argument types), or have the argument types inferred but use `Promise<any>` for the output. Inferring is also not possible due to the typing of Generator (this is a long discussion in TS project but it's a different matter).

Fortunately, with TS 3 there is this new generic rest parameters which not only will improve the typing but also make it much simpler. See: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#generic-rest-parameters. I'm willing to create an MR once mobx is upgraded to TS 3.

Thanks!

mweststrate:
Master now uses TS 3, feel free to contribute!

ismailhabib:
https://github.com/mobxjs/mobx/pull/1825

mweststrate:
released in 4.8.0 / 5.8.0

mweststrate:
Thanks for the PR!

PR #1825, opened 2018-12-02
Improve flow() typing

ismailhabib:
See: https://github.com/mobxjs/mobx/issues/1816

### DIFF
--- a/src/api/flow.ts
+++ b/src/api/flow.ts
@@ -10,32 +10,9 @@ export interface FlowIterator<T> {
     throw?(e?: any): IteratorResult<T> | Promise<IteratorResult<T>>
 }
 
-export function flow<R>(generator: () => FlowIterator<any>): () => CancellablePromise<R>
-export function flow<A1>(
-    generator: (a1: A1) => FlowIterator<any>
-): (a1: A1) => CancellablePromise<any> // Ideally we want to have R instead of Any, but cannot specify R without specifying A1 etc... 'any' as result is better then not specifying request args
-export function flow<A1, A2>(
-    generator: (a1: A1, a2: A2) => FlowIterator<any>
-): (a1: A1, a2: A2) => CancellablePromise<any>
-export function flow<A1, A2, A3>(
-    generator: (a1: A1, a2: A2, a3: A3) => FlowIterator<any>
-): (a1: A1, a2: A2, a3: A3) => CancellablePromise<any>
-export function flow<A1, A2, A3, A4>(
-    generator: (a1: A1, a2: A2, a3: A3, a4: A4) => FlowIterator<any>
-): (a1: A1, a2: A2, a3: A3, a4: A4) => CancellablePromise<any>
-export function flow<A1, A2, A3, A4, A5>(
-    generator: (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5) => FlowIterator<any>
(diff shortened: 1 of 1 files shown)


ismailhabib:
Impact: before this MR, flow() only need a generic type when the passed function does not have any parameter. Now, it's always required (for a good reason).

src/api/flow.ts
```diff
@@ -10,32 +10,9 @@ export interface FlowIterator<T> {
...
-export function flow<A1, A2, A3, A4, A5, A6, A7>(
-    generator: (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7) => FlowIterator<any>
-): (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7) => CancellablePromise<any>
-export function flow<A1, A2, A3, A4, A5, A6, A7, A8>(
-    generator: (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7, a8: A8) => FlowIterator<any>
-): (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7, a8: A8) => CancellablePromise<any>
-export function flow(generator: Function) {
+export function flow<T, U extends any[]>(
+    generator: (...args: U) => FlowIterator<any>
+): CancellablePromise<U> {
```
mayorovp: Why `<U>`?

ismailhabib: No significant meaning here. It can also be changed to <R, A> to follow the previous naming.

mayorovp: No, it cannot. Right result type is `(...args: U) => CancellablePromise<any>`

ismailhabib: You're right, I missed the `() =>` part. Strange that no tests are failing.

xaviergonz:
There's a way to make flow properly detect the return argument as well (except when the return argument is a promise, but that can be circumvented with a fake function such as castFlowReturn or similar)

See https://github.com/mobxjs/mobx-state-tree/blob/5775437e5267281906ca2ef73a49e7e1bbb410bc/packages/mobx-state-tree/src/core/flow.ts for example

src/api/flow.ts
```diff
@@ -10,32 +10,9 @@ export interface FlowIterator<T> {
...
-export function flow<A1, A2, A3, A4, A5, A6, A7>(
-    generator: (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7) => FlowIterator<any>
-): (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7) => CancellablePromise<any>
-export function flow<A1, A2, A3, A4, A5, A6, A7, A8>(
-    generator: (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7, a8: A8) => FlowIterator<any>
-): (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5, a6: A6, a7: A7, a8: A8) => CancellablePromise<any>
-export function flow(generator: Function) {
+export function flow<T, U extends any[]>(
+    generator: (...args: U) => FlowIterator<any>
+): (...args: U) => CancellablePromise<U> {
```
mayorovp: `<U>` again?

ismailhabib: fixed

PR #1827, opened 2018-12-04
Improved flow typing (alternate version)

xaviergonz:
* [x] Added unit tests
* [x] Updated changelog

This PR is kind of like #1825 but with some improvements, since the other PR makes the user give the type of return types and arguments.

This pr:
* Argument and return types will be correctly inferred automatically. The only exception is when the last returned value is a Promise, but in this case a function named `castFlowReturn` was added to allow this be inferred properly in typescript (note however that this should be a very rare case). This also coherent with the current typing of flow in Mobx-state-tree.
* Flow will now use as iterator the standard IterableIterator type rather than its custom one

Additionally:
* Adds tslint-config-prettier to disable tslint rules covered by prettier already
* Removes some unneeded rules (due to prettier usage) from tslint.json

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,77 +1,78 @@
+*   Improved flow support in typescript. It will now infer arguments and return values correctly. The only exception is that when the `flow` wrapped function returns a `Promise` then it should `return castFlowReturn(somePromise)` to keep the type inference happy.
+
 # 5.7.0 / 4.7.0
 
-* Upgraded typings to TypeScript 3
-* Fixed [#1742](https://github.com/mobxjs/mobx/issues/1742), change detection fails when multiple mobx instances were active.
-* Fixed [#1624](https://github.com/mobxjs/mobx/issues/1624), use built-in flow types for iterators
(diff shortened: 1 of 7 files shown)


xaviergonz:
Future: once TS implements Partial type argument inference (https://github.com/Microsoft/TypeScript/pull/26349) then "arguments" argument could be set to be inferred and then the user would be able to only specify the return type (in this case it would only still be needed for promises) and castFlowReturn could become deprecated.

mayorovp:
But how nested promise even possible?..

xaviergonz:
In case the async action wants to return a promise as result, but it should be a really really rare use case. Still supported though

mayorovp:
But this is not supported by runtime!

    > new Promise(resolve => resolve(Promise.resolve(42)))
    < Promise {<resolved>: 42}

xaviergonz:
ahh, so when a flow actually returns a promise it waits for it and returns its result!
I'll make changes then

xaviergonz:
fixed and updated the top comment

xaviergonz:
Another option of course is just to remove castFlowReturn and force typescript users to yield over it and return the yielded result (now that I know that actually returned promises are awaited first)

xaviergonz:
Removed castFlowReturn as the user can just write instead
```
const x = yield Promise.resolve(2);
return x as number;
```

mweststrate:
LGTM, Thanks!

mweststrate:
Released in 4.9.0 / 5.9.0

PR #1833, opened 2018-12-06
fix: earlier onBecome(Un)Observed disposer will not dispose later listener (fix #1537)

### LINKED ISSUE #1537: onBecome(Un)Observed intercept hooks can remove each other

fi3ework:

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

This PR is to fix #1537, in `become-observed.ts`, there is a shared object to store the `onBecomeObserved` and `onBecomeUnobserved` for each Atom instance by name. Each Set of the hook of an atom will be initialized only when `onBecomeUnobserved` or `onBecomeUnobserved` invoked. So I think there's no harm to the performance.

Looking forward to your opinion 😆

Special thanks to #1771

### DIFF
--- a/src/api/become-observed.ts
+++ b/src/api/become-observed.ts
@@ -38,16 +38,25 @@ function interceptHook(hook: "onBecomeObserved" | "onBecomeUnobserved", thing, a
     const atom: IObservable =
         typeof arg2 === "string" ? getAtom(thing, arg2) : (getAtom(thing) as any)
     const cb = typeof arg2 === "string" ? arg3 : arg2
-    const orig = atom[hook]
+    const listenersKey = `${hook}Listeners`
 
+    if (!atom[listenersKey]) {
+        atom[listenersKey] = new Set<Lambda>([cb])
+    } else {
+        ;(atom[listenersKey] as Set<Lambda>).add(cb)
+    }
+
+    const orig = atom[hook]
     if (typeof orig !== "function")
         return fail(process.env.NODE_ENV !== "production" && "Not an atom that can be (un)observed")
 
-    atom[hook] = function() {
-        orig.call(this)
-        cb.call(this)
-    }
     return function() {
-        atom[hook] = orig
+        const hookListeners = atom[listenersKey] as Set<Lambda> | undefined
+        if (hookListeners) {
+            hookListeners.delete(cb)
+            if (hookListeners.size === 0) {
+                delete atom[listenersKey]
+            }
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


mayorovp:
I see two problems on this code:

1. atom names is not guaranteed to be unique;
2. properties on static object `atomsObservedListeners` are never deleted, so this code is source of memory leaks.

mayorovp:
User can give any name to atoms manually.

mayorovp:
3. Who calls `orig` now? Looks like original handler is forgotten

fi3ework:
Oh.. It's a fatal mistake I did not realize. Thanks, I' ll try to resolve it.

fi3ework:
This pr has been rewrote. Welcome review.

src/api/become-observed.ts
```diff
@@ -38,16 +38,25 @@ function interceptHook(hook: "onBecomeObserved" | "onBecomeUnobserved", thing, a
     const atom: IObservable =
         typeof arg2 === "string" ? getAtom(thing, arg2) : (getAtom(thing) as any)
     const cb = typeof arg2 === "string" ? arg3 : arg2
-    const orig = atom[hook]
+    const listenersKey = `${hook}Listeners`
+
+    if (!atom[listenersKey]) {
+        atom[listenersKey] = new Set<Lambda>([cb])
+    } else {
+        ;(atom[listenersKey] as Set<Lambda>).add(cb)
```
mayorovp: Why you used type assertion instead of editing `IObservable` interface?

Now the type system allows existence of observable implementation which does not know about listeners...

fi3ework:
Thanks for reviewing. The pr has updated, there's still an `as` because I don't know how to infer `${hook}Listeners` to be ` "onBecomeObservedListeners" | "onBecomeUnobservedListeners"` gracefully so far😅

mweststrate:
Thanks @fi3ework !

PR #1848, opened 2018-12-15
Add clarification about faux-arrays

scottmas:
Faux arrays don't play nicely with native methods like concat. Updated docs.


mweststrate:
I couldn't entirely understand the sentences you are proposing / it is unclear to me why you are using ``[].concat(observable([]))` and not `observableArray.concat(` ? But I think something went wrong with editing as the sentences look like they are in the middle of some other sentences. Could you take a second look?

mweststrate:
Thanks!

PR #1851, opened 2018-12-21
Replace comparer.deep with comparer.structural in the computed decorator documentation

vkrol:


mweststrate:
Good catch, thanks!

PR #1863, opened 2019-01-07
fix: inherit prop from subclass of ES6 Map (fix #1858)

### LINKED ISSUE #1858: Cannot extend Map
Hey,
Mobx behaves incorrectly with classes extended from build-ins like Map<K, V>
For example (TS)

```
class StringToStringMap extends Map<string, string> {
  public findOne(id: string): Optional<string> {
    /// code
  }
 }

class Store  {
  @observable
  public stringToStringMap: StringToStringMap  = new StringToStringMap(); 
  // actual type is ObservableMap, method findOne is lost

  @action
  public doStuffWithString(someString: string) {
    this.stringToStringMap.findOne(someString); // throws an error TypeError: findOne is not a function
  } 
}
```

fi3ework:

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

Copy prop from prototype of merge source to prototype of target. Prop already exist will not be override to protect props of observable map.  (fix #1858)

### DIFF
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -291,6 +291,21 @@ export class ObservableMap<K = any, V = any>
         for (const [key, value] of this) callback.call(thisArg, value, key, this)
     }
 
+    mergeProtoProp(initialData: IObservableMapInitialValues<K, V>) {
+        const copyProp = (from, to) => {
+            // do not override exists prop to protect props of observable map
+            const existProps = Object.getOwnPropertyNames(to)
+            Object.getOwnPropertyNames(from).forEach(key => {
+                if (existProps.indexOf(key) === -1) {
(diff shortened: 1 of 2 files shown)


mweststrate:
This won't work, as it would affect *all* observable maps, not just this instance. So to avoid that, you would need to generate a new class on the fly. I'll add an error to just forbid this behavior (otherwise it would be inconsistent with for example arrays as well)

The more maintainable solution is to throw. If you need shared functionality between maps, I recommended to use static / global utility functions instead.

fi3ework:
Yes, this pr only support extends once. Too many inheritances can lead to complex attribute override rules. So I just merge the prototype of instance. Leaving the method merge to user is a better choice!

fi3ework:
Glad to help!

PR #1862, opened 2019-01-06
feat: add `equals` option to `observable.box` (resolve #1580)

fi3ework:
Thanks for taking the effort to create a PR!

If you are creating an extensive PR, you might want to open an issue with your idea first, so that you don't put a lot of effort in an PR that wouldn't be accepted. Please prepend pull requests with `WIP: ` if they are not yet finished
PR checklist:

* [x] Added unit tests
* [ ] Updated changelog
* [ ] Updated docs (either in the description of this PR as markdown, or as separate PR on the `gh-pages` branch. Please refer to this PR). For new functionality, at least [API.md](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md) should be updated
* [ ] Added typescript typings
* [x] Verified that there is no significant performance drop (`npm run perf`)

Try to add `equals` to the options of `observable.box`.
Changelog & docs update will be updated later 😀

PR #1874, opened 2019-01-21
Fi3ework equals for box

mweststrate:
Follow up on #1862


mweststrate:
released in 4.9.0 / 5.9.0

PR #1877, opened 2019-01-23
[Docs] Clarification for reacting to mutation.

bfeigin:
In the example given the array is mutated with push implying a mutation will trigger a reaction, but the description claims the opposite. 
This change fixes the description to align with the code shown. 

PR checklist:
* [ N/A ] Added unit tests
* [ ? ] Updated changelog 
* [ N/A ] Added typescript typings
* [ N/A ] Verified that there is no significant performance drop (`npm run perf`)

Questions:
Happy to update the change log but seemingly isn't needed for the minor doc update?

### DIFF
--- a/docs/best/react.md
+++ b/docs/best/react.md
@@ -201,7 +201,7 @@ autorun(() => {
 message.likes.push("Jennifer");
 ```
 
-This will react as expected. All array functions that do not mutate the array are tracked automatically.
+This will react as expected. All array functions that mutate the array are tracked automatically.
 
 ---
 


ItamarShDev:
but the working example _does not_ mutate the array, `join` returns a new data and doesn't mutate the array itself.

So why the change?

bfeigin:
@ItamarShDev you're totally correct, I misunderstood the context for that sentence as referencing the push instead of the function wrapped by autorun. Thank you, I'll close this.

Issue #1887, opened 2019-01-30
Action.object in spy utility


rdrezner:
Hi there.

I have a:

1. [x] Question: 

We are using MobX in our project. In one of our files we are using `spy` utility to track actions. Recently I upgraded from version 3 to 4 and came across an issue that I would like to ask you about. In MobX version 3 `action.object` in the `spy` utility for bound actions is keeping a reference to the object in which an action was defined. In MobX version 4 `action.object` seems to be keeping a reference to the object in which an action was called.

Is this an expected behavior or a bug? I couldn't find any information about it in the migration guidelines. Is it possible to get the reference to the object in which bound action was defined when using `spy` util in MobX 4? I need this to make my code working again. 

Here is the reproduction (click a checkbox and see the log in the console):
- MobX 3: https://jsfiddle.net/rdrezner/4Lo9swu6/2/ (action.object = TodoView)
- MobX 4: https://jsfiddle.net/rdrezner/9vL2xr6k/1/ (action.object = null)

Thanks in advance for your answers.

mweststrate:
Confirmed, looks like a regression bug indeed

ItamarShDev:
dibs :)

ItamarShDev:
@mweststrate seems that the issue lies here: https://github.com/mobxjs/mobx/blame/e5178f2fc3ab0a83048b71633601484922bd1c7b/src/core/action.ts#L19

ItamarShDev:
closing as the PR solving it merged to main branch.
@rdrezner thank you for submitting the issue!

PR #1894, opened 2019-02-06
[action/createAction] - Keep observable reference.

ItamarShDev:
fix issue #1887

### DIFF
--- a/src/api/action.ts
+++ b/src/api/action.ts
@@ -66,7 +66,7 @@ export var action: IActionFactory = function action(arg1, arg2?, arg3?, arg4?): 
     // @action fn() {}
     if (arg4 === true) {
         // apply to instance immediately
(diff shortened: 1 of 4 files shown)


ItamarShDev:
@mweststrate Not sure if this is the best solution.
Will be happy for some input

mweststrate:
Thanks for picking up! I think the code would be slightly easier to follow, if `scope` is an optional third argument to `createAction`, and pass it in from the relevant `defineBoundAction` calls explicitly. (Possibly the `bind` there becomes optional as well then).  

Could you also add some unit tests verifying the original bug?

Thanks a lot!

ItamarShDev:
Sure, will do

ItamarShDev:
@mweststrate updated, what do you think about this approach?

mweststrate:
@ItamarShDev looking good! approved

PR #1899, opened 2019-02-09
fix: type error of TypeScript

fi3ework:

PR checklist:

* [ ] Added unit tests
* [ ] Updated changelog

Run `yarn run quick-build` will fail because of some type errors. This PR fix them and quick-build works fine now 🤩

### DIFF
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -302,11 +302,10 @@ export class ObservableMap<K = any, V = any>
             else if (Array.isArray(other)) other.forEach(([key, value]) => this.set(key, value))
             else if (isES6Map(other)) {
                 if (other.constructor !== Map)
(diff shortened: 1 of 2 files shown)


src/types/observablemap.ts
```diff
@@ -306,6 +306,7 @@ export class ObservableMap<K = any, V = any>
                 other.forEach((value, key) => this.set(key, value))
             } else if (other !== null && other !== undefined)
                 fail("Cannot initialize map from " + other)
+            return
```
mweststrate: NIT: I think dropping `return` here and on line 305 woudl achieve the same. (`fail` always throws)

mweststrate:
Thanks! Looking good

src/types/observablemap.ts
fi3ework: Agreed. This would be better as there is nothing need to be returned 😀

mweststrate:
Thanks!

PR #1917, opened 2019-03-01
Make ObservableSet backport compatible with IE11

kalmi:
The original implementation and the backport made use of Array.from.
Mobx4 appears to avoid the usage of Array.from by instead using its
internal iteratorToArray function. All Array.from calls have been
replaced by iteratorToArray calls.
Also, IE11 lacks the "values" function on Sets,
and even ObservableSet.toString ended up depending on it.
A workaround has been implemented that allows the
ObservableSet.values to work even in IE11.

### DIFF
--- a/src/api/object-api.ts
+++ b/src/api/object-api.ts
@@ -12,6 +12,7 @@ import {
     endBatch,
     defineObservableProperty,
     invariant,
+    iteratorToArray,
     getAdministration,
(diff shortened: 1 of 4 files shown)


mweststrate:
Great, thanks a lot!

mweststrate:
Released as 4.9.3, thanks!

PR #1930, opened 2019-03-14
Allow Symbol keys in observable maps (fixes #1925)

### LINKED ISSUE #1925: Can't use symbol keys in ObservableMap
Any use of symbols as a key in an observable map throws an exception, complaining because symbols can't be implicitly convert into strings.

* Error messages including stacktrace:
	```
	Cannot convert a Symbol value to a string
	    at ObservableMap$$1../node_modules/mobx/lib/mobx.module.js.ObservableMap$$1._updateHasMapEntry (app.js:113461)
	    at ObservableMap$$1../node_modules/mobx/lib/mobx.module.js.ObservableMap$$1.has (app.js:113392)
	    at ObservableMap$$1../node_modules/mobx/lib/mobx.module.js.ObservableMap$$1.get (app.js:113518)
	```

* Provide a minimal sample reproduction
	```js
	
	const map = observable.map();
	map.get(Symbol('test'))
	```

* [x] Did you check this issue wasn't filed before?
  I don't think this has been filed before, but it is a bit related to https://github.com/mobxjs/mobx/issues/1809 (but I'm interested in keys in ES6 maps specifically, not object property keys which I suspect are much more complicated)

* [x] Elaborate on your issue. What behavior did you expect?
  I'd like to be able to use Symbols as map keys, and I don't think that's an especially uncommon thing to do.

  I think the main reason this fails is the use of the key [in string templates](https://github.com/mobxjs/mobx/blob/master/src/types/observablemap.ts#L177), because `` `${Symbol('test')}` `` always throws the error above. It would work with `` `${Symbol('test').toString()}` `` instead. I haven't tested that though, so there may well be other issues here as well.
 
* [x] State the versions of MobX and relevant libraries. Which browser / node / ... version?
  MobX 5.6.0, Chrome 70

pimterry:
* [x] Added unit tests
* [x] Updated changelog
	- Not clear what the 'right' way to do this is, seems lots of PRs ignore it? I've added an 'Unreleased' section ready to rename for the next release for now.
	- No changes necessary imo
	- No types have changed

---

@mweststrate you mentioned `safeStringify` in #1925, I couldn't find that - looks like it's [a function in the mobx-state-tree codebase](https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/core/type/type-checker.ts#L38-L45)?

I've written a separate key stringification function instead. Here it actually doesn't use JSON.stringify, because JSON can't represent symbols, and [they're normally omitted from JSON stringified objects](https://esdiscuss.org/topic/json-stringify-and-symbols). For the same reason, I've also preserved symbol-ness in `toPOJO` here, instead of stringifying them, which feels correct, and helpfully means we get the correct behaviour (symbol keys omitted) in `toJSON` here too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Allow symbol keys in `ObservableMap`, see [#1930](https://github.com/mobxjs/mobx/pull/1930) by [pimterry](https://github.com/pimterry)
+
 # 4.9.3
 
 * Fixed `observable.set` compatibility with IE 11, see [#1917](https://github.com/mobxjs/mobx/pull/1917) by [kalmi](https://github.com/kalmi)
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -174,7 +174,12 @@ export class ObservableMap<K = any, V = any>
         if (entry) {
             entry.setNewValue(value)
         } else {
-            entry = new ObservableValue(value, referenceEnhancer, `${this.name}.${key}?`, false)
+            entry = new ObservableValue(
+                value,
+                referenceEnhancer,
+                `${this.name}.${stringifyKey(key)}?`,
+                false
+            )
             this._hasMap.set(key, entry)
         }
         return entry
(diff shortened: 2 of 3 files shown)


pimterry:
@mweststrate any thoughts on this one? I'd love to get this fix in.

mweststrate:
@pimterry looking great! Merging and releasing. I think this should be pack-ported to mobx4 too?

mweststrate:
Backported, released as 4.9.4 / 5.9.4

pimterry:
Amazing, thank you!

PR #1931, opened 2019-03-17
update tooling (rollup, etc) (ignore red x, it is coveralls)

xaviergonz:
This kind of big PR is all about tooling:
* updated all dev deps to latest versions (except for flow and chalk)
* fixed configurations to work with the changes
* changed the build script so it only uses rollup + plugins (no more envify, browserify, uglify, etc)
* since rollup uglify uses terser rather than the buggy uglify-es the code is less likely to have weird minification bugs
* changed some eslint settings that were conflicting with TS itself
* fixed eslint errors on tests (mostly about the usage of var and unused variables)
* fixed ts tests not actually using the proper tsconfig

no actual source code was changed other than tests (AFAI remember) :)

note that the decrease in coverage is most probably the update of jest doing a better job at detecting uncovered lines


mweststrate:
@xaviergonz in the category of "if it ain't broken, don't fix it", any special highlights that make it important to merge this? In general I don't have any trouble swapping tools around, but in MobX's case it is a little bit more trickier, since every change on master is cherry-picked to the mobx4 branch, it is really valuable to not touch or divert too many things (or, obviously, do it on both). That is for example the reason why the tests are all still in `es5 / var` style.

mweststrate:
I think I'm just asking, how hard is it to do the same on the mobx4-master branch :see_no_evil:

xaviergonz:
windows compatibility mostly :) It is a bit annoying to be forced to switch to WSL just to work on mobx.

Also using terser for minification I'd say is a big plus, since uglifyJs has had bugs that haven't been fixed in like... forever, and those have bitten me more than once in production versions, while I still haven't had a single issue with terser.

How hard? dunno, I could try to cherry pick it into the mobx4 branch and see what happens?

mweststrate:
Yeah, feel free

xaviergonz:
Similar pr open for mobx 4

mweststrate:
Merged! Thanks!

PR #1960, opened 2019-05-08
Updated PropertyDescriptor type to use the latest flow core definition

tbezman:
We're updating our MobX to version 4 at Coinbase. We came across an issue with the flow typings for `PropertyDescriptor`. The first commit is us just running prettier. The second commit is where we bring in the Flow core definition of `PropertyDescriptor`.

Can we make sure this is released under version 4?

### DIFF
--- a/flow-typed/mobx.js
+++ b/flow-typed/mobx.js
@@ -3,31 +3,31 @@
 export type IObservableMapInitialValues<K, V> = IMapEntries<K, V> | KeyValueMap<V> | IMap<K, V>
 
 export interface IMobxConfigurationOptions {
-    +enforceActions?: boolean | "strict" | "never" | "always" | "observed",
-    computedRequiresReaction?: boolean,
-    isolateGlobalState?: boolean,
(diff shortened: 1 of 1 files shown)


ItamarShDev:
can you please remove the prettier commit? 
prettifying better be done in a single PR to be as clean as possible

thank you

capaj:
@ItamarShDev if we don't squash it it will be displayed as two separate commits in the git history-who cares if it's 2 PRs or 1? The source of truth is the git history, not the github PR history. 
Also I think we have at least a few PRs we have merged that mixed changes to functionality and formatting so I don't think this is a rule we follow here in mobx org.

mweststrate:
@tbezman ideally keep prettier out next time indeed (or do it as separate PR), makes reviewing and merging harder. I'll try to cherry-pick the descriptor commit to the master branch as well, hope it will work out :)

mweststrate:
Released as 4.10.0 / 5.10.0

PR #1964, opened 2019-05-11
fix update array event using this as object instead of proxy

xaviergonz:
I just noticed that array "splice" would return the array proxy as object, yet "update" would return the array itself
This PR fixes that

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -305,175 +305,175 @@ class ObservableArrayAdministration
 }
 
 const arrayExtensions = {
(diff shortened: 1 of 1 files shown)


ItamarShDev:
Seems OK
any possible side affects we can think of?
maybe add some tests?( i am not sure they will really add any value)

xaviergonz:
None that I can think of, since the other one was updated to use proxy I can imagine this one was not just due to an oversight

ItamarShDev:
Let's merge then :)

xaviergonz:
merged, don't think this needs backporting anyway since mobx 4 does not use proxies

mweststrate:
Released as 4.10.0 / 5.10.0

PR #1966, opened 2019-05-14
fix uglifyjs preamble (fixes #1921)

### LINKED ISSUE #1921: License header is removed in minified production build
We were doing a compliance check on our code base and realized that MobX license header is stripped off by minifier.

According to webpack-contrib/uglifyjs-webpack-plugin#222, to solve this at least for Webpack + UglifyJs code base we can add `@license`, `@preserve` or `/*!`

This is something I am willing to make PR for, but what I would like to have opinions on is what to put in the license header because as many bytes will be added to the (default) production build.

runnez:
Hi there. It fixes removing license header from minified version according this issue https://github.com/mobxjs/mobx/issues/1921
Details https://github.com/mishoo/UglifyJS2/issues/1985

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -61,6 +61,8 @@ function runTypeScriptBuild(outDir, target, declarations) {
 
 const rollupPlugins = [require("rollup-plugin-node-resolve")(), require("rollup-plugin-filesize")()]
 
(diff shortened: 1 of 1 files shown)


ItamarShDev:
Cool, thanks!

PR #1975, opened 2019-05-26
Update size-limit

pustovalov:
from v.`0.9`: Deprecate argument and option for Babili.
https://github.com/ai/size-limit/blob/master/CHANGELOG.md#09

Example:
```
$ size-limit
  Size limit:   20 KB
  Package size: 15.41 KB with all dependencies, minified and gzipped
  Loading time: 309 ms   on slow 3G
```

https://travis-ci.org/pustovalov/mobx/builds/537320499

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -1,0 +1,7 @@
+[
+    {
+        "path": "lib/mobx.js",
+        "limit": "20 KB",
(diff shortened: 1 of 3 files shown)


mweststrate:
Merged by hand, thanks for the PR!

PR #1981, opened 2019-05-29
fix TS for observable set

xaviergonz:
fixes TS typings for this case
```
const obsSet: ObservableSet = observable(new Set<any>())
```


mweststrate:
Released as 4.10.0 / 5.10.0

PR #1985, opened 2019-06-04
Fix array update proxy (important fix)

xaviergonz:
Apparently my fix to send the array proxy in the update event wasn't ok.
I didn't take into account that "set" is called with the actual proxied array as "this", therefore adm.proxy has to be used instead

Without this fix change.object in array update events is set to "undefined" :-/

### DIFF
--- a/src/types/observablearray.ts
+++ b/src/types/observablearray.ts
@@ -305,175 +305,175 @@ class ObservableArrayAdministration
 }
 
 const arrayExtensions = {
-    intercept(handler: IInterceptor<IArrayWillChange<any> | IArrayWillSplice<any>>): Lambda {
-        return this[$mobx].intercept(handler)
-    },
+        intercept(handler: IInterceptor<IArrayWillChange<any> | IArrayWillSplice<any>>): Lambda {
+            return this[$mobx].intercept(handler)
+        },
 
-    observe(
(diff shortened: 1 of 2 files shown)


xaviergonz:
@mweststrate sorry to bother you, but I think it is an important fix that might require a bugfix release (sorry!)

src/types/observablearray.ts
```diff
@@ -444,7 +444,7 @@ const arrayExtensions = {
             if (hasInterceptors(adm)) {
                 const change = interceptChange<IArrayWillChange<any>>(adm as any, {
                     type: "update",
-                    object: this.proxy,
+                    object: adm.proxy as any, // since "this" is the real array we need to pass its proxy
```
aretecode: shouldn't this be tested to make sure it doesn't happen again and to show what it fixed?

xaviergonz: unit test added

mweststrate:
Approved

@ItamarShDev, interested in releasing this as a patch on 5.10? Doesn't need a back-port. Invoking `scripts/publish.sh` should do the trick. Let me know your npm name if permissions are lacking :)

Thanks!

ItamarShDev:
Sure! Will try soon
@mweststrate

xaviergonz:
Just published it myself since it was taking a bit too long, hope you don't mind

PR #1989, opened 2019-06-05
Removed dangling link to www.faire.com

thomaschriskelly:
I assume this was left here unintentionally @mweststrate


FredyC:
Sounds about right :)

PR #1993, opened 2019-06-10
modify types

JSlife23:
I found two TS type checking errors.

### DIFF
--- a/src/types/observableobject.ts
+++ b/src/types/observableobject.ts
@@ -87,7 +87,7 @@ export class ObservableObjectAdministration
         public name: string,
         public defaultEnhancer: IEnhancer<any>
     ) {
-        this.keysAtom = new Atom(name + ".keys")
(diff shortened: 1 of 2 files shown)


src/types/observableobject.ts
```diff
@@ -87,7 +87,7 @@ export class ObservableObjectAdministration
         public name: string,
         public defaultEnhancer: IEnhancer<any>
     ) {
-        this.keysAtom = new Atom(name + ".keys")
+        this.keysAtom = new Atom(name + ".keys") as IAtom
```
FredyC: Perhaps it might be better to set up an actual property `private keysAtom: IAtom` so it's not necessary to force type here.

FredyC: @JSlife23 Bump? :)

mweststrate: It is already declared as IAtom, so cast shouldn't be needed.

mweststrate:
Merged!

Issue #1995, opened 2019-06-10
[Bug] Uncaught TypeError: Cannot read property 'get' of undefined at extendObservableObjectWithProperties


RosenTomov:
Since ver. 5.10 I'm getting the following error:

`Uncaught TypeError: Cannot read property 'get' of undefined at extendObservableObjectWithProperties`

It looks like descriptor (from `Object.getOwnPropertyDescriptor()`) is null (`descriptor.get`):
```
function extendObservableObjectWithProperties(target, properties, decorators, defaultDecorator) {
    if (false) { var key, i, keys; }
    startBatch();
    try {
        var keys = getPlainObjectKeys(properties);
        for (var i in keys) {
            var key = keys[i];
            var descriptor = Object.getOwnPropertyDescriptor(properties, key);
            if (false) {}
            var decorator = decorators && key in decorators
                ? decorators[key]
                : descriptor.get // <-- descriptor is null
                    ? computedDecorator
                    : defaultDecorator;
            if (false)
                {}
            var resultDescriptor = decorator(target, key, descriptor, true);
            if (resultDescriptor // otherwise, assume already applied, due to `applyToInstance`
            )
                Object.defineProperty(target, key, resultDescriptor);
        }
    }
    finally {
        endBatch();
    }
}
```

https://github.com/mobxjs/mobx/blob/6c1d807a44dfc32567f3e55231eb1464a2155826/src/api/extendobservable.ts#L102

Version 5.9.4 works fine.

Thanks.

aretecode:
https://github.com/mobxjs/mobx/issues/1994
^ as with this issue, both have issues with the `getPlainObjectKeys`

recover758126:
> #1994
> ^ as with this issue, both have issues with the `getPlainObjectKeys`
The parameter `properties` of `getPlainObjectKeys` and `getOwnPropertyDescriptor `  cannot be `null` or `undefined`,  is it necessary to call the`extendObservableObjectWithProperties ` function ?

stolenng:
Getting the same error, any solutions ?

ItamarShDev:
@FredyC do you know the status of this issue?

FredyC:
@ItamarShDev Well, I tagged it that we need reproduction. Otherwise, it's guesswork probably.

PR #1988, opened 2019-06-05
Revert mobx 4 symbol support

mweststrate:
I backported symbol support from MobX 5 to MobX 4, without thinking it really through. Although symbols could potentially be used in practice, this probably misses the point that mobx 4 is used in more backward compatible modi, and didn't require polyfills as of yet. 

This PR reverts symbol support. A downside of this PR is that this increases the semantical difference between mobx 4 and 5

Fixes #1986, #1987

PR #1998, opened 2019-06-11
fix(extendObservable): use regular `for` on array of keys

kobiburnley:
use regular `for` on array of keys

seems that it caused bugs in v5.10 (in my case I was extending the Array prototype with enumerable properties, and those properties were iterated because of `for i in keys`. 

can add a test, not sure where is the right place or if it's required for such a small change.

### DIFF
--- a/src/api/extendobservable.ts
+++ b/src/api/extendobservable.ts
@@ -67,7 +67,7 @@ export function extendObservableObjectWithProperties(
         )
         if (decorators) {
             const keys = getPlainObjectKeys(decorators)
-            for (let i in keys) {
+            for (let i = 0; i < keys.length; i++) {
                 const key = keys[i]
                 if (!(key in properties!))
                     fail(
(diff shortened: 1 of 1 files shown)


aretecode:
https://github.com/mobxjs/mobx/issues/1995

kobiburnley:
Not sure why automation decreased or how to fix this decrease

FredyC:
You can see what made it decrease here: https://coveralls.io/builds/23939994/source?filename=src/api/extendobservable.ts#L107

kobiburnley:
> You can see what made it decrease here: https://coveralls.io/builds/23939994/source?filename=src/api/extendobservable.ts#L107

There is not difference except total number of lines from master:
https://coveralls.io/builds/23891775/source?filename=src/api/extendobservable.ts

I guess it's related to how `for(const key of keys)` is compiled to JS

FredyC:
@kobiburnley Can you please make another PR for `mobx4-master` branch if it's relevant? I am not honestly sure what's going on here :)

kobiburnley:
@FredyC looks like there's a PR to revert the Symbol property support in Mobx4, I think it covers this issue.

https://github.com/mobxjs/mobx/pull/1988

FredyC:
@kobiburnley The #1988 was merged and released. That makes this PR obsolete I assume?

kobiburnley:
@FredyC #1988 was merged to mobx4-master, this PR is to master.

PR #2006, opened 2019-06-16
Update api.md

lqzhgood:
Redundant symbol, It should be a typo


FredyC:
It is actually correct denoting that name is optional and can be skipped. See https://mobx.js.org/refguide/action.html

Issue #1888, opened 2019-01-31
Mobx testing


mishk0:
Since v5.9.0 computed properties became not configurable. That's probably why Sinon.js throws an error, while stubbing a getter: ```TypeError: Cannot redefine property```.

Are there any ways to stub/spy a getter, best practices?

mweststrate:
@mishk0 

```javascript
class X {

   @computed get nonStubable() {
     return this._stubablle()
   }

   /* private */ stubable() {
      // this one is stubable, but the indirection won't loose any of the semantic benefits of `nonStubable` 
   }
}
```

ItamarShDev:
@mishk0 Any updates on this issue?

mishk0:
@mweststrate thanks!

Issue #1867, opened 2019-01-13
Create better error for missing decorated getters


jefffriesen:
In the case of using the decorate syntax (in an environment where decorators aren't enabled), there is a really confusing error message.

For example, in the documentation on how to not use decorators, you have this example:
```
class Timer {
  start = Date.now();
  current = Date.now();

  get elapsedTime() {
    return this.current - this.start + "milliseconds";
  }

  tick() {
    this.current = Date.now();
  }
}
decorate(Timer, {
  start: observable,
  current: observable,
  elapsedTime: computed,
  tick: action
});
```

If you delete the `get elapsedTime(){` function, you get this error:
```
Uncaught TypeError: Cannot read property 'get' of undefined
    at Object.propertyCreator (mobx.module.js:596)
    at initializeInstance$$1 (mobx.module.js:373)
    at MobxStore.set [as activeFile] (mobx.module.js:358)
    at new MobxStore (MobxStore.js:33)
```
The problem is that the `elapsedTime: computed` property doesn't get deleted with the function, which is easy to do when they aren't co-located (without decorators). This keeps happening to me and I've memorized the error, but the first couple times where hard to track down. It seems like there is no intentional error checking - it just errors down the stack at instantiation of the class.  

I think it should error (give the dev a chance to clean up cruft) but give them more information why it's throwing an error.

Thank you

mweststrate:
Computed properties should not be deleted, as it only removes the descriptor, but the computed value is still there, potentially leaking memory. (I think as a rule of thumb `delete` should never used in JavaScript, but that is a different point). 

The simple prevention for this is to mark the properties as non-configurable. Downside of that is that stubbing will be harder, but it might fix some other errors.

mweststrate:
Fixed and released in 4.9.0 / 5.9.0

mweststrate:
Oh sorry! I understood it as programmatically deleting :) Reopeneing.

Op di 22 jan. 2019 om 16:06 schreef Jeff Friesen <notifications@github.com>:

mweststrate:
Yes, that is precisely why we keep configurable true in general, but in this specific case it can also lead to quite confusing bugs. Not that both spy and inheritance issues can be worked around by extracting the logic like: 

```
class X {
  @computed getMyFummies() { return this._getMyFummies() } 
  
_getMyFummies() { 
   /* actual logic here */ 
  }
```

That makes it both possible to inherit `_getMyFummies` and spying it.

[...]

PR #2011, opened 2019-06-18
Allow overwriting computed values if configured

jillesme:
We ran into a problem where we couldn't overwrite computed properties in tests. Some of our computed properties are very complicated and it would be easier to just overwrite them in tests rather than overwrite the depending properties. 

This fixes https://github.com/mobxjs/mobx/issues/1888 and https://github.com/mobxjs/mobx/issues/1867)

PR FOR MOBX 5

* [x] Added unit tests
* [x] Updated changelog https://github.com/mobxjs/mobx/pull/2013
* [x] Updated docs https://github.com/mobxjs/mobx/pull/2012 

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "4.12.0",
+  "version": "4.11.0",
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
   "umd:main": "lib/mobx.umd.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "4.11.0",
+  "version": "4.12.0",
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "4.11.0",
+  "version": "4.12.0",
```
FredyC: Please do not change the version in PR, that will be done in release.

jillesme:
@FredyC thank you for the quick response. I've updated the package.json and will create a separate PR on the gh-pages branch to update the docs.

FredyC:
I wonder, you have linked #1888 which is clearly about MobX 5 and yet you are doing this change in V4 only? Can you please also make another PR with merge to `master`? Normally it should be the other way around.

jillesme:
You're right @FredyC - updating all related PRs.

jillesme:
Last 4 PRs are mine... @FredyC - Is there anything else that needs to be done on my side?

PR #2012, opened 2019-06-18
Update docs for #2011

jillesme:
https://github.com/mobxjs/mobx/pull/2011

PR #2014, opened 2019-06-18
Allow overwriting computed values if configured (MobX 5)

jillesme:
We ran into a problem where we couldn't overwrite computed properties in tests. Some of our computed properties are very complicated and it would be easier to just overwrite them in tests rather than overwrite the depending properties. 

This fixes https://github.com/mobxjs/mobx/issues/1888 and https://github.com/mobxjs/mobx/issues/1867

PR FOR MOBX 4 https://github.com/mobxjs/mobx/pull/2011

* [x] Added unit tests
* [x] Updated changelog https://github.com/mobxjs/mobx/pull/2013
* [x] Updated docs https://github.com/mobxjs/mobx/pull/2012 

### DIFF
--- a/src/api/configure.ts
+++ b/src/api/configure.ts
@@ -9,6 +9,7 @@ import {
 export function configure(options: {
     enforceActions?: boolean | "strict" | "never" | "always" | "observed"
     computedRequiresReaction?: boolean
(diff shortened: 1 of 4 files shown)


mweststrate:
LGTM!

Issue #2015, opened 2019-06-19
Setup a stale bot


FredyC:
It's not like we would be having hundreds of issues here, but still, might be worth to be auto closing stale issues.

@xaviergonz Can you please do an initial setup?

ItamarShDev:
this? https://probot.github.io/apps/stale/

FredyC:
@ItamarShDev Yea, that one :)

ItamarShDev:
Added to mobx repo

FredyC:
Great, I am not sure if it needs some other configuration. I would be expecting it take care of old issues and mark them as stale. I only saw it in action, never configured that one.

ItamarShDev:
we do, submitting PR

PR #2016, opened 2019-06-19
Add stale bot config

ItamarShDev:
implementation for issue #2015

### DIFF
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -7,12 +7,16 @@ exemptLabels:
   - bug
   - has PR
 # Label to use when marking an issue as stale
-staleLabel: wontfix
+staleLabel: stale
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
   recent activity. It will be closed if no further activity occurs. Thank you
   for your contributions.
 # Comment to post when closing a stale issue. Set to `false` to disable
-closeComment: >
-  Closing issue due to inactivity
+closeComment: false
+# Limit to only `issues` or `pulls`
+only: issues
+# Comment to post when removing the stale label.
+unmarkComment: >
+    This issue has been automatically unmarked as stale. Please disregard previous warnings.
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -4,9 +4,6 @@ daysUntilStale: 60
 daysUntilClose: 10
 # Issues with these labels will never be considered stale
 exemptLabels:
-  - pinned
-  - security
-  - discuss
   - bug
   - has PR
 # Label to use when marking an issue as stale
--- a/.github/stale.yaml
+++ b/.github/stale.yaml
@@ -1,11 +1,21 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 60
+daysUntilStale: 14
 # Number of days of inactivity before a stale issue is closed
-daysUntilClose: 10
+daysUntilClose: 4
 # Issues with these labels will never be considered stale
 exemptLabels:
   - bug
   - has PR
+  - discuss
+  - enhancement
+  - candidate-for-next-version
+  - ready-to-be-implemented-by-volunteer
+  - needs investigation
+  - breaking-change
+  - bug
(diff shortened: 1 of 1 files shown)


ItamarShDev:
@xaviergonz @FredyC @mweststrate Lets define the best configs for the bot and merge it!

.github/stale.yaml
```diff
@@ -0,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 10
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+  - discuss
```
FredyC: We don't have these three labels. Do we need them? Perhaps the `discuss` one, otherwise I would remove these from here.

FredyC:
It LGTM like that, we can tweak it over time if necessary.

The React repo has some additional feature of auto locking stale issues after bot closes them. It's useful so people don't tend to spam into dead issues but create a new one with a link to old one if relevant. Strangely enough, I don't see such option [in the configuration](https://github.com/probot/stale#usage).

.github/stale.yaml
ItamarShDev: will remove

ItamarShDev:
@FredyC Can I merge to master?

ItamarShDev:
@xaviergonz i have set the days to a higher value the in MST.
seems reasonable to set it to the same values as in MST

your thoghts?

xaviergonz:
Up to you, I set those because I thought 2 weeks gives people plenty of time to interact if they are serious into getting something fixed.

I'd also revamp the labels and make sure the ones that don't have to be stale are on the exemption list, since once stale is enabled it will start closing old issues as crazy

FredyC:
> I'd add only: issues so it doesn't auto close prs

What's wrong about auto closing PRs? I mean if it's valuable PR, we can pin, but if it's stale because the author decided not to continue on it, there is no reason to keep it open, right? Perhaps for that purpose, we should have a `pinned` label after all or something similar. 

> I'd also revamp the labels and make sure the ones that don't have to be stale are on the exemption list, since once stale is enabled it will start closing old issues as crazy

I would keep these labels as exempt

```
breaking-change
bug
candidate-for-next-version
discuss
duplicate
enhancement
has PR
help wanted
needs investigation
pinned
ready-to-be-implemented-by-volunteer
waiting-for-standardized-decorators
```

xaviergonz:
Sure, but make sure every open issue and pr has a label before the bot is enabled or it will be too late (talking from experience 😇)

ItamarShDev:
@FredyC @xaviergonz updated according to the middle ground (imo)
didn't add PRs, maybe we need to have a test run without it, and add PRs if we see fits.

FredyC:
I wouldn't worry about PRs much really, we have only 3 (besides this one) and I've pinned one. The other two are probably stale candidate anyway.

Nonetheless, even a closed PR can be reopened, so it's not a big deal.

ItamarShDev:
@FredyC 
good.
awaiting merging permission :)

.github/stale.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - bug
+  - has PR
+  - discuss
+  - enhancment
+  - candidate-for-next-version
+  - ready-to-be-implemented-by-volunteer
+  - needs investigation
+  - breaking-change
+  - bug
+  - duplicate
```
FredyC: Ah, this one is probably nonsense, I missed it :)

.github/stale.yaml
```diff
@@ -0,0 +1,33 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 14
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 4
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - bug
+  - has PR
+  - discuss
+  - enhancment
```
FredyC:
```suggestion
  - enhancement
```

FredyC:
I wonder when the magic starts happening, would be expecting something by now.

PR #2017, opened 2019-06-19
rename faulty file extension name in  #2016

ItamarShDev:
rename faulty file extension name in  #2016
I gave it the `yaml` extenstion instead of `yml`
@FredyC


ItamarShDev:
I am so sorry, I don't know how I didn't even noticed that silly mistake

FredyC:
That's fine, happens to best of us :) I did not notice that either.

PR #2018, opened 2019-06-21
minor typings improvement

ssured:

PR checklist:

* [ ] Added unit tests
* [ ] Updated changelog


FredyC:
What's the use cases for that? Can you show some code, please?

ssured:
`entries` is an API method which provides an iterator over the key-value combinations of an object.

This PR allows for:

```
const data = observable<{[key: string]: number}>({
a: 1,
b: 2
});

for (const [prop, value] of entries(data)) {
  // prop is of type string
  // value is of type number. Without the fix value would be of type any
}
```

FredyC:
Alright, thanks for explanation and contribution.

ssured:
Nice, thanks!

PR #2032, opened 2019-07-04
Use onBecomeUnobserved for ObservableMap._hasMap garbage collection

### LINKED ISSUE #2031: Memory leaking of Observable Maps (), because _hasMap keeps links on all once used keys (key could be object) and never delete them
Issue:
I find that observable Maps absolutely always loose a memory. For every key of the map, you create same key in your private map called _hasMap and place there a boolean value. But these keys are kept there even the origin map is empty or already doesn't contain the key. Don't forget that key could be large object or linked with many other objects. So we loose all memory even we don't keep links (have variables) to these key object in our code, because the _hasMap keeps links on all once used keys (key - is the big object, remember ?) and never delete them.

Please see the example:
https://codesandbox.io/s/minimal-mobx-react-project-1uyb0

Of cause we expected that we don't see the our objects in memory heap, after clearing the map.

Idea:
You should use another mechanism to register of using the map methods and of getting values.

mayorovp:
fixes #2031

* [x] Added unit tests
* [ ] Updated changelog
* ~Updated docs (either in the description of this PR as markdown, or as separate PR on the `gh-pages` branch. Please refer to this PR). For new functionality, at least [API.md](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md) should be updated~
* ~Added typescript typings~

### DIFF
--- a/test/base/map.js
+++ b/test/base/map.js
@@ -227,7 +227,28 @@ test("cleanup", function() {
     disposer()
     expect(aValue).toBe(2)
     expect(observable.observers.size).toBe(0)
-    expect(x._hasMap.get("a").observers.size).toBe(0)
+    expect(x._hasMap.has("a")).toBe(false)
+})
+
+test("getAtom encapsulation leak test", function() {
+    const x = map({})
+
+    let disposer = autorun(function() {
+        x.has("a")
+    })
+
+    let atom = mobx.getAtom(x, "a")
+
+    disposer()
+
+    expect(x._hasMap.get("a")).toBe(undefined)
+
+    disposer = autorun(function() {
+        x.has("a")
+        atom && atom.reportObserved()
+    })
+
+    expect(x._hasMap.get("a")).not.toBe(atom)
 })
 
 test("strict", function() {
--- a/src/types/observablemap.ts
+++ b/src/types/observablemap.ts
@@ -31,6 +31,7 @@ import {
     stringifyKey,
     transaction,
     untracked,
+    onBecomeUnobserved,
     globalState
 } from "../internal"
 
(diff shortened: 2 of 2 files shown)


FredyC:
@mayorovp Please fix tests if you can. Also what about adding some additional tests to ensure we won't break in the future?

And we will need to port this to `mobx4-master` eventually, but looks like it shouldn't be such a problem.

mayorovp:
@FredyC I don't know why `getAtom()` return `has` atom when no `data` atom is present, so I don't know how fix this error.

mweststrate:
I'll try to dive into it early next week. It might also be solvable without the atoms, if `onBecomeOpbserved(this.hasMap, key, callback)` is used

test/base/map.js
```diff
@@ -227,7 +227,28 @@ test("cleanup", function() {
     disposer()
     expect(aValue).toBe(2)
     expect(observable.observers.size).toBe(0)
-    expect(x._hasMap.get("a").observers.size).toBe(0)
+    expect(x._hasMap.get("a")).toBe(undefined)
+})
+
+test("getAtom encapsulation leak test", function() {
```
BanderasPRO: Please add a test, to check existence of a key in the _hasMap, not only empty value of the key

mayorovp: `Map<K, ObservableValue<boolean>>` cannot contain empty values.

BanderasPRO: I'm not sure that this condition of definition works in runtime. 
Native JS environment allows that.
you can just check the _hasMap size

mayorovp: Of course it works except when some typehacks broke all type safety.

But I already changed this line.

mweststrate:
Looking good to me as well, merging!

mweststrate:
Released as 4.13.0 / 5.13.0

PR #2033, opened 2019-07-05
Fixing stuck reactivity, increasing resilience against stale state

pavelserbajlo:
We were experiencing non-deterministic issues with mobx causing the reactivity to go stale, without any errors, warning or any other information that could lead to proper debugging of the problem. While inspecting the globalState object, It was obvious all actions and unobservations were cumulated as pending, but no more reactions were scheduled after the issue occured. More information on the problem here: https://github.com/mobxjs/mobx-react/issues/714

This PR aims to fix this issue and also focuses on increasing resilience to similar issues as the existing code can't reliably ensure that for every `startBatch()` call there's a corresponding `endBatch()` call.

My understanding of MobX internals is limited and I can't be sure if this fix introduces any potential issue, but I verified that:

- Our rather complex app works as expected
- I'm not seeing any duplicated or abundant reactions
- The problem with the stuck/stale reactivity has been fixed

I also tested performance, where the results are rather interesting, each before/after were concluded 5 times on my Macbook Pro:
Before (median): ~18.5 sec.
After (median): ~17.5 sec.

Not sure if this indicates that batches are now run more often with smaller amount of actions due to the updated condition or just false positive caused by my system.

Looking forward to discuss this change.

### DIFF
--- a/src/core/observable.ts
+++ b/src/core/observable.ts
@@ -106,7 +106,11 @@ export function startBatch() {
 }
 
 export function endBatch() {
-    if (--globalState.inBatch === 0) {
+    if (
+        --globalState.inBatch === 0 ||
+        globalState.pendingReactions.length > 0 ||
+        globalState.pendingUnobservations.length > 0
+    ) {
         runReactions()
         // the batch is actually about to finish, all unobserving should happen here.
         const list = globalState.pendingUnobservations
@@ -240,7 +244,7 @@ function logTraceInfo(derivation: IDerivation, observable: IObservable) {
 
         // prettier-ignore
         new Function(
-`debugger;
+            `debugger;
 /*
 Tracing '${derivation.name}'
 
--- a/src/api/object-api.ts
+++ b/src/api/object-api.ts
@@ -65,7 +65,9 @@ export function values(obj: any): string[] {
 export function entries<K, T>(map: ObservableMap<K, T>): ReadonlyArray<[K, T]>
 export function entries<T>(set: ObservableSet<T>): ReadonlyArray<[T, T]>
 export function entries<T>(ar: IObservableArray<T>): ReadonlyArray<[number, T]>
-export function entries<T = any>(obj: T): ReadonlyArray<[string, T extends object ? T[keyof T] : any]>
+export function entries<T = any>(
+    obj: T
+): ReadonlyArray<[string, T extends object ? T[keyof T] : any]>
 export function entries(obj: any): any {
     if (isObservableObject(obj)) {
         return keys(obj).map(key => [key, obj[key]])
@@ -94,8 +96,8 @@ export function set<T extends Object>(obj: T, key: PropertyKey, value: any)
 export function set(obj: any, key: any, value?: any): void {
     if (arguments.length === 2 && !isObservableSet(obj)) {
         startBatch()
-        const values = key
         try {
+            const values = key
             for (let key in values) set(obj, key, values[key])
         } finally {
             endBatch()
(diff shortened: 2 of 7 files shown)


mweststrate:
It would be interesting to see if any of the specific changes also solve the issue. Did you check doesn't have unexpected (but probably caught) exceptions somewhere? (The Chrome devtools can help you discovering those)

pavelserbajlo:
I’m certain no exception is thrown (either handled or unhandled)

mweststrate:
But then finally blocks shouldn't make a difference. I'm puzzled 🤔.

Op vr 5 jul. 2019 16:31 schreef pavelserbajlo <notifications@github.com>:

pavelserbajlo:
I added those as the code was inconsistent in this regard. What makes the difference (in terms of the core issue) is the condition in endBatch()

src/core/observable.ts
```diff
@@ -106,7 +106,11 @@ export function startBatch() {
 }
 
 export function endBatch() {
-    if (--globalState.inBatch === 0) {
+    if (
+        --globalState.inBatch === 0 ||
+        globalState.pendingReactions.length > 0 ||
```
mayorovp: This breaks all batch logic! In fact, batches become useless after this change.

pavelserbajlo: yes, that's the outcome to my understanding, too. Considering performance test results and tests in real-life app that easily generates ~50 actions a second, how are the batches still useful? For us, this is currently causing more pain than any measurable difference known to me. What are some other benefits of batches I'm not seeing?

mweststrate:
Please preserve batching semantics as they are. We can't change the
semantics for thousands if users to fix it for one. I  think we need an
reproducible setup before making such fundamental changes. I suspect the
origin of this problem is in the original app, somehow using mobx not
correctly. This mechanism barely hasn't changed in the last 3 years, and
hasn't caused similar problems before afaik, so I think a more thorough
understanding of the problem is required before diving into solutions.

Op do 11 jul. 2019 10:39 schreef pavelserbajlo <notifications@github.com>:

pavelserbajlo:
That is perfectly fair point. Unfortunately, we were unable to identify the possible core problem in our app and we spent weeks trying to. Following all the recommendations and documentation, mobx is not giving us any clue on where the possible issue could be. I'd seek for help in the original issue linked above several times and didn't get much help. If there really is a problem in our app, mobx should be able to help me identify it in some way. Especially if some sort of app error leads to complete mobx break-down without any further errors or warnings.

I'd be more than happy to identify the core issue we're facing, but at this point I don't know what/how to debug what is causing the batching semantics to malfunction.

src/core/observable.ts
mayorovp: Simple ObservableMap,set call (and any observable object property assignment!) can generate 3 different changes in one batch. You cannot maintain model consistency without batches.

pavelserbajlo: Makes sense. Happy to take this to the original issue if you're aware of any ways to perform proper debugging in order to find the core issue. Otherwise we have to run mobx from our fork without the proper model consistency.

mweststrate:
Did you try spy / the browser devtools to investigate after which
(re)action mobx stops working?

Op do 11 jul. 2019 12:30 schreef pavelserbajlo <notifications@github.com>:

mweststrate:
Closing this PR. Although it saddens me that MobX apparently causes some unexplainable trouble which did cost you a lot of time, there seems to be no consistent reproduction, nor other people running into the same problem. Since we can't accept a shot-in-the-dark solution I recommend to maintain your own fork or use patch-package.

PR #2039, opened 2019-07-10
Update MobX Globals

OverseePublic:
PR checklist:

* [x] Added unit tests
* [ ] Updated changelog


FredyC:
Wait, I am confused. So you have added `enableTypeCheck` which isn't used anywhere in the mobx code. Then you made PR https://github.com/mobxjs/mobx-state-tree/pull/1332 which only expands on using env variable and not this option. That doesn't make any sense :)

OverseePublic:
@FredyC My commits are missing :) I hope I did it correctly. Could you take a look at the updated PR in  mobxjs/mobx-state-tree#1332 ?
Thanks!

FredyC:
Ok, but still I think it's as a strange solution. You add a configuration option to MobX which does not use it in any way and it's only utilized by MST. Why?

I think it would make much more sense to keep this contained within MST, perhaps even that env variable you had there before or some sort of runtime config.

Additionally, it would require for MST to depend on specific (to be released) version of MobX. How is that a good solution? What about people that need to use MobX 4?

Sorry, I don't we want to merge this in here unless you come up with some explanation why is it needed.

OverseePublic:
No problem! I might have done it incorrectly. I did not see any configuration object in MST, so I followed @mweststrate 's comment in the MST PR and resulted in this.

Indeed, I agree it is bizarre. I think it is bizarre because I am noob at Mobx and MST and I probably did it wrong. You are correct in terms of controlling MST behavior with MobX, and I think ENV might be a better solution. Michael, do you have any inputs on this discussion?

Thanks.

Issue #2025, opened 2019-06-29
uncheck dirty mark on a filled which has been changed multitimes and at last it's value wasnot changed


MHA15:
3. [x] Idea: check observable changes before run trackable functions to be sure if it was really changed  
  * [ ] What problem would it solve for you? less component renders 

  * [ ] Do you think others will benefit from this change as well and it should in core package (see also mobx-utils)? no it's a core feature
  * [ ] Are you willing to (attempt) a PR yourself? maybe not so sure (I didn't read the source code yet)

I was checking some cases in mobx observable and I figure out that if this action gets run it would not call trackable functions (which is ok and good for app performance)
```js
// store = {x: 10}
runInAction(() => {
store.x = 10;  // not checked as dirty
})
```

but in a case like this it will call trackable functions but data is still not changed and it is a redundant call
```js
// store = {x: 10}
runInAction(() => {
store.x = 5; // checked as dirty
store.x = 10; // should remove dirty mark
})
```

I think that when the most outer action will finish it should not call functions which was listening on changes which have been undo in actions

FredyC:
Can you provide a small runnable example to show the problem in action?

urugator:
Note that you can workaround it like this: 
```javascript
class Store {
  @observable _x = 10;
  @computed get x() {
    return this._x;
  }
}

const store = new Store();

reaction(
  () => store.x,
  x => console.log(x)
);

runInAction(() => {
  store._x = 5; 
  store._x = 10;
});
```
Which raises the question - can't we simply wrap observable accessors into computed?

aretecode:
Wouldn't this add additional complexity in the codebase, and _some_ overhead by doing an additional `isEqual` for _every_ `action`? 
Just to satisfy the edge case where the `property` is changed to a `new value`, and then back to the `original value` inside of the **same** `action`?

If you're changing the `property` to a `new value`, and then back to the `original value`, why not use a temporary variable in your application code instead?

urugator:
> isEqual for every action

Considering the "each observable as computed idea", we could remove the equality check preformed during *every* modification, because the equality check would be performed by computed lazily later during *get* against the cached value - and if there is no reaction, then there is no cache and therefore no equality check...
So in case you're updating single value multiple times, there would be less checks, no checks if value isn't observed and the same number of checks in other cases.
However these checks are really trivial and quite unimportant from the performance perspective.
[...]

PR #2049, opened 2019-07-19
[PoC, do not merge] observables will use same update protocol as computeds

mayorovp:
Experimental implementation of #2025 for perf tests

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -8,6 +8,8 @@ import {
     onBecomeObserved,
     onBecomeUnobserved,
     propagateChanged,
+    propagateMaybeChanged,
+    propagateChangeConfirmed,
     reportObserved,
     startBatch
 } from "../internal"
@@ -27,7 +29,7 @@ export class Atom implements IAtom {
 
     diffValue = 0
     lastAccessedBy = 0
-    lowestObserverState = IDerivationState.NOT_TRACKING
+    lowestObserverState = IDerivationState.UP_TO_DATE
     /**
      * Create a new atom. For debugging purposes it is recommended to give it a name.
      * The onBecomeObserved and onBecomeUnobserved callbacks can be used for resource management.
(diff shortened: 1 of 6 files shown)


mayorovp:
I closed this PR to avoid accidental merge. Please do not delete the branch.

FredyC:
Just for a future reference, there is a new feature when you are creating PR, you can select draft PR instead which cannot be merged by anyone until you switch it to be ready.

FredyC:
@mayorovp Are you still going to pursue this? I went on a cleaning spree of branches, so I wonder if we need to keep this one around. In the case you want to keep it, it would be better to reopen or it might die with next spree :)

mayorovp:
@FredyC #2025 is still open

FredyC:
That doesn't answer the question if you want to pursue this path :) Reopening just to keep in on the radar.

FredyC:
So I guess after 3 months it's safe to assume this is an abandoned idea? Feel free to reopen if you still want to work on that.

mayorovp:
@FredyC #2025 is still open

FredyC:
@mayorovp I know, but I am not really sure what this is all about. Up to you if you want to continue with it then reopen.

urugator:
Personally I find this likekable, not because of the potential optimization benefit (#2025), but because it seems to reduce the overall complexity a little bit (one protocol instead of two).
If that's truly the case - so there is a space for some code simplification/reduction, I think it's worth finalizing and merging.
We shouldn't be able to rollback the change in case it turns out to be problematic for whatever reason, right?

mayorovp:
@FredyC this is a just one commit which is attached to issue #2025. There are no reason to delete it while #2025 is open.
@urugator this change can have negative impact on perfomance. Or on memory usage.,

urugator:
@mayorovp Yea, but do we know how significant unless we try it in wild? Do we have a perf test with somehow meaningful output?

mayorovp:
i don't know

>

Issue #2048, opened 2019-07-18
Flow-Bin 0.103.0 Error


wadethestealth:
When upgrading to flow-bin 0.103.0, I get this flow error.

```
Error ------------------------------------------------------------------------- node_modules/mobx/lib/mobx.js.flow:118:8

Cannot extend `Array` [1] with `IObservableArray` because:
 - boolean [2] is incompatible with mixed [3] in the return value of the first argument of property `find`.
 - boolean [4] is incompatible with mixed [5] in the return value of the first argument of property `findIndex`.

   node_modules/mobx/lib/mobx.js.flow:118:8
               v-----------------------------------------------
   118| export interface IObservableArray<T> extends Array<T> {
   119|     spliceWithArray(index: number, deleteCount?: number, newItems?: T[]): T[];
   120|     observe(
   121|         listener: (changeData: IArrayChange<T> | IArraySplice<T>) => void,
   122|         fireImmediately?: boolean
   123|     ): Lambda;
   124|     intercept(handler: IInterceptor<IArrayWillChange<T> | IArrayWillSplice<T>>): Lambda;
   125|     intercept(handler: IInterceptor<IArrayChange<T> | IArraySplice<T>>): Lambda;
   126|     intercept<T>(handler: IInterceptor<IArrayChange<T> | IArraySplice<T>>): Lambda;
   127|     clear(): T[];
   128|     replace(newItems: T[]): T[];
   129|     find(
   130|         predicate: (item: T, index: number, array: Array<T>) => boolean,
   131|         thisArg?: any,
   132|         fromIndex?: number
   133|     ): T | any;
   134|     findIndex(
   135|         predicate: (item: T, index: number, array: Array<T>) => boolean,
   136|         thisArg?: any,
   137|         fromIndex?: number
   138|     ): number;
   139|     remove(value: T): boolean;
   140| }
        ^

References:
   node_modules/mobx/lib/mobx.js.flow:118:46
   118| export interface IObservableArray<T> extends Array<T> {
                                                     ^^^^^^^^ [1]
   node_modules/mobx/lib/mobx.js.flow:130:65
   130|         predicate: (item: T, index: number, array: Array<T>) => boolean,
                                                                        ^^^^^^^ [2]
   ../../../AppData/Local/Temp/flow/flowlib_379ba966/core.js:260:68
   260|     find(callbackfn: (value: T, index: number, array: Array<T>) => mixed, thisArg?: any): T | void;
[...]

PR #2050, opened 2019-07-20
Fix flow definition for Flow 103

SebastienGllmt:
Flow 103 throws an error on the current mobx types (see #2048). This is a simple fix (we just mirror the change Flow made).

I'd like to get this released sooner than later because this issue is a blocker for every Mobx user that also has Flow as a dependency 🙏 

PR checklist:

* [ ] Updated changelog

### DIFF
--- a/flow-typed/mobx.js
+++ b/flow-typed/mobx.js
@@ -127,12 +127,12 @@ export interface IObservableArray<T> extends Array<T> {
     clear(): T[];
     replace(newItems: T[]): T[];
     find(
-        predicate: (item: T, index: number, array: Array<T>) => boolean,
+        predicate: (item: T, index: number, array: Array<T>) => mixed,
         thisArg?: any,
         fromIndex?: number
     ): T | any;
     findIndex(
-        predicate: (item: T, index: number, array: Array<T>) => boolean,
+        predicate: (item: T, index: number, array: Array<T>) => mixed,
         thisArg?: any,
         fromIndex?: number
     ): number;


flow-typed/mobx.js
```diff
@@ -127,12 +127,12 @@ export interface IObservableArray<T> extends Array<T> {
     clear(): T[];
     replace(newItems: T[]): T[];
     find(
-        predicate: (item: T, index: number, array: Array<T>) => boolean,
+        predicate: (item: T, index: number, array: Array<T>) => mixed,
```
FredyC: I don't know flow that much, but shouldn't it return T instead of mixed?

flow-typed/mobx.js
```diff
@@ -127,12 +127,12 @@ export interface IObservableArray<T> extends Array<T> {
...
         thisArg?: any,
         fromIndex?: number
     ): T | any;
     findIndex(
-        predicate: (item: T, index: number, array: Array<T>) => boolean,
+        predicate: (item: T, index: number, array: Array<T>) => mixed,
```
FredyC: The index is always numeric, right? Why not return number then?

flow-typed/mobx.js
SebastienGllmt: Not according to https://github.com/facebook/flow/commit/00be88084a2ef41fadce99352b15f683f3c67f36

flow-typed/mobx.js
SebastienGllmt: Again, `mixed` here is the return type of the `predicate`, not of `findIndex`. It's `mixed` and not `boolean` because anything that can be coerced to a boolean is okay as a return type.

FredyC:
Thanks for comments and contribution, make sense.

danrot:
Any chance we can have that backported for mobx 4? :slightly_smiling_face:

danrot:
My bad, just saw #2051, but seems as if this was not released yet... Would it be possible to have a release soonish? :slightly_smiling_face:

SebastienGllmt:
Yeah I'd really appreciate if this was released sooner than later also since we're also stuck waiting for this release to upgrade our version of Flow

PR #2054, opened 2019-07-21
Setup the lock bot

FredyC:
Adding a stale bot (#2015) was surely a good step. Let's take it further and add the [lock bot](https://github.com/apps/lock) as well. People tend to comment on long-closed issues. It's a bad habit as nobody else except those who participated (or watching a whole repo) can't see such a comment.

I've used pretty much default config except I've set 60 days for locking closed issues. I've excluded pulls.

All we need now is to enable the app which I don't have permission to do.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,26 @@
+# Configuration for Lock Threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 60
+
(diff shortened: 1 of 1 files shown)


xaviergonz:
Shouldn't the list of exemptions be the same than the ones from stale bot?

FredyC:
Lock bot works over closed issues only. So once the Stale bot closes issue (based on labels) it's safe to assume we want to lock it as well eventually :)

xaviergonz:
Ahh good to know. Bot installed

PR #2055, opened 2019-07-24
Fixing import in the example

devlato:
Just a small update for the docs. On the page about `@computed`, they import `computed` from `mobx` but in the code below, the `computedFn` from `mobx-utils` is used instead. So I'm just fixing the imports.

PR checklist:

* [ ] Added unit tests
* [ ] Updated changelog


devlato:
@mweststrate

FredyC:
No need to ping mweststrate for that, there are other maintainers.

Thanks for the contribution, looks good to me.

Issue #1980, opened 2019-05-29
ObservableMap.replace() - breaks entities order!


sergii-aleksieiev-ring:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Specifications

> The Map object holds key-value pairs and remembers the original insertion order of the keys.

The Map should guarantee the order of entities, but when I use method ObservableMap.replace() I got Map with broken order of entities.

The doc (https://mobx.js.org/refguide/map.html) says:

> replace(values). Replaces the entire contents of this map with the provided values. Short hand for .clear().merge(values)

So as I understand it should physically replace old Map by a new one, but in fact, it is trying to do a "smart" merge inside:

```
replace(values: ObservableMap<K, V> | IKeyValueMap<V> | any): ObservableMap<K, V> {
    transaction(() => {
        // grab all the keys that are present in the new map but not present in the current map
        // and delete them from the map, then merge the new map
        // this will cause reactions only on changed values
        const newKeys = (getMapLikeKeys(values) as any) as K[]
        const oldKeys = Array.from(this.keys())
        const missingKeys = oldKeys.filter(k => newKeys.indexOf(k) === -1)
        missingKeys.forEach(k => this.delete(k))
        this.merge(values)
    })
    return this
}
```

As a workaround I can use the following:

```
myMap.clear();
myMap.merge(newMap);
```

but it calls observe twice during one operation.

ItamarShDev:
Hi!
can you please create a CodeSandbox with a reproduction of this?
This will greatly help us to see and understand the case. 

thank you

urugator:
Reproduction: https://codesandbox.io/s/mobx-mapreplace-breaks-entities-order-cebhk

FredyC:
Since the `replace` is the addition of MobX, it doesn't necessarily need to adhere to official Map behavior imo.

Besides that "smart merge" is probably much more important to minimalize reactions for things that haven't changed.

Ultimately, I think we should just add a warning to docs with the workaround. Or perhaps introduce an additional method that will do what that workaround does? It's not very often you need to rely on the order in the Map in my opinion.

urugator:
These `replace` methods are mainly intended as an alternative to reassigning new array/map. 
The problem with re-assigment is that it will completely unnecessarily invalidate everything which depends on the `array/map` reference alone (without touching it's content).
So imo the behavior shouldn't differ from ordinary assigment.
[...]

Issue #2003, opened 2019-06-13
Trivial performance bug in ObservableMap.replace


Maaartinus:
I have a:

2. [x] Issue:
  * [x] Did you check this issue wasn't filed before?

No behavior change; it's just performance bug.

  * [x] State the versions of MobX and relevant libraries. Which browser / node / ... version?

https://github.com/mobxjs/mobx/blob/6c1d807a44dfc32567f3e55231eb1464a2155826/src/types/observablemap.ts#L334

I'm proposing the following trivial change:

-            const newKeys = (getMapLikeKeys(values) as any) as K[]
            const oldKeys = Array.from(this.keys())
-            const missingKeys = oldKeys.filter(k => newKeys.indexOf(k) === -1)
+           const missingKeys = oldKeys.filter(k => !values.has(k))

The current behavior has quadratic complexity, with two nested loops (over all old and new keys). So for a map with a few thousands elements it may take ages, while `values.has` is supposed to be a constant-time operation (leading to linear overall complexity).

urugator:
Note that `values` is not neccessarily a `Map`, it can be object `{ key: value }` or array `[[key,value]]`, so you may need to convert it first:
```javascript
if (isPlainObject(values)) {
  // could be done in single pass, but whatever
  values = Object.entries(values);
}
if (Array.isArray(values)) {
  values = new Map(values);
} 
```

However I think we should resolve #1980 first.

FredyC:
PR #2057 merged and waiting for the release

Bnaya:
published

PR #2057, opened 2019-07-25
fix issues 1980 and 2003

haimrait:
Fix issues #1980 and #2003.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 5.13.0 / 4.13.0
 
+* Fixed bug that observable map replace function breaks entities order. related issues:[#1980](https://github.com/mobxjs/mobx/issues/1980) and [#2003](https://github.com/mobxjs/mobx/issues/2003)
 * Fixed potential memory leak in observable maps, when non-primitive values are used as keys. Fixes [#2031](https://github.com/mobxjs/mobx/issues/2031) through [#2032](https://github.com/mobxjs/mobx/pull/2032).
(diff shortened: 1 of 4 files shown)


mweststrate:
Great PR, thanks!

FredyC:
I guess no need to hold it longer here since you got approval from the highest authority :)

Can you also prepare PR to `mobx4-master` branch, please?

urugator:
Thanks for the PR! I see it's based on the snippet I posted, but you removed the part which was supposed to notify observers when only the order is changed. A situation like this:
```javascript
const map = observable.map({ a: "a", b: "b" });
autorun(() => map.entries());
map.replace({ b: "b", a: "a" }); // should invoke autorun
```
Is it handled correctly? Do we have a test for that?

haimrait:
Yea... I checked this issue and it invokes autorun one time for your case.

haimrait:
@FredyC I made a pr for mobx4- master branch
https://github.com/mobxjs/mobx/pull/2058

urugator:
> Yea... I checked this issue and it invokes autorun one time for your case.

Interesting... I wonder what invalidates the reaction then ... `this.delete` isn't called and `this.set` shouldn't since the value isn't modified...

FredyC:
@mweststrate I see you are on issue spree but don't publish V5 right now. Chances are it's actually broken, see #2058

mweststrate:
:) I won't, but I'm not the only publisher in here, so probably best to
revert that merge, and then immediately open a PR with a revert of the
revert, linking to here

PR #2058, opened 2019-07-25
fix issues 1980 and 2003

haimrait:
Fix issues 1980 and 2003

### DIFF
--- a/test/base/map.js
+++ b/test/base/map.js
@@ -618,9 +618,20 @@ test("issue 1243, .replace should not trigger change on unchanged values", () =>
 
     expect(() => {
         m.replace("not-an-object")
-    }).toThrow(/Cannot get keys from 'not-an-object'/)
+    }).toThrow(/Cannot convert to map from 'not-an-object'/)
 
     d()
+})
+
+test("#1980 .replace should not breaks entities order!", () => {
+    const original = mobx.observable.map([["a", "first"], ["b", "second"]])
+    const replacement = new Map([["b", "first"], ["a", "second"]])
+    original.replace(replacement)
+    const newKeys = Array.from(replacement)
+    const originalKeys = Array.from(replacement)
+    for (let i = 0; i < newKeys.length; i++) {
+        expect(newKeys[i]).toEqual(originalKeys[i])
+    }
 })
 
 test("#1258 cannot replace maps anymore", () => {
--- a/test/base/map.js
+++ b/test/base/map.js
@@ -634,6 +634,18 @@ test("#1980 .replace should not breaks entities order!", () => {
     }
 })
 
+test("#1980 .replace should should invoke autorun", () => {
+    const original = mobx.observable.map({ a: "a", b: "b" })
+    const replacement = { b: "b", a: "a" }
+    let numOfInvokes = 0
+    autorun(() => {
+        numOfInvokes = numOfInvokes + 1
+        return original.entries()
+    })
+    original.replace(replacement)
+    expect(numOfInvokes).toBe(1)
+})
+
 test("#1258 cannot replace maps anymore", () => {
     const items = mobx.observable.map()
     items.replace(mobx.observable.map())
--- a/test/base/map.js
+++ b/test/base/map.js
@@ -651,6 +651,21 @@ test("#1980 .replace should should invoke autorun", () => {
     expect(numOfInvokes).toBe(2)
 })
 
+test("#1980 .replace should not report changed unnecessarily", () => {
+    const mapArray = [["swappedA", "swappedA"], ["swappedB", "swappedB"], ["removed", "removed"]]
+    const replacementArray = [mapArray[1], mapArray[0], ["added", "added"]]
+    const map = mobx.observable.map(mapArray)
+    let autorunInvocationCount = 0
+    autorun(() => {
+        map.get("swappedA")
+        map.get("swappedB")
+        autorunInvocationCount++
+    })
+    map.replace(replacementArray)
+    expect(Array.from(map.entries())).toEqual(replacementArray)
+    expect(autorunInvocationCount).toBe(1)
+})
+
 test("#1258 cannot replace maps anymore", () => {
     const items = mobx.observable.map()
     items.replace(mobx.observable.map())
--- a/test/base/map.js
+++ b/test/base/map.js
@@ -640,10 +640,15 @@ test("#1980 .replace should should invoke autorun", () => {
     let numOfInvokes = 0
     autorun(() => {
         numOfInvokes = numOfInvokes + 1
-        return original.entries()
+        return original.entries().next()
     })
     original.replace(replacement)
-    expect(numOfInvokes).toBe(1)
+    const orgKeys = Array.from(original.keys())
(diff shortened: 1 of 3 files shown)


FredyC:
Thanks! Can we perhaps add one more test here for the case @urugator [mentioned](https://github.com/mobxjs/mobx/pull/2057#issuecomment-515122535)? Just for sake of everyone :) Make it in a separate commit and I will cherry-pick it to master also.

test/base/map.js
```diff
@@ -634,6 +634,18 @@ test("#1980 .replace should not breaks entities order!", () => {
...
+test("#1980 .replace should should invoke autorun", () => {
+    const original = mobx.observable.map({ a: "a", b: "b" })
+    const replacement = { b: "b", a: "a" }
+    let numOfInvokes = 0
+    autorun(() => {
+        numOfInvokes = numOfInvokes + 1
+        return original.entries()
+    })
+    original.replace(replacement)
+    expect(numOfInvokes).toBe(1)
```
urugator: You have to expect 2 - autorun always runs at least once - during definition and later on every notification

haimrait:
@FredyC Added test for this case.
Don't forget add it to master

test/base/map.js
FredyC: So chances are it actually doesn't. Good thing we did add that test.

Btw, I would also add expect inside the autorun to check for correct order, just to be sure.

urugator: Alternatively you can use `reaction(() => map.entries(), () => numOfInvokes++)` and expect `1`

haimrait: I checked this issue, you're right but at the previous implementation, it failed too.
About the solution in the mobx new version, i can use **this._keysAtom.reportChanged()** but at mobx4 i dont have **k_eysAtom**.
Do you have any suggestions?

FredyC:
I am thinking about reverting the other merged V5 PR so it's not accidentally published and breaks things. At least until a solution is discovered. Or can you @haimrait prepare a PR with added test and fix?

test/base/map.js
urugator: I will try to take a look at v4 later this weekend or the next week. I think we have to modify `this._keys` or something ... not even sure why the impl is different (as there is no need for proxies?)

mweststrate: @haimrait awesome work so far!

In addition, we should probably test that there are _no_ changes in the keys, the keyset is not updated (e.g. `{ a: "a", b: "b"}` -> `{ a: "a", b: "b" }` should not result in new keys nor entries, and `{ a: "a", b: "b" }` -> ` { a: "A", b: "B" }` should result in new entries but no new keyset)

If we get this correct, getting it right in v4 shouldn't be too hard indeed. (I'd actually expect the implementation to be the same indeed :P)

haimrait: @mweststrate you mean that we should check and not constantly assign
this._data = orderedData.
Because in { a: "a", b: "b"} -> { a: "a", b: "b" } case we have no change.
and { a: "a", b: "b" } -> { a: "A", b: "B" } case we want to change only keys value and not create a new Map object.
Can you confirm my statements?

mweststrate: Hi @haimrait the constant reassignment is fine. I think the only thing needed to pass the test cases described above is to only trigger the `keysAtom.reportChanged` if there is actually a change in the keyset (including changes in order). So that should be conditional.

haimrait: but if we are doing the actual reassignment its change the this._data - Map entries and keys at the memory because it's a different map?

urugator: First of all I think that the V4 could be refactored into V5 (`this._keys` => `this._keysAtom`) (?)
For the time being, if I follow the V4 code correctly, the `this._data` can stay in whatever order, but we have to make sure that `this._keys` is ordered correctly:
```javascript
    transaction(() => {
      const replacementMap = ObservableMap.convertToMap(values);      
      const oldKeys = this._keys;      
      const newKeys = Array.from(replacementMap.keys());      
      let keysChanged = false;
      for (let i = 0; i < oldKeys.length; i++) {
        const oldKey = oldKeys[i];
        // order key change        
        if (oldKeys.length === newKeys.length && oldKey !== newKeys[i]) {
          keysChanged = true;
        }
        // deleted key
        if (!replacementMap.has(oldKey)) {          
          keysChanged = true;
          this.delete(oldKey);
        }
      }
      replacementMap.forEach((value, key) => {
        // new key
        if (!this._data.has(key)) {
          keysChanged = true;
        }
        this.set(key, value);                        
      })      
      if (keysChanged) {
        this._keys.replace(newKeys);
      }      
    })
    return this
}
```

mweststrate: @haimrait if you get stuck / confused somewhere, just let us know! I'll gladly help, and admittedly not following up very closely, as I'm currently in the middle of a move. But hope to have some more spare time again in 2 weeks or so!

haimrait: yea im sorry for the late i hope to commit changes today.

test/base/map.js
```diff
@@ -618,11 +618,34 @@ test("issue 1243, .replace should not trigger change on unchanged values", () =>
...
+    }
+})
+
+test("#1980 .replace should should invoke autorun", () => {
+    const original = mobx.observable.map({ a: "a", b: "b" })
+    const replacement = { b: "b", a: "a" }
+    let numOfInvokes = 0
+    autorun(() => {
+        numOfInvokes = numOfInvokes + 1
+        return original.entries()
```
urugator: I haven't realized that `original.entries` returns iterator, not the array directly, please change it to: `return original.size()` or `return original.entries().next()` that should do the trick.

haimrait: done :)

urugator:
Good job, if you could add one more test it would be excellent:
```javascript
test("#1980 .replace should not report changed unnecessarily", () => {
    const mapArray = [["swappedA", "swappedA"], ["swappedB", "swappedB"], ["removed", "removed"]];
    const replacementArray = [mapArray[1], mapArray[0], ["added", "added"]];   
    const map = observable.map(mapArray)        
    let autorunInvocationCount = 0;
    autorun(() => {
      map.get("swappedA");
      map.get("swappedB");
      autorunInvocationCount++;
    });    
    map.replace(replacementArray);
    expect(Array.from(map.entries())).toEqual(replacementArray);
    expect(autorunInvocationCount).toBe(1);
})
```

urugator:
@mweststrate When/if you will have a time, could you take a look whether there is a particular reason for using observable `keys` array rather than `keysAtom` in v4?

mweststrate:
From the top of my head there isn't, I think this just evolved over time
and the implications could be the same in this regard. If making it the
same still passes the tests, feel free to go ahead 😊

Op za 3 aug. 2019 11:38 schreef urugator <notifications@github.com>:

haimrait:
@urugator @FredyC can we merge this pr?

FredyC:
I did not give it a proper thought, but @urugator approved, so let's roll :) 

@haimrait Can you please make a second PR for MobX 5 as well?

haimrait:
sure

PR #2078, opened 2019-08-16
chore: Add & apply yarn deduplicate

Bnaya:
Motivation:
yarn is actually missing automatic dedup of deps,
Which leads to multiple-but-compatible versions of deps.
[yarn-deduplicate](https://github.com/atlassian/yarn-deduplicate) is a complementary tool to solve that. 

It's recommended to run it after any dependencies add/remove/upgrade

Related:
https://github.com/yarnpkg/yarn/issues/3778
(And there are more issues around yarn repo related to that)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "small-build": "node scripts/build.js",
     "lint": "eslint src/**/*.ts",
     "size": "size-limit",
-    "publish-script": "node scripts/publish.js"
+    "publish-script": "node scripts/publish.js",
+    "yarn-deduplicate": "yarn-deduplicate -s fewer yarn.lock"
   },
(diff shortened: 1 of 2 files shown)


mweststrate:
Thanks! Interesting tool, not sure we really needed, but if it works without prob, might make it slightly faster for people to contribute

Bnaya:
The yarn.lock got little bit smaller, 
means the node_modules was little bit bigger and nested than it could be.
in some repos deduping cleaning up maybe third of the lock file.

FredyC:
Wouldn't it be better to just `npx yarn-deduplicate`? Adding it to deps it's not exactly sure what it has done because some other deps got in there.

Issue #1197, opened 2017-10-10
suggestion: log warning when computed property is accessed outside reaction/observer


marty-wang:
One of my component got great performance downgrade after refactoring. It turns out that the computed property is accessed in a hot code path that is not in reaction/observer, therefore the computed property got recomputed over and over again. 

The is one of those errors that is really hard to track down. It would be nice that mobx would log warning in the debug mode when the computed property is used this way. After all, computed properties are primarily supposed to be used in reaction/observer, which enables its memorization capability. However, because of the fact that you can still use computed property in non reaction/observer context, it makes really hard to track down the performance problem caused by it.

Please consider it, or provide other way to warn developer when they use computed properties in wrong context.

urugator:
What about accessing computed inside "writing" context?
```javascript
@computed get sum() {
  return this.a + this.b
}

@action doSomething() {  // not necessarily action/tx/batch?
  this.a = 5;
  console.log(this.sum) // do something with sum
}  
```

urugator:
What about accessing computed inside "writing" context (not neccessarily action/tx/batch)?
```javascript
@computed get sum() {
  return this.a + this.b
}

@action doSomething() {
  this.a = 5;
  console.log(this.sum) // do something with sum
}  
```

marty-wang:
@urugator you can always extract the body of the computed property to a method, and use that method in other context. This way the expectation is clear, i.e. the computed is expected to be memorized and the extraction method is expected to re-compute every time, just a normal method is supposed to do.

urugator:
Being used in writing context doesn't mean it needs to recompute every time the context is invoked. It needs to be recomputed only if the dependencies were changed. That may happen conditionaly or not at all.
Also from the perspective of consumer, how do you recognize that something is computed, the semantic is indistinguishable from non-computed prop, which can be used anywhere...

marty-wang:
@urugator 

Before we discuss further, I hope we can agree that a good API should have the following properties

1. clear in semantic, meaning that it does exactly what it says it does, no more no less.
2. consistent in behavior, not contextual, meaning that it does not work one way in some context and work another in other context
3. makes doing right thing easier and makes doing wrong thing harder
4. communicate with developers with helpful messages when they do something wrong.

If we cannot agree on these, I don't think we can be productive in the thread and might as well stop here.

[...]

PR #2079, opened 2019-08-18
feat: derivationRequiresObservable & observablesRequiresReaction

Bnaya:
PR checklist:

* [x] Added unit tests
* [x] Updated changelog
* [x] Backport to mobx 4 branch (?)

### DIFF
--- a/src/api/configure.ts
+++ b/src/api/configure.ts
@@ -9,6 +9,16 @@ import {
 export function configure(options: {
     enforceActions?: boolean | "strict" | "never" | "always" | "observed"
     computedRequiresReaction?: boolean
+    /**
+     * (Experimental)
+     * Warn if you try to create to derivation / reactive context without accessing any observable.
+     */
+    reactionRequiresObservable?: boolean
+    /**
+     * (Experimental)
+     * Warn if observables are accessed outside a reactive context
+     */
+    observableRequiresReaction?: boolean
     computedConfigurable?: boolean
     isolateGlobalState?: boolean
     disableErrorBoundaries?: boolean
@@ -19,7 +29,9 @@ export function configure(options: {
         computedRequiresReaction,
         computedConfigurable,
         disableErrorBoundaries,
-        reactionScheduler
+        reactionScheduler,
+        reactionRequiresObservable,
+        observableRequiresReaction
     } = options
     if (options.isolateGlobalState === true) {
         isolateGlobalState()
@@ -53,6 +65,14 @@ export function configure(options: {
     }
     if (computedRequiresReaction !== undefined) {
         globalState.computedRequiresReaction = !!computedRequiresReaction
+    }
+    if (reactionRequiresObservable !== undefined) {
+        globalState.reactionRequiresObservable = !!reactionRequiresObservable
+    }
+    if (observableRequiresReaction !== undefined) {
+        globalState.observableRequiresReaction = !!observableRequiresReaction
+
+        globalState.allowStateReads = !globalState.observableRequiresReaction
     }
     if (computedConfigurable !== undefined) {
         globalState.computedConfigurable = !!computedConfigurable
--- a/src/core/derivation.ts
+++ b/src/core/derivation.ts
@@ -55,6 +55,11 @@ export interface IDerivation extends IDepTreeNode {
     __mapid: string
     onBecomeStale(): void
     isTracing: TraceMode
+
+    /**
+     *  warn if the derivation has no dependencies after creation/update
+     */
+    requiresObservable?: boolean
 }
 
 export class CaughtException {
@@ -155,12 +160,23 @@ export function checkIfStateModificationsAreAllowed(atom: IAtom) {
         )
 }
 
+export function checkIfStateReadsAreAllowed(observable: IObservable) {
+    if (
+        process.env.NODE_ENV !== "production" &&
+        !globalState.allowStateReads &&
+        globalState.observableRequiresReaction
+    ) {
+        console.warn(`[mobx] Observable ${observable.name} being read outside a reactive context`)
+    }
+}
+
 /**
  * Executes the provided function `f` and tracks which observables are being accessed.
  * The tracking information is stored on the `derivation` object and the derivation is registered
  * as observer of any of the accessed observables.
  */
 export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T, context: any) {
+    const prevAllowStateReads = allowStateReadsStart(true)
     // pre allocate array allocation + room for variation in deps
     // array will be trimmed by bindDependencies
     changeDependenciesStateTo0(derivation)
@@ -181,7 +197,26 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T, con
     }
     globalState.trackingDerivation = prevTracking
     bindDependencies(derivation)
+
+    warnAboutDerivationWithoutDependencies(derivation)
+
+    allowStateReadsEnd(prevAllowStateReads)
+
     return result
+}
+
+function warnAboutDerivationWithoutDependencies(derivation: IDerivation) {
+    if (process.env.NODE_ENV === "production") return
+
+    if (derivation.observing.length !== 0) return
+
+    if (globalState.reactionRequiresObservable || derivation.requiresObservable) {
+        console.warn(
+            `[mobx] Derivation ${
+                derivation.name
+            } is created/updated without reading any observable value`
+        )
+    }
 }
 
 /**
@@ -276,6 +311,16 @@ export function untrackedStart(): IDerivation | null {
 
 export function untrackedEnd(prev: IDerivation | null) {
     globalState.trackingDerivation = prev
+}
+
+export function allowStateReadsStart(allowStateReads: boolean) {
+    const prev = globalState.allowStateReads
+    globalState.allowStateReads = allowStateReads
+    return prev
+}
+
+export function allowStateReadsEnd(prev: boolean) {
+    globalState.allowStateReads = prev
 }
 
 /**
--- a/src/core/observable.ts
+++ b/src/core/observable.ts
@@ -7,7 +7,8 @@ import {
     TraceMode,
     getDependencyTree,
     globalState,
-    runReactions
+    runReactions,
+    checkIfStateReadsAreAllowed
 } from "../internal"
 
 export interface IDepTreeNode {
@@ -131,6 +132,8 @@ export function endBatch() {
 }
 
 export function reportObserved(observable: IObservable): boolean {
+    checkIfStateReadsAreAllowed(observable)
+
     const derivation = globalState.trackingDerivation
     if (derivation !== null) {
         /**
@@ -151,6 +154,7 @@ export function reportObserved(observable: IObservable): boolean {
     } else if (observable.observers.size === 0 && globalState.inBatch > 0) {
         queueForUnobservation(observable)
     }
+
     return false
 }
 
--- a/test/base/strict-mode.js
+++ b/test/base/strict-mode.js
@@ -1,3 +1,6 @@
+/**
+ * @type {typeof import("../../src/mobx")}
+ */
 const mobx = require("../../src/mobx.ts")
 const utils = require("../utils/test-utils")
 
@@ -224,7 +227,7 @@ test("enforceActions 'strict' should not throw exception while observable array 
     }
 })
 
-test("warn on unsafe reads", function() {
+test("warn on unsafe reads of computed", function() {
     try {
         mobx.configure({ computedRequiresReaction: true })
         const x = mobx.observable({
@@ -239,6 +242,124 @@ test("warn on unsafe reads", function() {
     } finally {
         mobx.configure({ computedRequiresReaction: false })
     }
+})
+
+describe("observableRequiresReaction", function() {
+    test("warn on unsafe reads of observable", function() {
+        try {
+            mobx.configure({ observableRequiresReaction: true })
+            const x = mobx.observable({
+                y: 3
+            })
+            utils.consoleWarn(() => {
+                x.y
+            }, /being read outside a reactive context/)
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+
+    test("warn on unsafe reads of observable also when there are other subscriptions", function() {
+        try {
+            mobx.configure({ observableRequiresReaction: true })
+            const x = mobx.observable({
+                y: 3
+            })
+
+            const dispose = mobx.autorun(() => x.y)
+
+            utils.consoleWarn(() => {
+                x.y
+            }, /being read outside a reactive context/)
+
+            dispose()
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+
+    test("warn on unsafe reads of observable array", function() {
+        try {
+            mobx.configure({ observableRequiresReaction: true })
+            const x = mobx.observable({
+                arr: [1, 2, 3]
+            })
+            utils.consoleWarn(() => {
+                x.arr[1]
+            }, /being read outside a reactive context/)
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+    test("don't warn on reads inside a computed", function() {
+        try {
+            mobx.configure({ observableRequiresReaction: true })
+            const x = mobx.observable({
+                y: 1
+            })
+
+            const fooComputed = mobx.computed(() => x.y + 1)
+
+            const messages = utils.supressConsole(() => {
+                const dispose = mobx.autorun(() => fooComputed.get())
+                dispose()
+            })
+
+            expect(messages.length).toBe(0)
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+
+    test("don't warn on reads inside an action", function() {
+        try {
+            mobx.configure({ observableRequiresReaction: true })
+            const x = mobx.observable({
+                y: 1
+            })
+
+            const fooAction = mobx.action(() => x.y)
+
+            const messages = utils.supressConsole(() => {
+                fooAction()
+            })
+
+            expect(messages.length).toBe(0)
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+
+    test("warn on reads inside a transaction", function() {
+        try {
+            mobx.configure({ observableRequiresReaction: true })
+            const x = mobx.observable({
+                y: 1
+            })
+
+            utils.consoleWarn(() => {
+                mobx.transaction(() => x.y)
+            }, /being read outside a reactive context/)
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+})
+
+describe("reactionRequiresObservable", function() {
+    test("warn on reaction creation without dependencies", function() {
+        try {
+            mobx.configure({ reactionRequiresObservable: true })
+
+            utils.consoleWarn(() => {
+                const dispose = mobx.reaction(() => "plain value", newValue => newValue)
+
+                dispose()
+            }, /is created\/updated without reading any observable value/)
+        } finally {
+            mobx.configure({ reactionRequiresObservable: false })
+        }
+    })
 })
 
 test("#1869", function() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 5.14.0 / 4.14.0
+
+* Added experimental `reactionRequiresObservable` & `observableRequiresReaction` config [#2079](https://github.com/mobxjs/mobx/pull/2079), [Docs](https://github.com/mobxjs/mobx/pull/2082)
+* Added experimental `requiresObservable` config to `reaction`, `autorun` & `when` options [#2079](https://github.com/mobxjs/mobx/pull/2079), [Docs](https://github.com/mobxjs/mobx/pull/2082)
+
 # 5.13.1 / 4.13.1
 
 * Don't use `global` and `self` keywords unless defined. Fixes [#2070](https://github.com/mobxjs/mobx/issues/2070).
--- a/flow-typed/mobx.js
+++ b/flow-typed/mobx.js
@@ -5,6 +5,8 @@ export type IObservableMapInitialValues<K, V> = IMapEntries<K, V> | KeyValueMap<
 export interface IMobxConfigurationOptions {
     +enforceActions?: boolean | "strict" | "never" | "always" | "observed";
     computedRequiresReaction?: boolean;
+    reactionRequiresObservable?: boolean;
+    observableRequiresReaction?: boolean;
     isolateGlobalState?: boolean;
     disableErrorBoundaries?: boolean;
     arrayBuffer?: number;
@@ -16,6 +18,10 @@ declare export function configure(options: IMobxConfigurationOptions): void
 export interface IAutorunOptions {
     delay?: number;
     name?: string;
+    /**
+     *  warn if the derivation has no dependencies after creation/update
+     */
+    requiresObservable?: boolean;
     scheduler?: (callback: () => void) => any;
     onError?: (error: any) => void;
 }
--- a/src/api/autorun.ts
+++ b/src/api/autorun.ts
@@ -15,6 +15,11 @@ import {
 export interface IAutorunOptions {
     delay?: number
     name?: string
+    /**
+     * Experimental.
+     * Warns if the view doesn't track observables
+     */
+    requiresObservable?: boolean
     scheduler?: (callback: () => void) => any
     onError?: (error: any) => void
 }
@@ -48,7 +53,8 @@ export function autorun(
             function(this: Reaction) {
                 this.track(reactionRunner)
             },
-            opts.onError
+            opts.onError,
+            opts.requiresObservable
         )
     } else {
         const scheduler = createSchedulerFromOptions(opts)
@@ -66,7 +72,8 @@ export function autorun(
                     })
                 }
             },
-            opts.onError
+            opts.onError,
+            opts.requiresObservable
         )
     }
 
@@ -89,8 +96,8 @@ function createSchedulerFromOptions(opts: IReactionOptions) {
     return opts.scheduler
         ? opts.scheduler
         : opts.delay
-            ? (f: Lambda) => setTimeout(f, opts.delay!)
-            : run
+        ? (f: Lambda) => setTimeout(f, opts.delay!)
+        : run
 }
 
 export function reaction<T>(
@@ -131,7 +138,8 @@ export function reaction<T>(
                 scheduler!(reactionRunner)
             }
         },
-        opts.onError
+        opts.onError,
+        opts.requiresObservable
     )
 
     function reactionRunner() {
--- a/src/api/when.ts
+++ b/src/api/when.ts
@@ -11,6 +11,11 @@ import {
 export interface IWhenOptions {
     name?: string
     timeout?: number
+    /**
+     * Experimental.
+     * Warns if the view doesn't track observables
+     */
+    requiresObservable?: boolean
     onError?: (error: any) => void
 }
 
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -11,6 +11,7 @@ import {
     untrackedEnd,
     untrackedStart
 } from "../internal"
+import { allowStateReadsStart, allowStateReadsEnd } from "./derivation"
 
 export interface IAction {
     isMobxAction: boolean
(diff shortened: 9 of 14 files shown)


src/api/configure.ts
```diff
@@ -54,6 +58,12 @@ export function configure(options: {
     if (computedRequiresReaction !== undefined) {
         globalState.computedRequiresReaction = !!computedRequiresReaction
     }
+    if (derivationRequiresObservable !== undefined) {
```
FredyC: Shouldn't both of these be enabled by default? I mean it just adds some safety checks which is a good thing. If someone wants to feel less protected, they can opt-out.

src/core/derivation.ts
```diff
@@ -181,9 +181,21 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T, con
...
+        warnAboutDerivationWithoutDependencies(derivation.name)
+    }
+
     return result
 }
 
+function warnAboutDerivationWithoutDependencies(name: string) {
+    if (process.env.NODE_ENV === "production") return
+    if (globalState.derivationRequiresObservable) {
+        console.warn(`[mobx] Derivation ${name} is created without reading any observable value`)
```
FredyC: I wonder if we can provide information given that name is usually generated and doesn't make much sense. It would make sense to spill out stack trace at least. Same for the other warning.

src/api/configure.ts
Bnaya: I went with the spirit of `computedRequiresReaction`, that is not on by default. 
I personally think developers will benefit from it, but i believe in soft changes

src/core/derivation.ts
Bnaya: I took the behaviour of `computedRequiresReaction`,
But i don't have any objection to add it across the board.
We can add to the docs that you can use the debugger to add a breakpoint to see the call stack

Bnaya:
I'm not sure how to update the changelog, do we have any script that do that, or manually?

src/core/derivation.ts
FredyC: > We can add to the docs that you can use the debugger to add a breakpoint to see the call stack

Um, that's hardly a good DX :) I believe the majority of people still don't know how to use debugger properly.

Alright, let's keep it like that in this PR and improve later. I am especially curious how well it will integrate with `mobx-react` and usefulness of these warnings.

FredyC:
> I'm not sure how to update the changelog, do we have any script that does that, or manually?

Still manually. And tick other checkboxes as well when completed :)

FredyC:
I see that 5.13.1 / 4.13.1 is still unpublished, but considering this is a new feature it should go into `5.14.0 / 4.14.0` and we shall publish before merging this I suppose.

Bnaya:
I've pushed the changelog update

urugator:
Change *observable**s**RequiresReaction* to *observableRequiresReaction*
so it's gramatically correct and matches other two

Change ***derivation**RequiresObservable* to ***reaction**RequiresObservable*
to match the other two

`requiresObservable` option should be added to `reaction/autorun` (similary to `computed`'s `requiresReaction`)

Note on defaults:
`observableRequiresReaction` can't be true by default, because `action`s are optional by default and working with observable models outside of reactive context is completely valid (in practice I think it can be useful to turn this on after init logic).

`reactionRequiresObservable` - while reactions without observables are de-facto an error, being able to slap `observer` everywhere without thinking about deps is also useful

Could you add some test cases where observables are accesses indirectly (through computed) or inside action (eg inside autorun)?

Bnaya:
> 
> `requiresObservable` option should be added to `reaction/autorun` (similary to `computed`'s `requiresReaction`)
> 
Can you please elaborate?

> Note on defaults:
> `observableRequiresReaction` can't be true by default, because `action`s are optional by default and working with observable models outside of reactive context is completely valid (in practice I think it can be useful to turn this on after init logic).
> 
> `reactionRequiresObservable` - while reactions without observables are de-facto an error, being able to slap `observer` everywhere without thinking about deps is also useful
> 
I personally think it's not a good practice, and the developer should be aware to what gonna render and why.
And when you are working in environment of mixed react & mobx state it can get really messy.
I don't want "dump" components to be wrapped with `observer`   
If i could do that in the type system/static analysis, i would totally do.
But i'm really OK with keeping them both optional.

> Could you add some test cases where observables are accesses indirectly (through computed) or inside action (eg inside autorun)?
I will

urugator:
> Can you please elaborate?

Ability to configure this per `reaction` - `reaction(deriv, effect, { requiresObservable: true })`
Related: https://github.com/mobxjs/mobx/issues/1197#issuecomment-368829413

OT:
> the developer should be aware to what gonna render and why

Why not to use explicit subscription then?

FredyC:
> being able to slap `observer` everywhere without thinking about deps is also useful

I also don't agree, why would someone want to deoptimize on purpose? I think it comes more from the current lack of validation fixed by this PR that kinda forces you to just "slap observer everywhere". I don't see an actual reason for doing it otherwise.

That said, with `requiresObservable` on reactions would give that choice which is fine. I would probably enable it by default for `observer` anyway. And it should fall back to the global setting if there is none specified for a reaction.

src/core/observable.ts
```diff
@@ -148,8 +148,11 @@ export function reportObserved(observable: IObservable): boolean {
             }
         }
         return true
-    } else if (observable.observers.size === 0 && globalState.inBatch > 0) {
-        queueForUnobservation(observable)
+    } else if (observable.observers.size === 0) {
+        if (globalState.inBatch > 0) {
+            queueForUnobservation(observable)
+        }
+        warnAboutUntrackedObservableRead(observable.name)
```
urugator: This warns only if there are no observers, but we want to warn anytime outside of derivation (and not in batch), right?.
The motivation behind `computedRequiresReaction` was a bit different - to avoid erasing it's cache - so it doesn't matter where the computed is accessed as far as there is at least one reaction keeping the cache alive.

urugator:
```javascript
test("warn on unsafe reads of observable 1", function() {
    try {
        mobx.configure({ observablesRequiresReaction: true })
        const x = mobx.observable({
            y: 3
        })               
 
        autorun(() => x.y); // <-- ARBITRARY OBSERVER (observers.size !== 0)

        utils.consoleWarn(() => {
            x.yy
        }, /being read outside a reactive context/)
    } finally {
        mobx.configure({ observablesRequiresReaction: false })
    }
})
```

urugator:
The lib is all about automatic subscription management. The idea is that you don't have to worry about what depends on what and you shouldn't experience staleness once you fulfill the contract - `observable` on data and `observer` on component.
Making everything observable/observer is an easy to follow and straithforward way to do that. 

> comes more from the current lack of validation

Agree, but as mentioned the fix is not generally applicable, eg it makes sense only with `enforceAction('strict')` and I am afraid there may be some false positives.

What's the expected behavior of:
```javascript
mobx.configure({ observablesRequiresReaction: true })
autorun(() => {
  untracked(() => x.y);
})
```
Can you add it to tests as well please?

How do you feel about labelling this "experimental" and waiting for some feedback?

Bnaya:
I prefer to keep it optional and it's totally can be marked experimental so we could tweak the behaviour later without breaking changes.

> reaction(deriv, effect, { requiresObservable: true })
We will need to also change mobx-react api to expose also that
should it fail or warn? looking at `computedvalue` code, i think there's something not optimal there.
on dev time it will `fail` and on prod time it will just work.
its different behaviour between prod & dev, not just warning on dev
https://github.com/mobxjs/mobx/blob/master/src/core/computedvalue.ts#L270-L289

urugator:
Ignore on production for performance sake I suppose...
I think that the global configuration only warns (rather failing) so there is some space for intentionally exceptional behavior (some workaround perhaps) without crashing the app.

Bnaya:
Another thing that came up in my mind:
Add warning when `action` didn't change any observable value.
But maybe better keep to to another PR?

urugator:
> Add warning when action didn't change any observable value.

The change can be conditional or deferred:
```javascript
@action doSomething(a) {
  if (this.cond) {
     this.x = a;
  }
}

@action doSomething() {
  // action is untracked to avoid accidental subscription (so it's not entirely useless even if you don't change anything):
  const param = this.y;
  xhr.post(param).then(() => this.x = "a");
}
```

Bnaya:
changes pushed
After we will settle this PR, i will make the respective changes to mobx 4 & docs PRs

test/base/strict-mode.js
```diff
@@ -241,6 +244,147 @@ test("warn on unsafe reads", function() {
...
 
+describe("observableRequiresReaction", function() {
+    test("warn on unsafe reads of observable", function() {
+        try {
+            mobx.configure({ observableRequiresReaction: true })
+            const x = mobx.observable({
+                y: 3
+            })
+            utils.consoleWarn(() => {
+                x.yy
```
urugator: typo?

src/core/derivation.ts
```diff
@@ -181,9 +186,25 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T, con
...
+    if (derivation.observing.length === 0) {
+        warnAboutDerivationWithoutDependencies(derivation)
+    }
+
     return result
 }
 
+function warnAboutDerivationWithoutDependencies(derivation: IDerivation) {
+    if (process.env.NODE_ENV === "production") return
+    if (globalState.reactionRequiresObservable || derivation.requiresObservable) {
```
urugator: fail on `derivation.requiresObservable` like in computed? https://github.com/mobxjs/mobx/blob/2bce97d1e60ae819c25c8f8b04c764f40855e4f6/src/core/computedvalue.ts#L272-L273

test/base/strict-mode.js
```diff
@@ -241,6 +244,147 @@ test("warn on unsafe reads", function() {
...
+                arr: [1, 2, 3]
+            })
+            utils.consoleWarn(() => {
+                x.arr[1]
+            }, /being read outside a reactive context/)
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+    test("don't warn on reads inside batch - computed", function() {
```
urugator: `inside batch - computed` => `inside computed` (not a batch)

test/base/strict-mode.js
```diff
@@ -241,6 +244,147 @@ test("warn on unsafe reads", function() {
...
+                mobx.transaction(() => x.y)
+            })
+
+            expect(messages.length).toBe(0)
+        } finally {
+            mobx.configure({ observableRequiresReaction: false })
+        }
+    })
+
+    test("don't warn on reads inside batch - untracked inside autorun", function() {
```
urugator: `inside batch - untracked inside autorun` => `inside untracked` (not a batch)

test/base/strict-mode.js
```diff
@@ -241,6 +244,147 @@ test("warn on unsafe reads", function() {
...
+
+            const messages = utils.supressConsole(() => {
+                const disposer = mobx.autorun(() => {
+                    mobx.untracked(() => x.y)
+                })
+
+                disposer()
+            })
+
+            expect(messages.length).toBe(0)
```
urugator: I don't understand why this passes....
The `untracked` sets derivation to null:
https://github.com/mobxjs/mobx/blob/251bb743860bc3f20809cc802a8330abf278dd68/src/core/derivation.ts#L294
and there is no batch ... therefore `x.y` should warn ... correct?

Bnaya: But its still inside a batch, i've checked the code.

src/core/derivation.ts
Bnaya: `computed` is different, it throws the error to the reading context.

 autorun/reaction wrapped in the mobx error boundaries,
I've actually played with `fail` and i'm really not sure what should be the behaviour.
So for simplicity i think it better be that way

test/base/strict-mode.js
Bnaya: its also inside a batch,
That why its not warning

test/base/strict-mode.js
Bnaya: its also inside a batch, that why its not warning

test/base/strict-mode.js
Bnaya: yes

test/base/strict-mode.js
Bnaya: Reaction uses startBatch inside,
And untracked only removed the `trackingDerivation`, not batch

https://github.com/mobxjs/mobx/blob/f944b5413f2f4b9de89e00e7de1ddde2a178e19d/src/core/reaction.ts#L92-L116

https://github.com/mobxjs/mobx/blob/f944b5413f2f4b9de89e00e7de1ddde2a178e19d/src/core/derivation.ts#L271-L279

urugator: Ah right ... all reactions are inside batch ...

urugator: Ok, so currently it doesn't warn because there is a batch ... but it shouldn't warn for different reason - existing derivation (or actually nonexsiting in case of `untracked`).
The goal is to force `x.y` being accessed either inside `reaction/autorun` or inside `action/tx`... right?
Is there a derivation which doesn't run inside batch so that `untracked` would warn?
Or similary ... is there a batch without `action/tx` so that `x.y` wouldn't warn?

urugator: `enforceActions` uses `globalState.allowStateChanges`, shouldn't we use it instead of `inBatch` to detect allowed access outside derivation?

urugator: Or perhpas we should have `globalState.allowStateReads`, which is true in action/derivation. That way the `untracked` wouldn't warn even with missing derivation.

Bnaya: if we got to that point,
Can we chat about it directly and make an implementation plan?

urugator: I've started putting together a list of what needs to be done, but there is something I came across:  
`checkIfStateModificationsAreAllowed` is called individually for each mutative operation per observable type (value/array/map):
https://github.com/mobxjs/mobx/search?q=checkIfStateModificationsAreAllowed&unscoped_q=checkIfStateModificationsAreAllowed 
I wonder why it isn't called always during `atom.reportChanged()` as it represents a *state change*...
It also means that if you create a custom atom, it doesn't respect the `enforceAction` by itself.  
My concern is where to place eventual `checkIfStateReadsAreAllowed` - whether it could be placed in `atom.reportObserved()`, or whether it has to be in `observablevalue.get()` and alike.

Bnaya: reportChanged sounds like something after the action have happened,
And `checkIfStateModificationsAreAllowed` needs to be checked before that.

If we are going only to warn, maybe reportObserved can work

urugator: Right, it would throw after the state was mutated. I don't think it would really matter in the end as I don't expect anyone will try recover from the exception, but it could definitely break some expectations.
I think it's impossible to obtain observable value without calling `reportObserved()` first and I can't imagine a scenario where you would want to call `reportObserved()` without warning, so let's stick with that for now.

urugator:
1. Add `allowStateReads` to `globalState`, defaults to `true` 
https://github.com/mobxjs/mobx/blob/master/src/core/globalstate.ts

2. Add `allowStateReads(Start/End)` methods to `derivation.ts` (dunno where else to place it).
It's a copy of `allowStateChanges` from `action.ts`: 
https://github.com/mobxjs/mobx/blob/2bce97d1e60ae819c25c8f8b04c764f40855e4f6/src/core/action.ts#L96-L115

3. Update `startAction`/`endAction` to call `allowStateReads[Start/End]` (add `prevAllowStateReads` to `runInfo`)
start
https://github.com/mobxjs/mobx/blob/2bce97d1e60ae819c25c8f8b04c764f40855e4f6/src/core/action.ts#L79
end
https://github.com/mobxjs/mobx/blob/2bce97d1e60ae819c25c8f8b04c764f40855e4f6/src/core/action.ts#L89
runInfo
https://github.com/mobxjs/mobx/blob/2bce97d1e60ae819c25c8f8b04c764f40855e4f6/src/core/action.ts#L82

4. Update `derivation` to call `allowStateReads[Start/End]`

```javascript
export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T, context: any) {
    const prevAllowStateReads = allowStateReadsStart(true); // <== HERE
    // pre allocate array allocation + room for variation in deps
    // array will be trimmed by bindDependencies
    changeDependenciesStateTo0(derivation)
    derivation.newObserving = new Array(derivation.observing.length + 100)
    derivation.unboundDepsCount = 0
    derivation.runId = ++globalState.runId
    const prevTracking = globalState.trackingDerivation
    globalState.trackingDerivation = derivation
    let result
    if (globalState.disableErrorBoundaries === true) {
        result = f.call(context)
    } else {
        try {
            result = f.call(context)
        } catch (e) {
            result = new CaughtException(e)
        }
    }
    globalState.trackingDerivation = prevTracking
    bindDependencies(derivation)
    allowStateReadsEnd(prevAllowStateReads);  // <== HERE
    return result
}
```

5. Add `checkIfStateReadsAreAllowed(observable: IObservable)` to `derivation.ts` (similary to `checkIfStateModificationsAreAllowed`)
```javascript
export function checkIfStateReadsAreAllowed(observable: IObservable) {    
   if (process.env.NODE_ENV !== "production" && !globalState.allowStateReads && globalState.observableRequiresReaction) {
      console.warn(`[mobx] Observable ${atom.name} being read outside a reactive context`)      
   }        
}
```

6. Update `reportObserved` to call `checkIfStateReadsAreAllowed`
```javascript
export function reportObserved(observable: IObservable): boolean {
    checkIfStateReadsAreAllowed(observable); // <== HERE
    const derivation = globalState.trackingDerivation
    if (derivation !== null) {
        /**
         * Simple optimization, give each derivation run an unique id (runId)
         * Check if last time this observable was accessed the same runId is used
         * if this is the case, the relation is already known
         */
        if (derivation.runId !== observable.lastAccessedBy) {
            observable.lastAccessedBy = derivation.runId
            // Tried storing newObserving, or observing, or both as Set, but performance didn't come close...
            derivation.newObserving![derivation.unboundDepsCount++] = observable
            if (!observable.isBeingObserved) {
                observable.isBeingObserved = true
                observable.onBecomeObserved()
            }
        }
        return true
    } else if (observable.observers.size === 0 && globalState.inBatch > 0) {
        queueForUnobservation(observable)
    }
    return false
}
```

7. Do wee need to update `computed` somehow?

Bnaya:
@urugator Thanks for your feedback
I've implemented the requested changes, after approval i will backport to mobx4

> 7. Do wee need to update computed somehow?

I think not, but i'm still going over things

Bnaya:
@urugator friendly ping :)
Just in case you've missed it, i know it took me long time

src/api/transaction.ts
```diff
@@ -8,10 +9,12 @@ import { endBatch, startBatch } from "../internal"
  * @returns any value that was returned by the 'action' parameter.
  */
 export function transaction<T>(action: () => T, thisArg = undefined): T {
+    const prevAllowStateReads = allowStateReadsStart(true)
```
urugator: Do we wanna do this? I think you should be able to supress the warning only by accessing observable from `action` or derivation.
You know, similary to strict mode (`enforceAction`). 
The low level constructs like `untracked/transaction` should be agnostic to these flags.
Eg. `useLocalStore` from `mobx-react` wraps every method inside `transaction`, because it provides batching without messing with subscriptions or `enforceAction`. 
It's explained here: https://github.com/mobxjs/mobx-react-docz/issues/12#issuecomment-532342147
So if we `allowStateReads` in `transaction` we will effectively turn of the `observableRequiresReaction` checks for all local store methods.
I know it makes the name a bit misleading - it should be `observableRequiresReactionOrAction` or even more precisely `readingObservableRequiresReactionOrAction`. Perhaps we should change it to `enforceReaction` or something? What do you think?

src/core/derivation.ts
```diff
@@ -260,11 +294,13 @@ export function clearObserving(derivation: IDerivation) {
 }
 
 export function untracked<T>(action: () => T): T {
+    const prevAllowStateReads = allowStateReadsStart(false)
```
urugator: Same as `transaction`: https://github.com/mobxjs/mobx/pull/2079/files#r327233937

Bnaya: Note that i set it here to false,
I've added that due your comment on the pre iteration,
its more correct that you will get warnings inside `untracked`

src/api/transaction.ts
Bnaya: I've removed that

src/core/derivation.ts
urugator: I missed the `false` value, nevertheless I still think it shouldn't switch the flag.
Untracked is used to disable subscription in any context ... if we disable reads inside untracked it becomes absolutely useless (because you're either outside or reaction so it has nothing to disable or it will report error).
The idea is that if you use `untracked` from `action`/`reaction`, `globalState.allowStateReads` is already `true`, so there is no error/exception. Which seems correct to me, because the possible error (missing subscription) is introduced explicitely and intentionally. 
Previously this wasn't possible, because reporting was bound to `globalState.trackingDerivation`. That was analogical to `enforceAction` checking for `inBatch` flag, rather than `allowStateChanges`... therefore my comment and introduction of `allowStateReads`...

Bnaya: done

urugator:
Just to clarify... I didn't mind that it didn't warn inside `untracked`, I was concerned by the reason why it doesn't warn.
It didn't warn because there was a `batch`, but it shouldn't depend on a presence of `batch` ... it should depend on the presence of derivation/action - and both derivation/action was missing - so at that time one would expect it to warn (regardless of what the actual behavior should be)... Hope it makes sense.

mweststrate:
Thanks for summarising! Appreciate it :) Both features sound pretty cool
for debugging purposes. Didn't review the code, but I think these are good
features to have. Naming is hard indeed, but at least there is a cool
symmetry :-P.

Bnaya:
I've updated the mobx 4 backport PR

src/api/configure.ts
```diff
@@ -54,6 +66,14 @@ export function configure(options: {
     if (computedRequiresReaction !== undefined) {
         globalState.computedRequiresReaction = !!computedRequiresReaction
     }
+    if (reactionRequiresObservable !== undefined) {
+        globalState.reactionRequiresObservable = !!reactionRequiresObservable
+    }
+    if (observableRequiresReaction !== undefined) {
+        globalState.observableRequiresReaction = !!observableRequiresReaction
```
mweststrate: if `observableRequiresReaction` is set, we might want to assert that some mode of `enforceActions` is also set?
It also implies `computedRequiresReaction` I guess

src/core/derivation.ts
```diff
@@ -155,12 +160,23 @@ export function checkIfStateModificationsAreAllowed(atom: IAtom) {
         )
 }
 
+export function checkIfStateReadsAreAllowed(observable: IObservable) {
+    if (
+        process.env.NODE_ENV !== "production" &&
+        !globalState.allowStateReads &&
+        globalState.observableRequiresReaction
+    ) {
+        console.warn(`[mobx] Observable ${observable.name} being read outside a reactive context`)
```
mweststrate: Maybe make the error a bit clearer for those not fully familiar with these concept names? ".. is being read outside an action or reactive context, such as 'autorun', 'reaction', 'when' or 'observer'"

src/core/derivation.ts
```diff
@@ -181,9 +197,27 @@ export function trackDerivedFunction<T>(derivation: IDerivation, f: () => T, con
     }
     globalState.trackingDerivation = prevTracking
     bindDependencies(derivation)
+
+    if (derivation.observing.length === 0) {
```
mweststrate: would be nice to move the production check to here for better code elimination (if the warning function is called only once)

src/core/observable.ts
```diff
@@ -131,6 +132,8 @@ export function endBatch() {
 }
 
 export function reportObserved(observable: IObservable): boolean {
+    checkIfStateReadsAreAllowed(observable)
```
mweststrate: NIT: I think this can be moved into the else branch at line 154? (since these lines are hit millions of time, trying to move stuff out of the critical as much as possible :))

urugator: We need to check for reads anytime an observable is accessed regardless of `derivation`/`batch`/`observers`.
Initially the PR relied on the presence of batch/derivation, but low level tools like `untracked/transaction` were modifing the behavior, conflicting with the intention of requiring `reaction/autorun`.        
So we introduced `allowStateReads`, so we don't depend on `batch/derivation`, similary like `enforceAction` doesn't depend on a presence `batch`, but uses it's own flag `allowStateChanges`.
The way to optimize this would be to move the check into individual "reading" operations of observable structures (value/array/map/object) etc. It's discussed here: 
https://github.com/mobxjs/mobx/pull/2079#discussion_r317235819

src/api/configure.ts
urugator: It's true that `computedRequiresReaction` makes most sense in conjunction with `enforceAction: "observed/always"`
But strictly speaking isn't required... you can still perform writes `this.x = 5;` (without reads) outside of `action/reaction`. 

`computedRequiresReaction` works a bit differently as it has different purpose - it was introduced to prevent `cache` purging.
As a result it actually doesn't matter from where the `computed` is accessed as long as there is at least one observer. 
In contrast `observableRequiresReaction` makes the `observable` inaccessible outside `reaction`/`action`, no matter how many observers.
Currently `observableRequiresReaction` has no effect on `computed`, as there is no need to ... it will throw anyway on the first observable accessed from within the computed...

src/core/derivation.ts
Bnaya: I want to point out that when using react for example, that message can be confusing.
The developer might thow an `autorun` inside render to make it go away

BTW I took that message from:
https://github.com/mobxjs/mobx/blob/master/src/core/computedvalue.ts#L282-L287
So maybe we should make the change in both place

src/api/configure.ts
mweststrate: Check, for the first one I hoped it would make error messages better, but now that I think of it, it wouldn't help if an non-annoted action would start of with a read, so probably we should add the possibliity that an action might be needed to the error message.

Agreed on the second one, it was more of an observation than anything else :)

src/core/derivation.ts
mweststrate: Agreed, should probably change it in both places. Also, in this case, we might also want to say something along the lines: "... or mark the surrounding function as `action` to indicate that there are imperative state reads and writes"? Not sure how to phrase it right, (maybe we should link to some document to have more explanation room), but I am concerned about the case discussed here: https://github.com/mobxjs/mobx/pull/2079#discussion_r328116458

If something is intended to be an action, but not marked, then getting the above error on the first read that happens in the 'action' might be very misleading.

src/core/observable.ts
mweststrate: Check! I'm clearly late to the show :)

src/core/derivation.ts
Bnaya: Done!

src/api/configure.ts
Bnaya: Is there any item from here?

src/core/derivation.ts
Bnaya: I'm also not very good with message :)
these options are opt-it,
Maybe we should just tell the developer to look for `observableRequiresReaction` and elaborate in the docs.

>observableRequiresReaction violation: observable ${observable.name} being read outside a reactive context

src/core/derivation.ts
Bnaya: Done

src/core/derivation.ts
Bnaya: As these feature marked as experimental, i would prefer to push it forward and get feedback, and then improve developer experience

mweststrate:
Sure, feel free to merge! I'll add a note to self to tinker another time
about messaging

Bnaya:
I've merged the PRS!

What's the policy regarding publishing?
also, i wish we had something that publishes canary versions for each PR or so

mweststrate:
There is currently no policy :'). Just some ideas to incorporate `np` like
done in immer. But I'll try to cut a release somewhere today. TY!

mweststrate:
Released!

Bnaya:
I've created a blog post regarding these features:
https://medium.com/@bnaya/mobx-5-14-0-will-help-you-stay-reactive-and-only-when-you-need-to-2e897d3bfabe
I would love to have some feedback before publishing

mweststrate:
Looks awesome! I think it is pretty clear. Make sure to link it in the
docs! Maybe a link on top of https://mobx.js.org/best/trace.html would be
neat as well.

PR #2081, opened 2019-08-19
Use yarn-deduplicate with NPX

FredyC:
Reverts #2078 by @Bnaya and the `yarn-deduplicate` is used with NPX instead of having it in direct deps which is unnecessary for something like that.

As a side effect, now we can actually see (in that separate 2nd commit) what has that tool done and apparently it's not such a huge problem as it it's claimed to be :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,8 +28,7 @@
     "small-build": "node scripts/build.js",
     "lint": "eslint src/**/*.ts",
     "size": "size-limit",
-    "publish-script": "node scripts/publish.js",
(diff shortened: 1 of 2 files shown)


Bnaya:
One more `yarn install` run after the dedup is missing,
the lock file is not stable

FredyC:
Well then, more reason not to worry about it :) I am not seeing a benefit of deduping unless there is some actual issue. Especially considering that all deps are dev only and won't affect consumers.

PR #2082, opened 2019-08-19
Docs for #2079

Bnaya:
Docs for #2079

### DIFF
--- a/docs/refguide/api.md
+++ b/docs/refguide/api.md
@@ -349,7 +349,8 @@ Returns a disposer function to cancel the side effect.
 - **`name?: string`**: A name for easier identification and debugging
 - **`delay?: number`**: the sideEffect will be delayed and debounced with the given `delay`. Defaults to `0`.
 - **`onError?: (error) => void`**: error handler that will be triggered if the autorun function throws an exception
-- **`scheduler?: (callback) => void`**: Set a custom scheduler to determine how re-running the autorun function should be scheduled
+- **`scheduler?: (callback) => void`**: Set a custom scheduler to determine how re-running the autorun function should be scheduled  
+- **`requiresObservable?: boolean`** Enables [`reactionRequiresObservable`](#reactionrequiresobservable-boolean) locally for the autorun
 
 
 ### `when`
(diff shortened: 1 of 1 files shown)


docs/refguide/api.md
```diff
@@ -520,6 +520,22 @@ Use this if you want to check whether you are using computed properties without
...
+Warns about any unobserved observable access.  
+Use this if you want to check whether you are using observables without a reactive context (eg react component without observer wrapping).
+
+```javascript
+configure({ observablesRequiresReaction: true });
+```
+
+#### `derivationRequiresObservable: boolean`
+Warns when a derivation (reaction, autorun, observer react component) is created without any observable access.  
+Use this to check whether you are unneededly wrapping react components with `observer`, or to find possible related bugs.
```
FredyC: Not sure if to mention React directly like that in here, it might sound like it's only use case :)

Besides, this is applied globally, is it influence everything, even manually created derivations inside the components.

docs/refguide/api.md
```diff
@@ -520,6 +520,22 @@ Use this if you want to check whether you are using computed properties without
 configure({ computedRequiresReaction: true });
 ```
 
+#### `observablesRequiresReaction: boolean`
+Warns about any unobserved observable access.  
+Use this if you want to check whether you are using observables without a reactive context (eg react component without observer wrapping).
```
FredyC: Reading it now like that, I wonder, are those two options kinda mutually exclusive? I mean when you would use observable in a component without an observer, it would trigger both, wouldn't? That could be more confusing than helpful imo.

docs/refguide/api.md
Bnaya: It's the most down to earth example i could come up with.
i can write a whole blog post on these 2 options, but i'm not sure if that api reference is the place. i can omit that it its a problem

docs/refguide/api.md
Bnaya: They kinda complete each other.

example user story:
I developer might have an observer component, but he made the access to the observable value behind an `if` that is not based on observable value (local state or regular prop)
on that case, `derivationRequiresObservable` will help him. (same for autorun/reaction)

Maybe that why its good they will be opt-in

FredyC: I am more worried about other scenario where you forget observer and then you would be getting a warning from both of these checks if I understand it correctly.

We should definitely let the whole thing boil a bit more and try it in React world.

Bnaya: If you forgot an observer, you won't get the `derivationRequiresObservable` warn,
because you are not inside a `derivation`

mweststrate:
I think it would be great if section 4.6 and/or 8.4 could be extended, hinting how these features could be used to detect mistakes

Bnaya:
Merge it after release?

Bnaya:
I don't see it in the docs site,
What's the status of gh-pages publishing?

mweststrate:
Sorry, only just published, as I need my old laptop for that :) (until new
docs are done)

PR #2086, opened 2019-08-21
Start track

coppyC:
PR checklist:

* [o] Added unit tests
* [ ] Updated changelog
* [o] Updated docs (either in the description of this PR as markdown, or as separate PR on the `gh-pages` branch. Please refer to this PR). For new functionality, at least [API.md](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md) should be updated

### startTrack
# [API.md]
### startTrack
now, you can endTrack in anywhere
```js
import {startTrack, observable} from 'mobx'

let a = mobx.observable.box(1)

const endTrack = startTrack(() => console.log('a had change'))
a.get()
endTrack()

a.set(2)  //  log 'a had change'
a.set(3)  // nothing log, it's one-time
```

something need to pay an attention，when you had child track,
you should guarantee the child track end before the parent track
```js
let a = mobx.observable.box(1)
let b = mobx.observable.box(1)
let values = []

const endTrackA = startTrack(() => console.log('A'))
a.get()
const endTrackB = startTrack(() => console.log('B'))
b.get()
// endTrackA() don't do this, this is error
endTrackB()
endTrackA()  // right

b.set(2)  // log 'B', without log 'A'
// trackA can't track the content of trackB

a.set(2)  // log 'A'
// but trackA can track the content out of trackB
```

can be disposed
```js
let a = mobx.observable.box(1)

const endTrack = startTrack(() => console.log('track successful'))
a.get()
const dispose = endTrack()

dispose()
a.set(2) // nothing log 
```


FredyC:
You need to update to latest master and resolve conflicts.

coppyC:
update to the latest master

urugator:
Is there some background why this is needed?
I have a feeling it was already suggested/discussed somewhere, dunno where.

Why the formatting changes in `observable.ts` and `object-api.ts`?

coppyC:
sorry，i have a cold. latter to reply.

coppyC:
I want to design a api fusion observable and observer like this 
```js
function A () {
  const state = useMobX({a: 0})
  return <div onClick={() => ++state.a}>{state.a}</div>
}
```
I don't know when the function execute over, I only can use the useEffect instand
```ts
import { startTrack, observable } from 'mobx'
import { useState, useEffect } from 'react'

function useForceUpdate() {
  const [_, update] = useState(() => Symbol())
  return () => update(Symbol())
}

export function useMobx<T extends object>(state: T = {} as T): T {
  state = useState(() => observable(state))[0]
  const forceUpdate = useForceUpdate()
  const endTrack = startTrack(forceUpdate)
  useEffect(() => {
    const dispose = endTrack()
    return dispose
  })
  return state
}
```
this code has many problem, if has 2 brother component, the first track will end before the next track
and it will track many component out of it self

I improvement the code, make the previous track end before the next track start
```ts
import { startTrack, observable } from 'mobx'
import { useState, useEffect } from 'react'

let endPrevTrack: () => void

function useForceUpdate() {
  const [_, update] = useState(() => Symbol())
  return () => update(Symbol())
}

export default function<T extends object>(state: T = {} as T): T {
  endPrevTrack && endPrevTrack()
  let dispose: () => void
  state = useState(() => observable(state))[0]
  const forceUpdate = useForceUpdate()
  const endTrack = startTrack(forceUpdate)
  endPrevTrack = () => {
    dispose = endTrack()
    endPrevTrack = undefined as any
  }
  useEffect(() => {
    endPrevTrack && endPrevTrack()
    return dispose
  })
  return state
}
```

it resolve sequence question, but the code is too dirty.
It may track more than one component, although the extra component don't have observable variable.
it's not perfect, and I stop it.

now I think it's an immature idea, and the startTrack export as a public api will easier to make more bugs, becauseof the endTrack sequence.

I will close this pr.

PR #2093, opened 2019-09-02
improved typings for flows in ts 3.6

xaviergonz:
I found a way to make flows strongly typed in TS3.6

Pros:
- proper return types, proper arguments, etc.

Cons:
- requires TS3.6, so it would be a breaking change (altenatively the multiple typings per version is leveraged, using this https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#version-selection-with-typesversions )
- requires the usage of yield* and a new _await function (though yield would still work as usual, so that one wouldn't be breaking)

e.g.
```ts
test("new flow typings", async () => {
    const otherFlow = flow(function* () {
        // x is type is number
        const x = yield* _await(Promise.resolve(3))
        return x
    })

    const otherFlow2 = flow(function* () {
        // typings could be changed to make it work, 
        // but I think it is better to enforce promises on the types
        const x = yield* _await(5 as any as Promise<number>)
        return x
    })

    const start = flow(function* () {
        // a is number
        const a = yield* _await(Promise.resolve(2))
        // b is number
        const b = yield* _await(otherFlow())
        // c is number
        const c = yield* _await(otherFlow2())
        return a + b + c
    })

    // res is number
    const res = await start()
    expect(res).toBe(3 + 2 + 5)
})
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,8 +76,8 @@
     "size-limit": "^1.3.3",
     "tape": "^4.10.1",
     "ts-jest": "^24.0.0",
-    "tslib": "^1.9.3",
-    "typescript": "^3.3.3333"
+    "tslib": "^1.10.0",
+    "typescript": "^3.6.2"
   },
   "dependencies": {},
   "keywords": [
--- a/src/api/flow.ts
+++ b/src/api/flow.ts
@@ -4,35 +4,9 @@ let generatorId = 0
 
 export type CancellablePromise<T> = Promise<T> & { cancel(): void }
 
-export interface FlowYield {
-    // fake, only for typing
-    "!!flowYield": undefined
-}
-
-export interface FlowReturn<T> {
-    // fake, only for typing
-    "!!flowReturn": T
-}
-
-// we skip promises that are the result of yielding promises (except if they use flowReturn)
-export type FlowReturnType<R> = IfAllAreFlowYieldThenVoid<
-    R extends FlowReturn<infer FR>
-        ? FR extends Promise<infer FRP>
-            ? FRP
-            : FR
-        : R extends Promise<any>
-        ? FlowYield
-        : R
->
-
-// we extract yielded promises from the return type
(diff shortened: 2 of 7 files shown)


FredyC:
Hm, breaking change is rather problematic in the current state of things. We could release V6 while still keeping V4 (without improved types) considering that LTS is about bugfixes and not new features. 

We should probably keep PR like these open for now and put them aside to collect more breaking changes to be worth the next major. I don't think that many people are upgrading to the latest TS that fast.

mweststrate:
I'm not entirely comfortable about this change, like for example the introduction of `_await`, and I'm wondering whether we wouldn't be better of by teach people to explicitly pass a a generic arg to flow? E.g. `flow<R>(function*() { untyped stuff })` until this is fixed on TS level?

xaviergonz:
The problem is mostly with yield, which is always typed to return any no matter what, while yield* transfers control to another generator and returns the proper return type.

In theory _await (or whatever name is chosen) could be avoided, but that'd mean to polyfill Promise.prototype[Symbol.iterator()] (which is currently undefined) and add a global declaration augmentation for it

```
declare global {
  interface Promise<T> {
    [Symbol.iterator](): Generator<Promise<T>, T, unknown>
  }
}
```

As for backwards compatibility, I think if we added a ts3.6 folder with files that'd need to be overridden (e.g. flow.ts, internal.ts and mobx.ts) and there was an extra build with those files overridden with ts3.6, with output to a folder such as dist/ts3.6, afterwards removing js files from that folder, and then the package json had something like
```
"types": "./index.d.ts",
  "typesVersions": {
    ">=3.6": { "*": ["ts3.6/*"] }
  }
```
Then there could be different declarations for both ts versions.

mweststrate:
I think this one can be closed in favor of https://github.com/mobxjs/mobx-utils/pull/217. Let me know if I'm mistaken

PR #2097, opened 2019-09-05
Fix for issue 2096: unintuitive behavior when an observable is created and then observed within the same computed

### LINKED ISSUE #2096: unintuitive behavior when an observable is created and then observed within the same computed
I have a:

2. [x] Issue:
  * [x] Are you willing to (attempt) a PR yourself?

Within a computed function you're allowed to create an observable and then modify it. This makes perfect sense and was resolved in issue #563. Yet if you then proceed to read from the observable you just created within the computed function, it will be observed. Which can lead to some surprising and unintuitive behavior. 

Consider the following example:

```javascript
class Foo {
  @computed
  get store() {
    return new Store({ name: 'foo' });
  }
}
class Store {
 @observable obj;
 @observable name;

 constructor(obj) {
  this.obj = obj;
  this.name = obj.name;
 }


}

const f = new Foo();
autorun(() => f.store);
const g = f.store;
console.log(g === f.store)
g.obj.name = 'new name';
console.log(g === f.store)
```

This will log 
```javascript
true
true
```

but within the constructor if you change  `this.name = obj.name;` to `this.name = this.obj.name` you would expect the same behavior but instead you will now get:

```javascript
true
false
```

since `this.obj` is now observed by the computed even though the `this` was created within that same computed. 

The proposal here would be to make created observables completely silent for both read and write operations to the computed function that created them. 

This will avoid the behaviour outlined above but also avoid a needless re-computation since changing an observable produced by a computed will not change the value that the computed produces.

Example:

```javascript
let o1;
const c = computed(() => {
  const o2 = observable.box(2);

  o2.set(o2.get() * 2);

  return o2;
});

autorun(() => {
   o1 = c.get();
  console.log(o1.get())
})
// logs 4
o1.set(10)
// re-evaluates but still logs 4
```
It is possible to produce a different value but that would involve modifying outside state within the computed making it an anti-pattern.

melnikov-s:
* [x ] Added unit tests
* [x ] Verified that there is no significant performance drop (`npm run perf`)

Fixes #2096 

Not sure if this is the best approach as I don't have much experience with the mobx code base but it seemed to be the path of least resistance.

### DIFF
--- a/src/core/atom.ts
+++ b/src/core/atom.ts
@@ -10,7 +10,8 @@ import {
     onBecomeUnobserved,
     propagateChanged,
     reportObserved,
-    startBatch
+    startBatch,
+    globalState
 } from "../internal"
 import { Lambda } from "../utils/utils"
 
@@ -26,6 +27,7 @@ export class Atom implements IAtom {
     isBeingObserved = false
     observers = new Set<IDerivation>()
 
+    computedCreatedIn: number | undefined = undefined
     diffValue = 0
     lastAccessedBy = 0
     lowestObserverState = IDerivationState.NOT_TRACKING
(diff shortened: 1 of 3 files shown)


urugator:
Closing due to the reasons explained in #2096

Issue #2070, opened 2019-08-08
Reserved keys `global`, `self` in appx of alipay


doxiaodong:
When I run mobx in appx of alipay，it's thow error caused by `self` in https://github.com/mobxjs/mobx/blob/f944b5413f2f4b9de89e00e7de1ddde2a178e19d/src/utils/iterable.ts#L6
and `global` in https://github.com/mobxjs/mobx/blob/6f095ed0330d5ee2433fc87359ebd44ded630e26/src/core/globalstate.ts#L183 

in Alipay appx, the `global` , `self`  cannot be used and defined.
![image](https://user-images.githubusercontent.com/6927481/62717132-ec2e4880-ba35-11e9-9c85-39bb660357af.png)

It can be fixed simply by
```
function _self() {
  return this;
}
```
and
```
const mockGlobal = {};
function getGlobal() {
  if (ypeof window !== "undefined") {
    return window;
  }
  if (typeof global !== "undefined") {
    return global;
  }
  return mockGlobal;
}
```

finally, it's better to update both v4 and v5

FredyC:
That's some really awkward platform having a problem with something like that :) I suppose it cannot hurt anything, so if you are willing to provide PR for both versions, it will likely be merged.

PR #2104, opened 2019-09-11
Use convergent global

doxiaodong:
supplement of https://github.com/mobxjs/mobx/issues/2070


FredyC:
Fairly innocent change, but not sure why a test is failing 🤔

PR #2117, opened 2019-09-21
expose startAction and endAction

xaviergonz:
* [X] Updated changelog

Basically adds a new `startActionWithFinisher` method that allows someone to start/finish an action without needing a code block.

This is mostly useful for library creators, and probably for a future async pattern I'm investigating that doesn't rely on generators in order to get better typings in TS.

Basically on this pattern, an asyncAction would go something like this:

```ts
const myAsyncAction = asyncAction("name", async(x) => {
  // the wrapper starts the action using startActionWithFinisher
  // currentActionContextFinisher = startWithActionFinisher(...)

  // some mutating actions

  const data = await _(getDataFromBackend(x))

  // more mutating actions

  // the wrapper try/catches the whole async block, runing currentActionContextFinisher(true) if
  // throwing, (false) if not
}

// decorator version
@asyncAction
async name(x) {
  // ...
}

async function _(whateverPromise) {
  currentActionContextFinisher(false)
  currentActionContextFinisher = undefined
  try {
    const ret = await whateverPromise
  } finally {
    currentActionContextFinisher = startWithActionFinisher(...)
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Don't use `global` and `self` keywords unless defined. Fixes [#2070](https://github.com/mobxjs/mobx/issues/2070).
 * onBecome(Un)Observed didn't trigger when using number as key of observable map. Fixes [#2067](https://github.com/mobxjs/mobx/issues/2067).
-* Exposed `startAction` and `endAction` to be able to start and action and finish it without needing a code block. This is low level stuff you shouldn't need that's mostly useful for library creators.
+* Exposed `_startAction` and `_endAction` to be able to start and action and finish it without needing a code block. This is low level stuff you shouldn't need that's mostly useful for library creators.
 
 # 5.13.0 / 4.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Don't use `global` and `self` keywords unless defined. Fixes [#2070](https://github.com/mobxjs/mobx/issues/2070).
 * onBecome(Un)Observed didn't trigger when using number as key of observable map. Fixes [#2067](https://github.com/mobxjs/mobx/issues/2067).
-* Added `startActionWithFinisher` to be able to start and action and finish it without needing a code block. Useful for library creators.
+* Exposed `startAction` and `endAction` to be able to start and action and finish it without needing a code block. Useful for library creators.
 
 # 5.13.0 / 4.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Don't use `global` and `self` keywords unless defined. Fixes [#2070](https://github.com/mobxjs/mobx/issues/2070).
 * onBecome(Un)Observed didn't trigger when using number as key of observable map. Fixes [#2067](https://github.com/mobxjs/mobx/issues/2067).
-* Exposed `startAction` and `endAction` to be able to start and action and finish it without needing a code block. Useful for library creators.
+* Exposed `startAction` and `endAction` to be able to start and action and finish it without needing a code block. This is low level stuff you shouldn't need that's mostly useful for library creators.
 
 # 5.13.0 / 4.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Don't use `global` and `self` keywords unless defined. Fixes [#2070](https://github.com/mobxjs/mobx/issues/2070).
 * onBecome(Un)Observed didn't trigger when using number as key of observable map. Fixes [#2067](https://github.com/mobxjs/mobx/issues/2067).
-
+* Added `startActionWithFinisher` to be able to start and action and finish it without needing a code block. Useful for library creators.
 
 # 5.13.0 / 4.13.0
 
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -30,14 +30,14 @@ export function createAction(actionName: string, fn: Function, ref?: Object): Fu
 }
 
 export function executeAction(actionName: string, fn: Function, scope?: any, args?: IArguments) {
-    const runInfo = startAction(actionName, scope, args)
+    const runInfo = _startAction(actionName, scope, args)
     try {
         return fn.apply(scope, args)
     } catch (err) {
         runInfo.error = err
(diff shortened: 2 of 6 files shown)


urugator:
Afaik this isn't possible because even the first line of `async` function is already deferred, therefore the `action` may be in whatever state, irrelevantly whether it was started/ended just before invoking async function:
```javascript
function asyncAction(asyncFn) {
  currentActionContextFinisher = startWithActionFinisher(...);
  asyncFn(); // during the execution of asyncFn the action may or may not be started and it can be completely different action than the one started at previous line
}
```
`_` has similar problem, but it could be rewritten to non-async function except the `finally` part.

xaviergonz:
> Afaik this isn't possible because even the first line of `async` function is already deferred, therefore the `action` may be in whatever state, irrelevantly whether it was started/ended just before invoking async function:
> 

In this PR you can see how it could be used:
https://github.com/mobxjs/mobx/pull/2118

urugator:
Nevermind... I was fairly convinced that async functions are immediately deferred...basically translated into:
```javascript
function asyncFn() {   
  return Promise.resolve().then(() => {
    console.log("async function body");
  });                    
}
```
But that doesn't seem to be the case, they run immediately until the first `await`.
Dunno if that was a just a bug in some implementation or how did I get to that conclusion ... I knew it had some quite impractical implications, thats why I remebered that ... really surprised and confused :D

doxiaodong:
This will need change all async code with state change to `_()`. But I think it's better to create a new repo rather than mobx

urugator:
Could we move the `threw` flag into `runInfo` and handle `globalState.suppressReactionErrors` inside `endAction` (exposing it instead of the new method + avoiding code duplication of `finally` block?):

```javascript
export function executeAction(actionName: string, fn: Function, scope?: any, args?: IArguments) {
    const runInfo = startAction(actionName, scope, args)
    let shouldSupressReactionError = true
    try {
      return fn.apply(scope, args)        
    } catch {
      runInfo.threw = true;    
    } finally {
       endAction(runInfo)       
    }
}

function endAction(runInfo: IActionRunInfo) {
  if (runInfo.threw) {
    globalState.suppressReactionErrors = true;
  }
  allowStateChangesEnd(runInfo.prevAllowStateChanges)
  endBatch()
  untrackedEnd(runInfo.prevDerivation)
  if (runInfo.notifySpy && process.env.NODE_ENV !== "production")
      spyReportEnd({ time: Date.now() - runInfo.startTime })
  globalState.suppressReactionErrors = false;    
}
```

urugator:
We could also consider some `runInfo.actionId` + `globalState.currentActionId` to guard against calling `start/endAction` in improper order.

mweststrate:
Could we make this first as a mobx-utils feature and see how it is
appreciated? I noticed that the amount of ways doing async is already
confusing to many people, so I'd be a bit hesitant to add yet another
solution that has it's own specific 'syntactical' requirements, such as
flow, or runInAction, to that.

urugator:
@mweststrate This issue is basically about exposing `startAction` / `endAction`, so that it's even possible to implement.

mweststrate:
jeez, commenting in the wrong thread, sorry.

xaviergonz:
> Could we move the `threw` flag into `runInfo` and handle `globalState.suppressReactionErrors` inside `endAction` (exposing it instead of the new method + avoiding code duplication of `finally` block?):
> 
> ```js
> export function executeAction(actionName: string, fn: Function, scope?: any, args?: IArguments) {
>     const runInfo = startAction(actionName, scope, args)
>     try {
>       return fn.apply(scope, args)        
>     } catch {
>       runInfo.threw = true;    
>     } finally {
>        endAction(runInfo)       
>     }
> }
> 
> function endAction(runInfo: IActionRunInfo) {
>   if (runInfo.threw) {
>     globalState.suppressReactionErrors = true;
>   }
>   allowStateChangesEnd(runInfo.prevAllowStateChanges)
>   endBatch()
>   untrackedEnd(runInfo.prevDerivation)
>   if (runInfo.notifySpy && process.env.NODE_ENV !== "production")
>       spyReportEnd({ time: Date.now() - runInfo.startTime })
>   globalState.suppressReactionErrors = false;    
> }
> ```

Done, though I didn't move `threw = true` inside a catch block since I imagine it was done that way to preserve the stack trace when "rethrowing".

> We could also consider some runInfo.actionId + globalState.currentActionId to guard against calling start/endAction in improper order.

Done through an stack of runInfos (which only gets used/verified in debug mode).

urugator:
I hoped that you will remove the `startActionWithFinisher` and exposes `startAction`/`endAction`.
Also now it occured to me it could be a useful to replace the `runInfo.threw = true` by `runInfo.error = error` (error being the catched error)

xaviergonz:
> I hoped that you will remove the `startActionWithFinisher` and expose `startAction`/`endAction`.

Ahh, I can do that :)

> Also now it occured to me it could be useful to replace the `runInfo.threw = true` by `runInfo.error = error` (error being the catched error)

I could do that, but to do it I'd need to catch and rethrow the error, which I think would modify the stack trace?

xaviergonz:
Either way, both comments addressed and done :)

urugator:
Gave it quick shot in codesandbox and unless I am missing something, rethrowing doesn't seem to alter the stack strace...

urugator:
One more idea ... is there a way to check that the `endAction` is called "synchronously"? Something like this?
```javascript
function startAction() {
   Promise.resolve.then(() => {
       console.log('If you haven't called endAction to this point, you messed up.')
   })
}
```

xaviergonz:
I guess there is, but I'm scare to create a promise for every action even in dev mode 😨

mweststrate:
Would it be an idea to prefix the names with `_`, like we did with others, to signal more clearly that it is not supposed to be used in normal scenarios and might change if the internal implementation changes?

mweststrate:
Reaally NIT: the unmatched close could be made a bit more efficient by just giving every new runInfo an increasing counter, and a global counter to compare against.

xaviergonz:
> Would it be an idea to prefix the names with _, like we did with others, to signal more clearly that it is not supposed to be used in normal scenarios and might change if the internal implementation changes?

done

> Reaally NIT: the unmatched close could be made a bit more efficient by just giving every new runInfo an increasing counter, and a global counter to compare against.

done

PR #2118, opened 2019-09-21
asyncAction

xaviergonz:
Related to PR https://github.com/mobxjs/mobx/pull/2117

Adds `asyncAction` as an alternative to flows in order to have better TS typings

usage:

```ts
const myAsyncAction = asyncAction("name", async(_: AsyncActionAwaiter, x: string) => {
  // mutable ops
  const data = await _(() => getBackendData(x))
  // more mutable ops
  return data.whatever
})

const res = await myAsyncAtion("hello")
```

The key here is that anything executed inside the awaiter (`_`) (in this case getBackendData) will be run "out of action", and anything after it back in an action.

It is ok to run inner asyncActions within asyncActions as long as the awaiter wrapper is used.

Summarizing, the only rule is to do `await _(() => whatever)` instead of `await whatever` inside an asyncAction for this to work

The above gets pseudo transformed into:
```ts
// start actionName step 0
// mutable ops
// end actionName step 0
await getBackendData
// start actionName step 1
// more mutable ops
// end actionName step 1
```


xaviergonz:
I just realized that it actually should be `_(() => `getBackendData(...))` so the first part of the promise doesn't get inside the action

xaviergonz:
Fixed and edited the top description

mweststrate:
Could we make this first as a mobx-utils feature and see how it is
appreciated? I noticed that the amount of ways doing async is already
confusing to many people, so I'd be a bit hesitant to add yet another
solution that has it's own specific 'syntactical' requirements, such as
flow, or runInAction, to that.

I think btw we should at least come up with a standarized name for `_`, as that is meaningless to the unaware reader. (this is clearly the hard part :-P. `cont`, `run`, `task`, `runTask`, something like that?)

xaviergonz:
Sounds good. Does the name matter that much in this case? It is a parameter so the implementers can actually name it whatever he wants

xaviergonz:
I mean, this is also valid code

```ts
const myAsyncAction = asyncAction("name", async(somewhereOverTheRainbow: AsyncActionAwaiter, x: string) => {
  // mutable ops
  const data = await somewhereOverTheRainbow(() => getBackendData(x))
  // more mutable ops
  return data.whatever
})

const res = await myAsyncAtion("hello")
```

xaviergonz:
Btw, I'll create the PR for mobx-utils once/if https://github.com/mobxjs/mobx/pull/2117 is merged, since it is a pre-requisite :)

mweststrate:
Yeah I know it doesn't matter technically, but it is good to guide users to
a standardized name, so that their code will be readable for their peers
once used to the pattern, like 'self' in mst, or 'draft' in immer.

Op di 24 sep. 2019 17:35 schreef Javier Gonzalez <notifications@github.com>:

urugator:
> `_(() => getBackendData(...))` so the first part of the promise doesn't get inside the action

Why is that a problem?

xaviergonz:
@urugator 
Hmm, I'm not 100% sure if it should or it shouldn't.

Is it better to make the first chunk of the "external" function in action while the other steps will be not (assuming it is not a flow / asyncAction)?
- Pro: just need to write `await _(getBackendData(...))`
- Cons: you will be confused when stuff after the first await inside that function is out of action

Or just ensure that external function will be run out of action?
- Pro: might be more obvious that the whole external thing is out of action
- Cons: syntax: `await _(() => getBackendData(...))`, if the external action is a flow/async action I think it won't be batched as part of the root action

@mweststrate 
task sounds short and nice to me :)

xaviergonz:
Closing in favor of mobx-utils one: https://github.com/mobxjs/mobx-utils/pull/217

mweststrate:
Released as mobx-utils 5.5.0: https://github.com/mobxjs/mobx-utils#actionasync. Thanks!

PR #2124, opened 2019-09-24
expose startAction and endAction [mobx4]

xaviergonz:
Backport of https://github.com/mobxjs/mobx/pull/2117 for mobx4

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -29,40 +29,27 @@ export function createAction(actionName: string, fn: Function): Function & IActi
     return res as any
 }
 
-export function startActionWithFinisher(
-    actionName: string,
-    scope?: any,
-    args?: IArguments
-): (threw: boolean) => void {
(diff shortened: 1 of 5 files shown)


xaviergonz:
The only thing that changed was from
`if (runInfo.notifySpy)`
to 
` if (runInfo.notifySpy && process.env.NODE_ENV !== "production") `

urugator:
All spies are disabled on production since V5 correct?

xaviergonz:
🤷‍♂ to be honest no idea, should I take that one back?

urugator:
https://github.com/mobxjs/mobx/blob/master/CHANGELOG.md#breaking-changes
3rd point
So it's fine I think...

urugator:
oh w8 it's the other way around...

urugator:
yea I think the production cond shouldn't be there

xaviergonz:
fixed

PR #2131, opened 2019-09-30
define the name property of function return by createAction

dbenson24:
currently all functions returned by createAction are named res, which is fine but isn't helpful. I'd like to define the name of the function to be the actionName.

PR checklist:

* [ ] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -27,8 +27,11 @@ export function createAction(actionName: string, fn: Function, ref?: Object): Fu
         return executeAction(actionName, fn, ref || this, arguments)
     }
     ;(res as any).isMobxAction = true
-    if (process.env.NODE_ENV !== "production" && Object.getOwnPropertyDescriptor(res, "name").configurable) {
-        Object.defineProperty(res, "name", {value: actionName});
+    if (process.env.NODE_ENV !== "production") {
+        const descriptor = Object.getOwnPropertyDescriptor(res, "name");
+        if (descriptor && descriptor.configurable) {
+            Object.defineProperty(res, "name", {value: actionName});
+        }
     }
     return res as any
 }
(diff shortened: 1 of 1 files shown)


mweststrate:
Hey, thanks for this patch and idea! 

Please note that function names are not always configurable (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name for details), so that would need to be checked on a per-browser basis first. 

Secondly, I think it would be great if this would be done only in development mode.

dbenson24:
Restricted the feature to development only, I'm checking to make sure that the descriptor is exists and is configurable before I set a new one. I'm not sure if this is enough, do you have any examples of other browser specific code in the mobx codebase?

src/core/action.ts
```diff
@@ -27,6 +27,12 @@ export function createAction(actionName: string, fn: Function, ref?: Object): Fu
         return executeAction(actionName, fn, ref || this, arguments)
     }
     ;(res as any).isMobxAction = true
+    if (process.env.NODE_ENV !== "production") {
```
mweststrate: Implementing it this way will run the check every time an action is created, however, since this code might be hit hunderds of thousands times, and the result only depends on the environment, you only need to run it once. I suggest to pull this code up, run the check only once during module initialization, and reuse the result.

mweststrate: (either `name` is configurable on all Functions, or none)

mweststrate: also, could you add a unit test verifying this behavior?

FredyC:
@dbenson24 Are you willing to continue with this PR?

nadavkaner:
Hey, can I continue with this? I opened a PR with the changes for this https://github.com/mobxjs/mobx/pull/2262

FredyC:
@nadavkaner definitely :)

PR #2136, opened 2019-10-04
Fix for #2135

### LINKED ISSUE #2135: Why am I getting an `invalid action stack` error?
Context:
I am trying to run some jest/react-testing-library tests with drone and get this error when the test tries to render my component.
`Error message: "[mobx] invalid action stack. did you forget to finish an action?"`
My component renders some data that is being loaded into a store via mobx. Every time this store is created, the `load` action is called inside the store's constructor. Every time this `load` action is called, the test fails. However, there are no errors that arise when I run my app and no errors that arise when I run my tests locally. This error only occurs when I run my tests in drone.

This error seems to originate from here:
https://github.com/mobxjs/mobx/blob/master/src/core/action.ts#L88
Which uses mobx globals which are defined here: https://github.com/mobxjs/mobx/blob/master/src/core/globalstate.ts#L157

1. Looks like there is global state stacking for actions that are being run (and when they finish)
2. Also looks like there are problems with the global state object if there are multiple versions on mobx active

Question:
What are some things I could try that could fix this issue? I tried looking on stackoverflow and had little luck finding others who had run into this issue :(

Hopefully someone can help me here!
Thank you in advance!

xaviergonz:
Rather than using the global state for the action stack checks (which might be uninitialized in older mobx versions) it will just rely on local state.
In theory this means the check will be less effective (since it will only check the stack of actions called using the same mobx version are ok), but less likely to break in multiple mobx version scenarios.

Should fix #2135


xaviergonz:
@mweststrate should this approach be used or the globlal state versioning one?

mweststrate:
I think this is ok

thecodejack:
@mweststrate can we please have a patch release of this?

mweststrate:
Released 4/5.14.1

PR #2138, opened 2019-10-07
Fix dead links in docs about Action

jdim:
Hi, it seems that there are some dead links in the documentation about `action`.

Please check out these pages
https://mobx.js.org/refguide/action.html#action
https://mobx.js.org/refguide/action.html#when-to-use-actions

So, I would like to fix this.


FredyC:
Thanks a lot. We have migrated to Docusaurus just recently, so it's expected there will be some slight issues. Keep fixes coming :)

PR #2144, opened 2019-10-09
Remove obsolete donation info

FredyC:
The obsolete section about card donations leading to a non-existing page. Since OpenCollective is preferred, let's keep it less confusing :)

@mweststrate I also wonder, those list of backers/sponsors is generated somehow? If not, there might be some people missing, right?


mweststrate:
Correct, the paypal sponsor section can be removed, doesn't make much sense anymore. For the OC sponsors, it normally doesn't need updating, as there is a large set of placeholder images, that position wise show one sponsor if it exists, and otherwise nothing. If there are sponsors missing, it is just a matter of copying more placeholders and increasing there numbers.

The top sponsor section is manually maintained though, as they mostly display active sponsors. Which is something that OC doesnt support. (and if it did, it wouldn't be accurate as some prefer to pay lump sum for an entire year instead of monthly)

PR #2153, opened 2019-10-13
Allow action to handle async functions

reaktivo:
This PR wraps each "step" of an Async function in its own `action`. This removes the need for `flow`, rewriting async functions as generators and most uses of `runInAction` when used in the context of an async function. 

In the following example, the both `data` and `loading` are run in a single action, removing the need for extra `runInAction` calls.

Of course, this PR is simply a proof-of-concept and there's quite a few other things to consider:
- Is there a drastic performance impact in having each async resolver wrapped in an action, even when there's possibly no observable being mutated?
- If this would ever be merged and supported in MobX, would async functions that already in include `runInAction` calls cause any issues?
- Last but not least, is this the correct place to have the resolver's being hooked into?

What do you think?

Instead of
```js
const store = observable({ loading: false, data: {} });
const someAsyncWork = action(async () => {
  store.loading = true;
  const data = await fetchSomething();
  runInAction(() => {
    store.data = data;
    store.loading = false;
  });
});
```

you can simply write:

```js
const store = observable({ loading: false, data: {} });
const someAsyncWork = action(async () => {
  store.loading = true;
  const data = await fetchSomething();
  store.data = data;
  store.loading = false;
});
```

PR checklist:

* [x] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -38,7 +38,7 @@ export function createAction(actionName: string, fn: Function, ref?: Object): Fu
 export function executeAction(actionName: string, fn: Function, scope?: any, args?: IArguments) {
     const runInfo = _startAction(actionName, scope, args)
     try {
-        return fn.apply(scope, args)
+        return _hookIntoAsyncResolvers(actionName, scope, () => fn.apply(scope, args))
     } catch (err) {
         runInfo.error = err
         throw err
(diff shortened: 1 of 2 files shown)


FredyC:
Recently @xaviergonz introduced asyncAsync (#2118), perhaps that covers your concern? I haven't investigated it yet, so I am not sure.

xaviergonz:
It would actually be this PR https://github.com/mobxjs/mobx-utils/pull/217

I also tried to make it magically work by tweaking promises (and actually got quite far) in this other PR https://github.com/mobxjs/mobx-utils/pull/217 but in the end native (non transpiled) async/await bit me.
Using native async/await in Chrome would not call the overridden promise constructor (a first approach I had), and in node it wouldn't call overridden "then", "catch", etc. :(

reaktivo:
@xaviergonz You're absolutely right. I was misled by the transpiled output and the tests, which worked seemingly well. Seems there's no way to hook into `await`s at all. Thanks for having a look. Closing this one.

PR #2154, opened 2019-10-14
Update README.md

codemilli:
fix a broken link


codemilli:
@FredyC applied

FredyC:
Thanks. Now thinking about it, the link should probably point directly to https://mobx.js.org/best/react.html instead of github source. Not sure why it was like that before.

FredyC:
Apply to both READMEs please 🙈

FredyC:
Thanks a lot!

PR #2157, opened 2019-10-15
Fix error when interception is removed in intercept (#1950)

### LINKED ISSUE #1950: TypeError when an interception is removed in intercept
I have created an interception using the `intercept` function and called another function inside, which removes another interception on the same locale. This looks something like this:

```javascript
intercept(this.locale, '', (change) => {
    this.listStore.destroy();
    return change;
});
```

The `destroy` method of the `listStore` removes an interception on the `this.locale` reference. And this seems to happen during MobX iterates over all the interceptions, which finally results in the following error:

```
Uncaught TypeError: interceptors[i] is not a function
    at interceptChange$$1 (mobx.module.js?8b18:2802)
    at ObservableValue$$1.prepareNewValue (mobx.module.js?8b18:774)
    at ObservableValue$$1.set (mobx.module.js?8b18:755)
    at ResourceStore.setLocale (ResourceStore.js?1b9b:108)
    at executeAction$$1 (mobx.module.js?8b18:649)
    at ResourceStore.res (mobx.module.js?8b18:640)
    at Object.onChange (MediaDetails.js?013a:193)
    at Object.Select._this.handleOptionClick [as onOptionClick] (Select.js?f6a9:56)
    at OptionList._this.handleOptionClick (OptionList.js?7462:23)
    at Option._this.handleOnClick (Option.js?f8b9:24)
```

This makes totally sense, once you got your head around it, but it took me some time and debugging to get there. I know that removing stuff from a collection while iterating over it is a very common problem, and could totally understand if you say that it should not be supported. However, the error was not very helpful, so I would at least expect an error saying that disposing interceptions in an interception is not supported.

Or, if there is no reason to not support it, I found this [stackoverflow answer](https://stackoverflow.com/a/16352560/1292378) describing a solution. I could also imagine to provide a PR implementing that, since it does not seem a change that hard.

fgerschau:
Currently, when disposing an interceptor within the provided callback, mobx throws an error stating that `interceptors[i] is not a function`. With this PR it will continue the execution by cloning the interceptors array before the execution loop.
Fixes #1950 

PR checklist:

* [x] Added unit tests
* [x] Updated changelog

### DIFF
--- a/src/types/intercept-utils.ts
+++ b/src/types/intercept-utils.ts
@@ -29,16 +29,15 @@ export function interceptChange<T>(
 ): T | null {
     const prevU = untrackedStart()
     try {
-        const interceptors = interceptable.interceptors
-        if (interceptors)
-            for (let i = 0, l = interceptors.length; i < l; i++) {
-                change = interceptors[i](change)
-                invariant(
-                    !change || (change as any).type,
-                    "Intercept handlers should return nothing or a change object"
-                )
-                if (!change) break
-            }
+        const interceptors = [...(interceptable.interceptors || [])]
+        for (let i = 0, l = interceptors.length; i < l; i++) {
+            change = interceptors[i](change)
+            invariant(
+                !change || (change as any).type,
+                "Intercept handlers should return nothing or a change object"
+            )
+            if (!change) break
+        }
         return change
     } finally {
         untrackedEnd(prevU)
--- a/src/types/intercept-utils.ts
+++ b/src/types/intercept-utils.ts
@@ -29,6 +29,7 @@ export function interceptChange<T>(
 ): T | null {
     const prevU = untrackedStart()
     try {
+        // Interceptor can modify the array, copy it to avoid concurrent modification, see #1950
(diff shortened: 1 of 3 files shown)


FredyC:
Thanks for the contribution with proper changelog update and test, it's rare to see those :)

Someone who has insight into these things will take a look soon.

src/types/intercept-utils.ts
```diff
@@ -29,16 +29,15 @@ export function interceptChange<T>(
...
-        if (interceptors)
-            for (let i = 0, l = interceptors.length; i < l; i++) {
-                change = interceptors[i](change)
-                invariant(
-                    !change || (change as any).type,
-                    "Intercept handlers should return nothing or a change object"
-                )
-                if (!change) break
-            }
+        const interceptors = [...(interceptable.interceptors || [])]
```
urugator: Add comment 
```javascript
// Interceptor can modify the array, copy it to avoid concurrent modification, see #1950
const interceptors = [...(interceptable.interceptors || [])]
```

PR #2160, opened 2019-10-17
Fix flow typings [requires TS3.6]

xaviergonz:
Fixes typings for flows (actually even improves them), but requires TS3.6 to use flows.

If you don't want to merge it now maybe just consider leaving it there for the future.


mweststrate:
I think we could release this as minor pretty soon, once we didn't have any patches for a couple of weeks

xaviergonz:
Also fixed async generators while at it

xaviergonz:
One question, there's a way to enforce TS typings so flow() has to always yield promises, would this be nice to have?

mweststrate:
Published as 5.15.0

PR #2161, opened 2019-10-17
Fix flow typings mobx4 [requires TS3.6, mobx4]

xaviergonz:
Like https://github.com/mobxjs/mobx/pull/2160 for mobx4


src/api/flow.ts
```diff
@@ -4,35 +4,9 @@ let generatorId = 0
...
-
-// we extract yielded promises from the return type
-export type IfAllAreFlowYieldThenVoid<R> = Exclude<R, FlowYield> extends never
-    ? void
-    : Exclude<R, FlowYield>
-
 export function flow<R, Args extends any[]>(
-    generator: (...args: Args) => IterableIterator<R>
-): (...args: Args) => CancellablePromise<FlowReturnType<R>> {
+    generator: (...args: Args) => Generator<any, R, any>
```
mweststrate: In the mobx5 branch a union of Generator and Async generator was returned. What is the reason (or impact) of the difference?

xaviergonz: Support for async generators, forgot to bsckport that one 😨

xaviergonz: Should be fixed now

PR #2166, opened 2019-10-18
Fix 2165 use prepare

### LINKED ISSUE #2165: Should be using prepare rather than postinstall step for website dependency install
In the latest Mobx 5.14.1, Installing mobx as a dependency is resulting in the `postinstall` script executing which is trying to install the dependencies of the website folder which is not distributed in the package resulting in errors like this during dependency installation:

```bash
$ yarn --cwd website install
Killed
yarn install v1.16.0
error Directory "/home/circleci/project/node_modules/mobx/website" doesn't exist
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
error Command failed with exit code 1.
```

Reproduction:
* install as a dependency of any project.

Easy fix is to change `postinstall` to `prepare`. `prepare` will execute only on local installation, and not during installation as a dependency.

benjamind:
Changes `postinstall` to `prepare` for website dependency install. Bumps to 5.14.2 so this can be republished, and added changelog entry.

* [x] Updated changelog

Fixes #2165

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "5.14.1",
+  "version": "5.14.2",
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
   "umd:main": "lib/mobx.umd.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "5.14.2",
+  "version": "5.14.1",
   "description": "Simple, scalable state management.",
   "main": "lib/mobx.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mobx",
-  "version": "5.14.1",
```
FredyC: Uh, please remove the bump, it's done automatically in the release process.

benjamind: Apologies, fixed.

FredyC:
Ok, I've tested it out locally and seems to work correctly.

PR #2175, opened 2019-10-23
Observable decorator on computed symbol keys

### LINKED ISSUE #2159: Decorators make properties readonly if their key is a symbol computed property in TS (phew)
Ok, that's a mouthful. Here's the situation in question:
```
const someSymbol = Symbol('whatever');

class MyClass {
  @observable
  [someSymbol] = true;

  @action.bound
  changeSymbolProperty() {
    this[someSymbol] = false;
  }
}

const class = new Class();

class.changeSymbolProperty()
// This triggers an error: TypeError: Cannot assign to read only property [someSymbol]
```

Details:
1. MobX version 5.14.0
2. TS version 3.6.4
3. From my understanding, TypeScript's decorators implementation does not provide any default meta data. So if a decorator doesn't specify writable, it falls back to the default `writable = false;`
4. For some reason this isn't an issue with normal property keys, or computed property keys that use strings. It only happens with symbols.
5. You have to use TypeScript to compile the example to get this error. If you use Babel, you get a syntax error (unexpected token on the equals sign in `[someSymbol] = true` – a bug in babel with decorators on computed property keys, unrelated to this problem.

Since it doesn't make sense that you'd want to observe something that's readonly, can we add an explicit `writable = true` to the observable decorator?

Is MobX even responsible for caring about how TypeScript implements decorators? I guess that's the downside of "experimental syntax". 

---

So, the fallback is to use `decorate` on these symbol computed property keys, which works fine. Most of the time.

`decorate` gets confused if you're creating an abstract class. It doesn't matter if the property is a string, symbol, computed, or not. This might be a separate issue, but here's an example:

```
abstract class MyClass {
  abstract someAbstractProperty = 'whatever';

  localProperty = 'some value';
}

decorate(MyClass, {
  localProperty: observable
});
// Error: Object literal may only specify known properties, and 'localProperty' does not exist in type '{ prototype?: MethodDecorator | PropertyDecorator | MethodDecorator[] | PropertyDecorator[] | undefined; }'
```
I'm guessing abstract class messes up that `prototype` typing.

---

What's it all mean? Two things:

First, @observable turns properties `readonly` in this situation:
1. Computed property key
2. Key is a symbol
3. When using TypeScript's decorator implementation (babel's fails too but it's a syntax error bug in babel)

Two, I don't think there's a way to make this situation observable:
1. Computed property key (string or symbol)
2. In an Abstract class

StephenHaney:
Resolves #2159 

1. I added a new test in `typescript-tests` that tries to use computed property keys... see test `#2159 - computed property keys`. Please validate this test is a supported use case (I think it is). It was failing before this change (see original issue for details). By the way, I verified the test passes if the key isn't a computed symbol to make sure the test itself is true.

2. Changed `decorators.ts` and `observable.ts` to apply their logic to symbol keyed properties in addition to string keyed properties.

PR checklist:

* [X] Added unit tests
* [X] Updated changelog
* [X] ~Updated `/docs`. For new functionality, at least `API.md` should be updated~
I don't think this fix needs any doc updates.
* [X] ~Added typescript typings~
As all changes were made in TS files, no need.
* [X] Verified that there is no significant performance drop (`npm run perf`)
Note: I didn't see `npm run perf` but I tried out `npm run test:performance`
Current develop branch: 66.22 real        67.58 user
This PR: 62.46 real        63.76 user

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 -   Fixed flow typings for Facebook's Flow. A new `CancellablePromise` Flow type is exported.
 
+-   Added support for symbol keys on observable properties.
+
 # 5.14.2
 
 -   Fixed installation issue trying to run `postinstall` hook for a website [#2165](https://github.com/mobxjs/mobx/issues/2165).
--- a/src/api/observable.ts
+++ b/src/api/observable.ts
@@ -79,7 +79,7 @@ function getEnhancerFromOptions(options: CreateObservableOptions): IEnhancer<any
  */
 function createObservable(v: any, arg2?: any, arg3?: any) {
     // @observable someProp;
-    if (typeof arguments[1] === "string") {
+    if (typeof arguments[1] === "string" || typeof arguments[1] === "symbol") {
         return deepDecorator.apply(null, arguments as any)
     }
 
--- a/src/utils/decorators.ts
+++ b/src/utils/decorators.ts
@@ -57,8 +57,10 @@ export function initializeInstance(target: DecoratorTarget) {
     const decorators = target[mobxPendingDecorators]
     if (decorators) {
         addHiddenProp(target, mobxDidRunLazyInitializersSymbol, true)
-        for (let key in decorators) {
-            const d = decorators[key]
+        // Build property key array from both strings and symbols
+        const keys = [...Object.getOwnPropertySymbols(decorators), ...Object.keys(decorators)]
+        for (const key of keys) {
+            const d = decorators[key as any]
             d.propertyCreator(target, d.prop, d.descriptor, d.decoratorTarget, d.decoratorArguments)
         }
     }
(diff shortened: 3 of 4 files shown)


StephenHaney:
Tests are now passing, including the new one for computed symbol property keys. If we think this is a good idea, I'll finish up this PR with changelogs updates, etc.

urugator:
Btw there seems to be quite a lot of functions expecting `string` as a prop key. I suppose changing it to `string | symbol` could help preventing these bugs (as a hint for programmer)?

StephenHaney:
> I suppose changing it to string | symbol could help preventing these bugs

I started going down this path but due to the TS limitation around using symbols as index keys, we'd end up having to cast it back to string everywhere. After TS patches that limitation, I think it'd be a good idea to use the `PropertyKey` type everywhere.

mweststrate:
Published as 5.15.0

PR #2184, opened 2019-10-30
allow mobx to run in web worker

shahata:
currently `getGlobal` is first invoked from an iife which initializes `globalState`, which means `mockGlobal` is not yet initialized.

PR checklist:

* [ ] Added unit tests
* [ ] Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+-   Allow mobx to run in web worker
 -   Fixed flow typings for Facebook's Flow. A new `CancellablePromise` Flow type is exported.
 
 # 5.14.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
--   Allow mobx to run in web worker
+-   Fix running mobx in web worker [#2184](https://github.com/mobxjs/mobx/pull/2184/files)
 -   Fixed flow typings for Facebook's Flow. A new `CancellablePromise` Flow type is exported.
 
 # 5.14.2
--- a/src/core/globalstate.ts
+++ b/src/core/globalstate.ts
@@ -200,6 +200,7 @@ export function resetGlobalState() {
 }
 
 declare const window: any
+declare const self: any
 
 const mockGlobal = {}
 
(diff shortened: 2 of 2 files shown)


src/core/globalstate.ts
```diff
@@ -143,6 +143,7 @@ export class MobXGlobals {
 
 let canMergeGlobalState = true
 let isolateCalled = false
+const mockGlobal = {}
```
FredyC: Why did you move it here? Let's keep it closer to the actual use.

shahata: There's a good reason to move it. I described the issue in the commit message. Currently when `getGlobal` is first invoked, it throws exception in case `mockGlobal` is returned because `mockGlobal` is not initialized yet.

shahata: I can move the whole function upward in order to keep them closer if you prefer

shahata: Just to make it clearer. This is in general the current code:

```js
export let globalState = (function() {
    const global = getGlobal()
    ... global._mobxGlobals ...
})()

const mockGlobal = {}
export function getGlobal() {
  ...
  return mockGlobal
}
```

since `globalState` is a IIFE which is invoked before the line `const mockGlobal = {}` is reached, the `mockGlobal` variable is still not initialized  with the `{}` value, therefore the call to `getGlobal` inside the IIFE will not work as expected.

FredyC: Ah, you are right, missed it's being called in IIFE. Yea, I think moving `getGlobal` on top as well makes sense to keep it collocated.

FredyC:
Also please, make another PR for `mobx4-master`. A cherry-pick should be fine probably.

src/core/globalstate.ts
shahata: done

shahata:
@FredyC I think it is ready.
1) hoisted the whole function to be before the IIFE
2) added changelog
3) created #2185 for applying changes also to `mobx4-master`

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+-   Allow mobx to run in web worker
```
FredyC:
```suggestion
-   Fix running mobx in web worker [#2184](https://github.com/mobxjs/mobx/pull/2184/files)
```

FredyC:
Thanks for contribution

shahata:
@FredyC any estimation when to expect next version to be published?

mweststrate:
Published as 5.15.0

PR #2196, opened 2019-11-08
fix #2195, add test

### LINKED ISSUE #2195: IDerivationState.POSSIBLY_STALE + observableRequiresReaction = warning
<s>I don't have a minimal repro or test case yet</s>

Minimal repro:
```ts
    test("don't warn on reads inside a computed, when reaction updates the observed value", function () {
        try {
            // #916
            mobx.configure({ observableRequiresReaction: true })
            const a = mobx.observable.box(0)
            const b = mobx.observable.box(0)
            const c = mobx.computed(() => a.get() + b.get())
            const values = []

            mobx.autorun(() => {
                values.push(c.get())
                b.set(100)
            })

            const messages = utils.supressConsole(() => {
                a.set(1)
            })

            console.log(messages)

// [warn] [mobx] Observable ComputedValue@4 being read outside a reactive context
            expect(messages.length).toBe(0)
        } finally {
            mobx.configure({ observableRequiresReaction: false })
        }
    })
```

The observableRequiresReaction is triggered  when hitting the following code path:
https://github.com/mobxjs/mobx/blob/master/src/core/derivation.ts#L93-L109 
```ts
        case IDerivationState.POSSIBLY_STALE: {
            const prevUntracked = untrackedStart() // no need for those computeds to be reported, they will be picked up in trackDerivedFunction.
```
Due to the `untrackedStart` call

Related #916

**Possible fix:**
Make untrackedStart/untrackedEnd to have own globalState flag?

Bnaya:
fix #2195, add test

PR checklist:

* [x] Added unit tests
* [x] Updated changelog
<s>* [ ] Updated `/docs`. For new functionality, at least `API.md` should be updated</s>
<s>* [ ] Added typescript typings</s>

### DIFF
--- a/src/core/derivation.ts
+++ b/src/core/derivation.ts
@@ -91,6 +91,8 @@ export function shouldCompute(derivation: IDerivation): boolean {
         case IDerivationState.STALE:
             return true
         case IDerivationState.POSSIBLY_STALE: {
+            // state propagation can occur outside of action/reactive context #2195
+            const prevAllowStateReads = allowStateReadsStart(true)
             const prevUntracked = untrackedStart() // no need for those computeds to be reported, they will be picked up in trackDerivedFunction.
             const obs = derivation.observing,
                 l = obs.length
@@ -113,12 +115,14 @@ export function shouldCompute(derivation: IDerivation): boolean {
                     // invariantShouldCompute(derivation)
                     if ((derivation.dependenciesState as any) === IDerivationState.STALE) {
                         untrackedEnd(prevUntracked)
+                        allowStateReadsEnd(prevAllowStateReads)
                         return true
                     }
                 }
             }
             changeDependenciesStateTo0(derivation)
             untrackedEnd(prevUntracked)
+            allowStateReadsEnd(prevAllowStateReads)
             return false
         }
     }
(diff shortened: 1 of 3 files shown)


FredyC:
Does MobX4 suffer of the same issue? Just reminding so it's not left out.

src/core/derivation.ts
```diff
@@ -91,6 +91,7 @@ export function shouldCompute(derivation: IDerivation): boolean {
         case IDerivationState.STALE:
             return true
         case IDerivationState.POSSIBLY_STALE: {
+            const prevAllowStateReads = allowStateReadsStart(true)
```
urugator: Maybe we could add a brief explanation? Something like:
// state propagation can occur outside of action/reactive context
Not sure about the exact phrasing

urugator: Hm now when I think about it... does the issue persist (without fix) if you modify the test like:
```javascript
const messages = utils.supressConsole(action(() => {
  a.set(1)
}))
```

Bnaya: yes

Bnaya:
> Does MobX4 suffer of the same issue? Just reminding so it's not left out.

I believe yes, i will backport the fix afterward

PR #2198, opened 2019-11-10
Getting Started: Fixed that "get next item" just returns first item!

Venryx:
The "getting started" documentation page has the example code just return the first todo, whenever it logs the "next item". Instead, it should search for the first todo item that is not yet completed.

Screenshot of old behavior (which doesn't make sense, in code or in what is displayed):

### DIFF
--- a/website/static/getting-started.html
+++ b/website/static/getting-started.html
@@ -102,7 +102,8 @@ class TodoStore {
 	report() {
 		if (this.todos.length === 0)
 			return "<none>";
-		return `Next todo: "${this.todos.find(todo => todo.completed === false).task}". ` +
+		const nextTodo = this.todos.find(todo => todo.completed === false);
+		return `Next todo: "${nextTodo ? nextTodo.task : "<none>"}". ` +
 			`Progress: ${this.completedTodosCount}/${this.todos.length}`;
 	}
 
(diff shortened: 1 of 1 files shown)


website/static/getting-started.html
```diff
@@ -102,7 +102,7 @@ <h3>A simple todo store...</h3>
 	report() {
 		if (this.todos.length === 0)
 			return "<none>";
-		return `Next todo: "${this.todos[0].task}". ` +
+		return `Next todo: "${this.todos.find(todo => todo.completed === false).task}". ` +
```
FredyC: `find` will return `undefined` if nothing is found. In this case, it doesn't matter that much, but let's show a proper code here that won't break in less contrived situations.

Venryx: What approach should I use?

My preference:
```
const nextTodo = this.todos.find(todo => todo.completed === false);
return `Next todo: "${nextTodo ? nextTodo.task : "n/a"}". `
```

Is that fine? Or does the code style for this project call for something different?

Some other options:
```
2) return `Next todo: "${(this.todos.find(todo => todo.completed === false) || {}).task || "n/a"}". `
```
```
3) return `Next todo: "${this.todos.filter(todo => todo.completed === false).map(todo => todo.task)[0] || "n/a"}". `
```

FredyC: Yea, I would go with the first one. There is no project preference I am aware off :)

Venryx: Updated the PR with the fix. (the first option -- though with "<none>" as text instead of "n/a" to match the string above it)

FredyC:
Thanks for your contribution.

PR #2202, opened 2019-11-12
Make initial values of observable set accept readonly array

upsuper:
`ObservableSet` doesn't need to change the initial values array, so it makes sense to allow passing in a `readonly` one.

### DIFF
--- a/src/types/observableset.ts
+++ b/src/types/observableset.ts
@@ -29,7 +29,7 @@ import {
 
 const ObservableSetMarker = {}
 
-export type IObservableSetInitialValues<T> = Set<T> | T[]
+export type IObservableSetInitialValues<T> = Set<T> | readonly T[]
 
 export type ISetDidChange<T = any> =
     | {


upsuper:
@FredyC Not quite sure about your process. Are you going to merge this PR?

FredyC:
Well, the process generally is that someone approves and other maintainer merges.

I am curious how did you even discover such an "issue"? Why does it bother you? :)

upsuper:
> Well, the process generally is that someone approves and other maintainer merges.

Oh, okay, thanks for the update.

> I am curious how did you even discover such an "issue"? Why does it bother you? :)

So, we try to use `.replace()` to update an `ObservableSet`, but with a readonly array returned from some other function. And TypeScript complains about it. We currently work around this issue via `set.replace([...readonlyArray])` which seems to be pointless.

ItamarShDev:
I can merge @FredyC

upsuper:
Ping on this? @ItamarShDev could you merge?

PR #2207, opened 2019-11-17
The disposer returned from reaction should be auto bound to correct c…

wangyiz4262:
* [x] Added unit tests

### DIFF
--- a/src/core/reaction.ts
+++ b/src/core/reaction.ts
@@ -178,7 +178,7 @@ export class Reaction implements IDerivation, IReactionPublic {
         globalState.globalReactionErrorHandlers.forEach(f => f(error, this))
     }
 
-    dispose() {
+    dispose = () => {
         if (!this.isDisposed) {
             this.isDisposed = true
             if (!this._isRunning) {
@@ -190,8 +190,8 @@ export class Reaction implements IDerivation, IReactionPublic {
         }
     }
 
-    getDisposer(): IReactionDisposer {
-        const r = this.dispose.bind(this) as IReactionDisposer
+    getDisposer = (): IReactionDisposer => {
+        const r = this.dispose as IReactionDisposer
         r[$mobx] = this
         return r
     }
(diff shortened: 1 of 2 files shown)


urugator:
Class instance methods aren't supposed to be destructurable, it doesn't make any sense, it's not a plain object. You could do that with any method and completely kill the point of having prototype (class).

wangyiz4262:
@urugator Thanks for raising your concern, and here are my thoughts:

1. The disposer was not correctly bound before using the arrow function. You have to call `reaction.dispose()` to bind the correct context.
2. With a method defined on prototype gives you the benefit of being able to override from a derived class. As a user, unless you intend to create a child class of `Reaction` and override the dispose method, otherwise I don't see the necessity of keeping the prototype.
3. Your argument of  class instance shouldn't be destructured is very interesting. Could you elaborate more about it?

Thanks!

FredyC:
Personally, I haven't see much of use cases for disposing within the reaction effect callback conditionally. Do you have some practical examples where it can be useful?

I think that explicit call to `reaction.dispose()` might be more sensible so it's apparent at a first glance what is being disposed.

Also, note that moving those methods to instance instead of the prototype might be breaking change for some people who are extending from Reaction (for whatever reasons) or are doing some shenanigans based on content on the prototype.

As such I don't think it's worth it to merge this for a very little gain and potentially big and ugly breaks.

urugator:
`reaction.dispose()`
is an equivalent of 
`Reaction.prototype.dispose.call(reaction)`
There is just a single function in memory - the one defined on prototype.

`dispose = () => {}`
Is an equivalent of 
```javascript
constructor() {
  this.dispose = Reaction.prototype.dispose.bind(this);
}
```
There is a seperate physical copy of that function per every `reaction` instance.

The main benefit you get is memory optimization.
Not every class is inheritable (by design) ... implying that the ability to inherit and override behavior is not the decisive factor for using classes (and therefore for binding it's methods).
In short: If you do use classes, then it doesn't make sense to bind their methods, which unfortunatelly means that you can't pass their methods as standalone references.

FredyC:
So I suppose we shouldn't be merging this. What do you think @urugator?

urugator:
I wouldn't.

FredyC:
@wangyiz4262 Thank you for contribution, but unless you come up with some good use case, it's not worth changing this.

PR #2208, opened 2019-11-17
The disposer returned from reaction should be auto bound to correct c…

wangyiz4262:
* [x] Added unit tests
* [x] Verified that there is no significant performance drop (`npm run test:performance`)


wangyiz4262:
Hi @mweststrate, do you know why some of the checks failed? I didn't see any useful information in the Netlify logs. How do I re-run those checks?

FredyC:
@wangyiz4262 That's because in this branch there are no docs to build, so it fails. We shall set up some exclusion rules in netlify.toml.

PR #2213, opened 2019-11-19
Fix type inference of the action callback arguments

kubk:
Fixes: https://github.com/mobxjs/mobx/issues/2211

* [x] Added typescript typings
* [x] Verified that there is no significant performance drop (`npm run test:performance`)

I prepared the links to TS playground with multiple test cases so you can verify that type inference works now.

[...]

PR #2217, opened 2019-11-22
Type tests for the action callback arguments

kubk:
* [x] Added unit tests
* [ ] Updated changelog

As discussed in https://github.com/mobxjs/mobx/pull/2213 we decided to add TS type tests to prevent typing regressions in the future. Currently, tests are failing because types of the action callback arguments are inferred incorrectly: they are all converted to 'unknown' type. This fix https://github.com/mobxjs/mobx/pull/2213 will make tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "@typescript-eslint/eslint-plugin": "^1.4.2",
     "babel-jest": "^24.5.0",
     "chalk": "^1.1.3",
+    "conditional-type-checks": "^1.0.4",
     "coveralls": "^3.0.3",
     "eslint": "^5.15.2",
     "flow-bin": "^0.59.0",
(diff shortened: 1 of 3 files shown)


FredyC:
I assume you are aware that tests are actually failing here? :)

kubk:
@FredyC Sure, this PR is blocked by  #2213

kubk:
@FredyC Are you suggesting to change these assertions to match actual behavior (e.g replace `assert<IsExact<typeof a1, string>>` with `assert<IsExact<typeof a1, unknown>>`) so we can merge it before #2213 and then fix these correct type assertions in #2213 pull request?

FredyC:
Oh sorry, I understand now :) Merged #2213 so please rebase here.

kubk:
@FredyC Thank you, rebased this PR onto master. Will push my changes to the `mobx4-master` branch.

PR #2218, opened 2019-11-23
Backport of #2217 and #2213 for Mobx 4

kubk:
Fixed type inference of the action callback arguments. Added type tests for the action callback arguments.

### DIFF
--- a/src/api/action.ts
+++ b/src/api/action.ts
@@ -11,33 +11,9 @@ import {
 
 export interface IActionFactory {
     // nameless actions
-    <A1, R, T extends (a1: A1) => R>(fn: T): T & IAction
-    <A1, A2, R, T extends (a1: A1, a2: A2) => R>(fn: T): T & IAction
-    <A1, A2, A3, R, T extends (a1: A1, a2: A2, a3: A3) => R>(fn: T): T & IAction
-    <A1, A2, A3, A4, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4) => R>(fn: T): T & IAction
-    <A1, A2, A3, A4, A5, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5) => R>(fn: T): T &
-        IAction
-    <A1, A2, A3, A4, A5, A6, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4, a6: A6) => R>(fn: T): T &
-        IAction
-
+    <T extends Function | null | undefined>(fn: T): T & IAction
     // named actions
-    <A1, R, T extends (a1: A1) => R>(name: string, fn: T): T & IAction
-    <A1, A2, R, T extends (a1: A1, a2: A2) => R>(name: string, fn: T): T & IAction
-    <A1, A2, A3, R, T extends (a1: A1, a2: A2, a3: A3) => R>(name: string, fn: T): T & IAction
-    <A1, A2, A3, A4, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4) => R>(name: string, fn: T): T &
-        IAction
-    <A1, A2, A3, A4, A5, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5) => R>(
-        name: string,
-        fn: T
-    ): T & IAction
-    <A1, A2, A3, A4, A5, A6, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4, a6: A6) => R>(
-        name: string,
-        fn: T
-    ): T & IAction
-
-    // generic forms
-    <T extends Function>(fn: T): T & IAction
-    <T extends Function>(name: string, fn: T): T & IAction
+    <T extends Function | null | undefined>(name: string, fn: T): T & IAction
 
     // named decorator
     (customName: string): (
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "@typescript-eslint/eslint-plugin": "^1.4.2",
     "babel-jest": "^24.5.0",
     "chalk": "^1.1.3",
+    "conditional-type-checks": "^1.0.4",
     "coveralls": "^3.0.3",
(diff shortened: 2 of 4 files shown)


kubk:
It seems like the build failure is not related to my changes:

> Error running command: fork/exec /usr/local/bin/build: no such file or directory

src/api/action.ts
```diff
@@ -11,33 +11,9 @@ import {
...
-    <A1, R, T extends (a1: A1) => R>(fn: T): T & IAction
-    <A1, A2, R, T extends (a1: A1, a2: A2) => R>(fn: T): T & IAction
-    <A1, A2, A3, R, T extends (a1: A1, a2: A2, a3: A3) => R>(fn: T): T & IAction
-    <A1, A2, A3, A4, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4) => R>(fn: T): T & IAction
-    <A1, A2, A3, A4, A5, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4, a5: A5) => R>(fn: T): T &
-        IAction
-    <A1, A2, A3, A4, A5, A6, R, T extends (a1: A1, a2: A2, a3: A3, a4: A4, a6: A6) => R>(fn: T): T &
-        IAction
-
+    <T extends Function | null | undefined>(fn: T): T & IAction
```
FredyC: Hm, I haven't realized in the previous PR, but why is it ok to expect `null | undefined` here? I mean actions should require a function, it's shouldn't be optional.

kubk: The `null | undefined` was added here to support type inference inside Promise methods, it is described here in [Example 3](http://www.typescriptlang.org/play/?ssl=32&ssc=1&pln=32&pc=40#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgSQEFsYEIk4BvAXwChbRJZEU0tcDjTyAxHGaAE8qtOGLgB6CXADmwJGgTY46aAFsAzqPEAeACpxQqJABMNcHgFckJMkgB8ACnRIAXHD0BKdwYBknW3JtMX1DEGMzC2tAh0ckTDVgdw0YKGQZABoVNw9vDzh-Ihj6cUlpeMSTOBNgbGhMAShguEdsSxSINQA5BKS4FLSkGTzHZtKYTCg5GHcAeQAjACtamAyx8QBrYEFk1PS4AB9+wTV5iAAbNdLS+cwNYAARYA1sNLBGgH53AAUoCDA0DBBE8Xm9Gs1PHAALz2OAANwgCBMJXEUjg1gqwCqNTqUAa0GajgmU2AMzgC2WJCyYD+ANgggA0ttdoMZIdjqcLlkaqCEO9oF84L9-oDgc9XnzGnkEUiUWI0QABfh2AB0Z2s2Nq9XBNwgGqJk2mcyWK2ptNFTJ2-T2Q3ZGhOZ0u1XFYIFP3N9JBEv5UGliORdFoaL0zxgWgY4Gg8Bx50meDhkzgyvI7iK3CQfBIQgA3PR0NF03BUCkAIyGjTOHKOTDuJCWU5oSEwpNIQSQmi0fM2QvFmAAJnLlfc1ZZ6Sy81r9fmjehsMwrfbdC7MSLoYA7IOXMOa9bWePJw2oFlsAfp0fqqOhlk+gMxypT2gsjJL8NZy221RA2ieolzs9zMmSBaL2ZZTBWgGONWJZNrCdRARcwAqucEAyFBnjobm34QB4ADKiAaBoljPO4uJQCsRaCACLDoGg8i4LQvYDmB1YxJBmAllkmB9jBcBwRoCFIShUGcdxGEMeug4QcJSZ9pxADMnEACycQArJxABsnFrjxfECchqGYNpYlBtIP5YkmMTAaGoEyOBrEAOS9vZWRoW+ul-oJBnQWJWG4fhhHEbx0BkSQFFUcgNFkTYwDiSkTG2Sx6aOI5obOS07EiTp5D8R5+nSVx6GeLmvYbsxEEpSkaX5bJSYKUmylJmpSaaUm2ludlelCYZhW5iZQp-GoCD3FoaL4ARRHuNYGxIBAADuFBAgCtDCoN9wqmROVwsAjjQSqMAABbyIldhsdB7XwblXVnWJK1DYhG0XFtO2eHth1IMd5DJU5LnsVlF2IXlv1wGJQA): https://github.com/mobxjs/mobx/pull/2213 

Let me describe it in greater depth. Here is TypeScript typings for `Promise.prototype.then()`: 
https://github.com/microsoft/TypeScript/blob/b9689228b5928b2524faf7f0e9ff5f5eba42a7b7/lib/lib.es5.d.ts#L1410
If we reformat this unreadable one-liner we will see the following:
```ts
interface Promise<T> {
    then<TResult1 = T, TResult2 = never>(
        onfulfilled?: ((value: T) => TResult1 | PromiseLike<TResult1>) | undefined | null,
        onrejected?: ((reason: any) => TResult2 | PromiseLike<TResult2>) | undefined | null
    ): PromiseLike<TResult1 | TResult2>;
}
```
The first argument of method `then` has signature `Function | undefined | null`. That's why I used `<T extends Function | null | undefined>` instead of `<T extends Function>`. Type `Function | null | undefined` is not a subtype of `Function`, the left type should not be widen than the right. If you replace `<T extends Function | null | undefined>` with `<T extends Function>` this test will fail: https://github.com/mobxjs/mobx/blob/20a2ffc58e0d12c687d397f876d73a022a64dc12/test/base/typescript-tests.ts#L1721

I agree it would be nice to get rid of this `null | undefined`, but the main problem is TS typings for `Promise.prototype.then()`. I also can document the reason why `null | undefined` added here: 
https://github.com/kubk/mobx/blob/52cd4ae23a6749c993c8526d3889c0a3bc0532d2/src/api/action.ts#L14

FredyC: Thanks for the explanation. Let's refer to this in the changelog, no need to put into the code.

kubk: @FredyC Thank you.

PR #2219, opened 2019-11-25
fix broken links

Xapuu:
remove links pointing to migrated docs

Part of the mobx docs had been moved to [Awesome mobx](https://github.com/mobxjs/awesome-mobx#awesome-mobx), which leads to navigation to non existing pages.

PR #2221, opened 2019-11-28
doc: Update wrong links in README.html file

alexakasanjeev:
The README.html file for the website has wrong links, which on click lead to 404 error pages

boilerplate


FredyC:
The #2219 has fixed it just yesterday, but it's not deployed yet.

PR #2229, opened 2019-12-10
Update browser support

jasoncrawford:
The FAQ and the README contradict each other re: browser support, and the README seems correct. I wasn't sure how much to duplicate the info here. Feel free to rewrite this, but I thought the FAQ should be updated. Thanks!

### DIFF
--- a/docs/faq/faq.md
+++ b/docs/faq/faq.md
@@ -10,7 +10,7 @@ hide_title: true
 
 ##### Which browsers are supported?
 
-MobX runs on any ES5 environment. That means that Node.js, Rhino and all browsers except for IE8 are supported. See [caniuse.com](http://caniuse.com/#feat=es5)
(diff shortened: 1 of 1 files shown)


FredyC:
Perhaps we should add a link to single source of truth instead of copying that information.

https://mobx.js.org/README.html#browser-support

jasoncrawford:
> Perhaps we should add a link to single source of truth instead of copying that information.

Yes, I did include that link. Would it be better as *just* that link and nothing else?

FredyC:
Yes, I think so, let's keep duplicate information at a minimum.

jasoncrawford:
@FredyC OK, done

docs/faq/faq.md
```diff
@@ -10,7 +10,7 @@ hide_title: true
 
 ##### Which browsers are supported?
 
-MobX runs on any ES5 environment. That means that Node.js, Rhino and all browsers except for IE8 are supported. See [caniuse.com](http://caniuse.com/#feat=es5)
+See the section on Browser Support in [the README](https://mobx.js.org/README.html#browser-support).
```
FredyC:
```suggestion
See the [browser support](https://mobx.js.org/README.html#browser-support).
```

Issue #2237, opened 2019-12-17
shallow comparer uses incorrect depth for Map comparison


hearnden:
The shallow comparer was introduced in fc70ee3e7596159ce9ac588f070ee4058aca5b2b.

A shallow comparison of Map objects should presumably compare their keys and values.  However, the current shallow comparer does not do this.  This can be seen by the following test failure:
```js
   expect(sh(new Map([[1, 2]]), new Map([[1, 2]]))).toBe(true)
```
that could be added to the shallow-comparer tests: https://github.com/mobxjs/mobx/blob/3fe3ada1ee2cd319a8f5d21d8f2a569ac054722a/test/base/extras.js#L491

I believe this is because Maps are unwrapped to an Array of entries:
https://github.com/mobxjs/mobx/blob/3fe3ada1ee2cd319a8f5d21d8f2a569ac054722a/src/utils/eq.ts#L138
which introduces an extra level of depth to the data being compared.  However, the `depth` argument to `deepEqual` remains at 1:
https://github.com/mobxjs/mobx/blob/fc70ee3e7596159ce9ac588f070ee4058aca5b2b/src/utils/comparer.ts#L16

This means that a shallow comparison of `Map`s is effectively performing an identity comparison of their `entries()`, which will probably always fail for non-empty maps, and that is unlikely to be useful.

This can potentially be fixed by increasing the depth by 1 whenever `unwrap` encounters a pair of `Map`s.

hearnden:
Proposed fix in PR: https://github.com/mobxjs/mobx/pull/2238

mweststrate:
Released as (4/5).25.2. Thanks again!

PR #2238, opened 2019-12-17
supports ES6 Sets and Maps in shallow comparer.

hearnden:
From: https://github.com/mobxjs/mobx/issues/2237

Adds these test:
```js
expect(sh(new Set([1, 2]), new Set([1, 2]))).toBe(true)
expect(sh(new Map([[1, 2]]), new Map([[1, 2]]))).toBe(true)
```
which fails.  The rest of the changes are to make it pass.

Note that this PR is a superset of the changes from #2240.

I have only prepared a minimal proof-of-concept to illustrate the issue and propose a fix.  I can add more thorough tests after some initial feedback.  I'm also happy for someone else to take this over if they prefer.

* [x] Added minimal unit tests
* [ ] Updated changelog

### DIFF
--- a/src/utils/eq.ts
+++ b/src/utils/eq.ts
@@ -56,7 +56,10 @@ function eq(a: any, b: any, depth: number, aStack?: any[], bStack?: any[]) {
             )
         case "[object Map]":
             // Maps are unwrapped to an array of entries, each entry being an intermediate array.
-            depth++
+            // Hide this intermediate level by increasing the depth.
+            if (depth >= 0) {
+              depth++
+            }
             break
     }
     // Unwrap any wrapped objects.
--- a/src/utils/eq.ts
+++ b/src/utils/eq.ts
@@ -55,10 +55,11 @@ function eq(a: any, b: any, depth: number, aStack?: any[], bStack?: any[]) {
                 typeof Symbol !== "undefined" && Symbol.valueOf.call(a) === Symbol.valueOf.call(b)
             )
         case "[object Map]":
-            // Maps are unwrapped to an array of entries, each entry being an intermediate array.
-            // Hide this intermediate level by increasing the depth.
+        case "[object Set]":
+            // Maps and Sets are unwrapped to arrays of entry-pairs, adding an incidental level.
+            // Hide this extra level by increasing the depth.
             if (depth >= 0) {
-              depth++
+                depth++
             }
             break
     }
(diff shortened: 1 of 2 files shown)


BrendanAnnable:
Any objections to this one @mweststrate?

FredyC:
Please remove WIP from the title if you are satisfied with it. And be patient please, less activity before the year-end is expected. I am sure eventually someone will get to this.

hearnden:
> Please remove WIP from the title if you are satisfied with it.

Done!

urugator:
Would you mind adding these tests? 
The idea is to test the observable counterparts (since they can be a subject of unwrapping) and to test for non-equality. It's just for completeness, it shouldn't break anything.
You can adapt them to fit the existing tests if you like (or correct them in case there is a mistake).
```javascript
expect(sh(new Set([1, 2]), new Set([1, 3]))).toBe(false)
expect(sh(new Map([[1, 2]]), new Map([[1, 3]]))).toBe(false)
expect(sh(observable({ a: 1, b: 2 }), observable({ a: 1, b: 2 }))).toBe(true)
expect(sh(observable({ a: 1, b: 2 }), observable({ a: 1, b: 3 }))).toBe(false)
expect(sh(observable(new Set([1, 2])), observable(new Set([1, 2])))).toBe(true)
expect(sh(observable(new Set([1, 2])), observable(new Set([1, 3])))).toBe(false)
expect(sh(observable(new Map([[1, 2]])), observable(new Map([[1, 2]])))).toBe(true)
expect(sh(observable(new Map([[1, 2]])), observable(new Map([[1, 3]])))).toBe(false)
expect(sh(observable([1, 2]), observable([1, 2]))).toBe(true)
expect(sh(observable([1, 2]), observable([1, 3]))).toBe(false)
```

If you're interested, we would also use deep equality tests for `Map/Set` as they seem to be completely omitted.

I think this is sufficient solution for the time being, however we will probably need an order agnostic comparator eventually.

hearnden:
> Would you mind adding these tests?

Done, and then some.

I pulled out the type-comparison tests to a separate `describe()`.

FredyC:
Do we need the same change for MobX4? If yes, can you please prepare it @hearnden? It should go to mobx4-master branch.

hearnden:
> Do we need the same change for MobX4? If yes, can you please prepare it @hearnden? It should go to mobx4-master branch.

Sure thing, I'm happy to prepare that.  I can look into it later this week.

PR #2243, opened 2019-12-21
Fixes #2230 - computedvalue: throw error object instead of string when options is e…

### LINKED ISSUE #2230: MobX errors thrown as strings aren't properly caught and obfuscate the actual error
Instead of seeing the error I was getting an error boundary error;

```
checkPropTypes.js:20 Warning: Failed prop type: Invalid prop `error` of type `String` supplied to `ErrorBoundaryFallbackComponent`, expected instance of `Error`.
    in ErrorBoundaryFallbackComponent (created by ErrorBoundary)
```

The actual error was 
```
"[mobx] missing option for computed: get" 
```

The problem is this is thrown as `string`

https://github.com/mobxjs/mobx/blob/728ad70feef5856060e1dd30516ca1c6161803dd/src/core/computedvalue.ts#L109

As most code catching errors expects actual Error objects It should be `Error`

I

ramkumarvenkat:
…mpty

Fixes #2230 

PR checklist:

* [ ] Added unit tests
* [X] Updated changelog

### DIFF
--- a/src/core/computedvalue.ts
+++ b/src/core/computedvalue.ts
@@ -105,8 +105,7 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
      * This is useful for working with vectors, mouse coordinates etc.
      */
     constructor(options: IComputedValueOptions<T>) {
-        if (process.env.NODE_ENV !== "production" && !options.get)
-            throw "[mobx] missing option for computed: get"
+        invariant(options.get, "missing option for computed: get")
         this.derivation = options.get!
         this.name = options.name || "ComputedValue@" + getNextId()
         if (options.set) this.setter = createAction(this.name + "-setter", options.set) as any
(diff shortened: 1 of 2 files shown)


ramkumarvenkat:
Is there a way I can unit test these changes? Any help appreciated!

src/core/computedvalue.ts
```diff
@@ -105,8 +105,7 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
      * This is useful for working with vectors, mouse coordinates etc.
      */
     constructor(options: IComputedValueOptions<T>) {
-        if (process.env.NODE_ENV !== "production" && !options.get)
-            throw "[mobx] missing option for computed: get"
+        invariant(options.get, "missing option for computed: get")
```
FredyC: Let's keep `[mobx]` prefix there.

FredyC:
Not sure about how to unit test this or if it's necessary. Update changelog, please. And also we will need another PR to `mobx4-master`. You can wait with that when this one is approved.

src/core/computedvalue.ts
ramkumarvenkat: https://github.com/mobxjs/mobx/blob/master/src/utils/utils.ts#L36-L38 adds the prefix

ramkumarvenkat:
@FredyC Thanks! I have updated the changelog to create a new version.

Also the PR on mobx4-master: https://github.com/mobxjs/mobx/pull/2244

Issue #2120, opened 2019-09-21
[Idea] Maintain and publish mobx 4 & 5 from the same branch


Bnaya:
I can't say i know mobx codebase very well,
But as what i've seen most of the code of 4 & 5 is very similar,
The difference is mostly the proxy part.

What if struct the code in a way that only the different parts will be behind a build flag (or even runtime flat if the developer wishes)
And keep all of the code on master.
Another idea, that i'm not sure if will work:
Use monorepo, one package will be 4.x  and the other 5.x, and the shared code will be in a shared package (That won't be published, but will be bundled)
 
Kinda related #1957

FredyC:
Monorepos are the pain, mainly because there are no solid tools to with them (Lerna is crap).

Merging codebase does sound appealing, but I can imagine it would make code fairly messy even with build flags. Not sure really. If you are willing to give it a try, it could be really interesting and easing the maintenance pain toward the future.

Bnaya:
The initial idea is to keep publish V4 & V5,
just from the same branch (with separate package.json, or writing the version on the fly before publish),
And build/bundle flags will produce the correct bundle before each publish.
I believe i will find time to POC it after #2079 will get finalized

mweststrate:
Separate subpackages in the same branch sounds fine to me, if it isn't confusing for new contributors. Mixing code bases is a real pain. I tried, so I can confirm that it at least is not easy :).

Bnaya:
My personal experience from my last PR, even a small PR to both of the branches was confusing.

Have you tried that with mobx or other projects?

FredyC:
> Separate subpackages in the same branch sounds fine to me, if it isn't confusing for new contributors.

I can imagine it could be confusing for reviews to basically see changes two times in there.

> Mixing code bases is a real pain. I tried, so I can confirm that it at least is not easy :).

Not easy but also not impossible I assume? I kinda agree it would be the best way if it can work out somehow.

mweststrate:
mobx-state-tree packages multiple projects in the same repo. Works pretty
smooth (I'm generally in favor of mono repo's and single branches of work,
as it makes coordination so much easier, so I do think it is a good idea :))

mweststrate:
> I can imagine it could be confusing for reviews to basically see changes
two times in there.

Agreed, but I think in the PR template we can signal that the change
initially should only be made to the mobx5 one, until approved, and then
backporting can be done (in same PR), but that reduces burden for both
[...]

PR #2248, opened 2019-12-31
Combine V4 & V5 together

FredyC:
We talked about in #2120.

I created `src/v4` and copied `mobx4-master` over to it. The `src/v5` is source of a current master. That was an easy part :) In some distant future, we might have a `common` folder with code that is the same for both versions.

## Next steps

- [x] Figure out tests to be executed against V4 and V5 (some tests might be specific)
- [x] Publish process to build & publish both versions at the same time
- [x] Flow types seem to be identical, any chance deviation will be required in the future?
- [ ] Tests for mixed versions are broken
- [x] Consolidate tsconfig.json from a separate src/v(4|5) folders


mweststrate:
@FredyC awesome for picking this up! I think it is greatly gonna help managing the two versions. 

As initial thought I would still keep the build, test and source process completely separate. The reason for this is that v4 has (iirc, and otherwise probably will in the future) different browser and compatibility requirements, so the build and TS configuration might very well differ over time, and it would be a shame that v5/6/7 is build with unnecessary polyfills or language restrictions because of v4.

For the same reason I would keep tests apart, so that we can be picky about what we backport and what not. Also it might make developing a feature easier if you can focus on first getting it right in one of the two. 

Note that MST has also a multi package setup.

I think always releasing both on the same time would be ok by the way, even if it means that one of them occassionally has no actual change.

FredyC:
> Edit: same for flow types.

I am not that familiar with Flow and I am not sure could that be split by version actually. However, I do think that having the same API layer for both versions and thus the same types is a sort of sanity check as well.

> Note that MST has also a multi package setup.

Yea, but that's just a regular Lerna based monorepo. I believe this is something different. I don't even think Lerna allows to publish to the same package with two different versions at once.

I can really imagine over time to have an `src/common` and `src/test` folders to contain bits and pieces that will definitely stay the same across versions. It might increase reliability and confidence for the stable core that should not break.

I agree with everything else. Take it slowly, for now, and expand on the idea later. I also want to update the PR template to explain how to contribute to this new structure.

Currently, I am waiting for that new PR to be resolved to avoid painful patching. Then I will jump right into it.

mweststrate:
> Yea, but that's just a regular Lerna based monorepo. I believe this is something different. I don't even think Lerna allows to publish to the same package with two different versions at once.

I'd just put two packages into one repo for starter, and wire up the scripts. Lerna is probably overkill here and has it's own issues. This is not an `n` packages problem, but 2 pacakages. So I think things can remain KISS, and we can just call build, test, publish etc in two different directories, and have both packages with a package.json that has the same name. The build setup doesn't change often, so I fear the complexity of tools like lerna more than having slightly longer scripts :)

FredyC:
Alright, I am done here. By tomorrow I will merge this and we can figure out possible issues later. Probably best not to keep this PR open for too long.

## Summary of changes

-   master/src -> master/src/v5
-   master/test -> master/test/v5
-   mobx4-master/src -> master/src/v4
-   mobx4-master/test -> master/test/v4
-   root package.json is source of truth for version (minor.patch)
-   build is done for both versions with output to `dist/vX` folder
-   package.json is copied over to these dist folders with correct major version
-   publish will bump version (patch or minor) in root package.json and in dist as well
-   test/perf modified to run against both versions
-   upgraded size-limit and configured to run on both versions

As for publishing, we agreed with Michel it's fine to keep versions in sync even if there would be change in either version only. Updated publishing script does streamline this process.

Generally tests are working. The "mixed-versions" test is broken completely and I did not want to bother with that now and delay this PR more.

The "webpack-regression" tests are green in CI build, but failing nonetheless. Not sure if it's even needed and what was purpose of that.

mweststrate:
> Generally, tests are working. The "mixed-versions" test is broken completely and I did not want to bother with that now and delay this PR more.

Move them over to a separate PR for now in that case?

> The "webpack-regression" tests are green in CI build, but failing nonetheless. Not sure if it's even needed anymore and what was the purpose of that.

It's really old, to verify somewhere in the past that we bundled correctly. Feel free to drop it if it is in the way.

One critical issue I found is that version 4 will be published under the `latest` tag in npm. It should get the the `mobx4` tag, so that people do get a 5 version installed always by default. 

Beyond that LGTM. So feel free to merge if above things are addressed. Than we can deal with the fallout later.

FredyC:
Thanks for pointing out about dist-tag, did not realize that, so that's fixed now. Removed webpack tests.

Who knows how long have been the mixed-versions tests broken, it wasn't executing as part of CI. I will open a PR with that added to CI and someone can hopefully look at that later.

mweststrate:
awesome, thanks for doing this!

PR #2252, opened 2020-01-09
extendObservable: remove constraint for existing property

davefeucht:
Currently extendObservable throws an error if you try to use it on a property which already exists in the target object/class. This makes it impossible to dynamically make an existing property observable.

* [x] Added unit tests
* [x] Updated changelog
* [x] Updated `/docs`. For new functionality, at least `API.md` should be updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,9 @@
-# 5.15.3 / 4.15.3 [NOT PUBLISHED YET]
-
--   Fixes #2250 - extendObservable: allows it to be used on an existing property [#2252](https://github.com/mobxjs/mobx/pull/2252)
-
 # 5.15.2 / 4.15.2 [NOT PUBLISHED YET]
 
 -   Fixes #2230 - computedvalue: throw error object instead of string when options is empty [#2243](https://github.com/mobxjs/mobx/pull/2243)
 -   supports ES6 Sets and Maps in shallow comparer. [#2238](https://github.com/mobxjs/mobx/pull/2238)
+
+-   Fixes #2250 - extendObservable: allows it to be used on an existing property [#2252](https://github.com/mobxjs/mobx/pull/2252)
 
 # 5.15.1 / 4.15.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 5.15.3 / 4.15.3 [NOT PUBLISHED YET]
 
--   Fixes #2250 - extendObservable: allows it to be used on an existing property [#2250](https://github.com/mobxjs/mobx/pull/2250)
+-   Fixes #2250 - extendObservable: allows it to be used on an existing property [#2252](https://github.com/mobxjs/mobx/pull/2252)
 
 # 5.15.2 / 4.15.2 [NOT PUBLISHED YET]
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 5.15.3 / 4.15.3 [NOT PUBLISHED YET]
+
+-   Fixes #2250 - extendObservable: allows it to be used on an existing property [#2250](https://github.com/mobxjs/mobx/pull/2250)
```
FredyC: Move this into a previous unpublished version please.

davefeucht: Oh, put it in 5.15.2 you mean?

test/base/extendObservable.js
```diff
@@ -0,0 +1,119 @@
...
+        },
+        width: 40
+    })
+
+    expect(isObservableProp(box, "height")).toBe(true)
+    expect(isObservableProp(box, "sizes")).toBe(true)
+    expect(isObservable(box.sizes)).toBe(true)
+    expect(isObservableProp(box, "someFunc")).toBe(true)
+    expect(isComputedProp(box, "someFunc")).toBe(true)
+    expect(isObservableProp(box, "width")).toBe(true)
```
mweststrate: Maybe overwritting a getter with an observable should just throw, as it is probably a mistake? Not sure

test/base/extendObservable.js
```diff
@@ -0,0 +1,119 @@
...
+            this.sizes.push([4])
+        }
+    }
+
+    box.undeclared = 1
+
+    extendObservable(box, {
+        height: 20,
+        sizes: [2],
+        get someFunc() {
```
mweststrate: Probably would be good to have a test that tests the `extendObservable` with three arguments as well; where the type of decorator to apply is set by hand (e.g. `extendObservable(box, { x: function() }, { x: action })`

test/base/extendObservable.js
davefeucht: So something like:

```
if (Object.getOwnPropertyDescriptor(target, key) && Object.getOwnPropertyDescriptor(target, key).get) {
    fail()
}
```
?

test/base/extendObservable.js
davefeucht: Ok, sounds good.

test/base/extendObservable.js
mweststrate: Yeah precisely that, if you agree with the constraint :)

davefeucht: On one hand, I kind of like an all or nothing approach with not special cases like this, but it is unlikely that overwriting a getter is what the user wants to do.

davefeucht: Hmm, but what if you want to overwrite a getter with a computed?

test/base/extendObservable.js
davefeucht: Test added to make someFunc() just a normal function and make it an action with extendObservable.

test/base/extendObservable.js
davefeucht: I think I would leave it with no constraint, and leave it up to the user to understand what they are overwriting.

mweststrate: Perfect, makes sense!

davefeucht:
Yeah, I actually specifically need it for Mobx4 :)

davefeucht:
Do you have an idea how long until it will be released, by the way?

mweststrate:
I'm a little behind with cutting releases, but early next week should be doable I think. Would that work for you? 

If interested, feel free to open a PR against mobx4-master as well and cherry-pick the changes (only test and implementation)

We are working on making that process smoother😅

davefeucht:
Yeah, early next week should be fine. I’ll make an MR for mobx4-master first thing in the morning with just the code and tests.

mweststrate:
THanks!

Issue #2253, opened 2020-01-09
Change of behavior in ObservableMap.replace when using intercept in Mobx v4


Lmzd:
### Version
Mobx: 4.13.1
Node: v13.2.0

### Issue
**Only in mobx4**

Following the changes in #1980 and #2003, visible in v4.13.1. We are facing a difference in the `replace` method of `ObservableMap` in v4.13.0.

When we `intercept` a change on a given property of the Map and **cancel it** (`return null`), the value is not set in the Map  (expected) but the `internal keys` array of the Map is updated with the name of the intercepted property. 

_**Change** the version of mobx to **4.13.0** to see the previous behavior_
[Code to reproduce](https://codesandbox.io/s/mobx-observable-map-replace-ts99u)

urugator:
Could you try this version?
```javascript
replace(values: ObservableMap<K, V> | IKeyValueMap<V> | any): ObservableMap<K, V> {
  transaction(() => {
      const replacementMap = convertToMap(values)
      const oldKeys = this._keys
      const newKeys: Array<any> = Array.from(replacementMap.keys())
      let keysChanged = false
      for (let i = 0; i < oldKeys.length; i++) {
          const oldKey = oldKeys[i]
          // key order change
          if (oldKeys.length === newKeys.length && oldKey !== newKeys[i]) {
              keysChanged = true
          }
          // deleted key
          if (!replacementMap.has(oldKey)) {
              this.delete(oldKey)
              // intercept could cancel the delete operation, check if key is truly deleted
              if (!this._data.has(oldKey)) {
                keysChanged = true
              }
          }
      }
      replacementMap.forEach((value, key) => {
          // new key
          const keyMaybeAdded = !this._data.has(key);
          this.set(key, value)
          // intercept could cancel the set operation, check if key is truly added
          if (keyMaybeAdded && this._data.has(key)) {
            keysChanged = true
          }
      })
      if (keysChanged) {
          this._keys.replace(newKeys)
      }
  })
  return this
}
```

Notes:
V5 is most likely affected as well.
Check if there are other operations that delegates to `this.set/this.delete`
Test also a removal of a key.
Provide internal `_set/_delete` with outcome rather than re-checking keys.

GLabat:
Hi thanks for your reply.

We checked MobX 5 and the code is the same as in v4.13.0, so **before** the change on `replace`.

[...]

PR #2255, opened 2020-01-09
[WIP]: Change of behavior in ObservableMap.replace

Lmzd:
### Issue #2253 

- Adding test to reproduce error in v4


FredyC:
What's missing here given it's still WIP?

Lmzd:
@FredyC  For the moment, I just implemented the test. I'm waiting for a little more context to implement a fix.

Lmzd:
@FredyC How could I finish a CI ? (mobx-docs and netlify errors ...)

Lmzd:
@FredyC How could I finish a CI ? (mobx-docs and netlify errors ...)

FredyC:
Don't worry about netlify, it doesn't work for MobX4 branch. I will soon fix that with #2248

FredyC:
Can you please add an extra test when something is already in the map? Like in a https://github.com/mobxjs/mobx/issues/2253#issuecomment-572977071

test/base/map.js
```diff
@@ -806,3 +806,20 @@ test("#2253 Map with replace method with intercept", () => {
...
 })
+
+test("#2253 filled Map replace method with intercept", () => {
+    const data = {
+        prop4: 4,
+        propIgnored: "ignored",
+        prop5: 5,
+        prop6: 6
+    }
+    const observedMap = map({ prop4: 1, prop2: 2, prop3: 3 })
```
Lmzd: @FredyC Like this ? not sure to really understand this [#2253 (comment)](https://github.com/mobxjs/mobx/issues/2253#issuecomment-572977071)

FredyC: @urugator What do you think? Is it sufficient or did I misunderstood something as well?

urugator:
First of all there is a problem with both `set` (when adding new key or when changing it's value) and `delete` (when deleting key, which isn't in replacement map) - both of these can be cancelled. 
I think it may be worth trying to fix both at the same as I can imagine the whole replace operation may need a rewrite.

Now there is a problem of what it means to cancel `set/delete` operation in regard to the placement of the element in the resulting map. Eg:
```javascript
{ 1:1, 2:2, 3:3 } // original
{ 3:3, 2:2, 1:1 } // replacement
// Consider there is a `set` interceptor on key 1, which returns false (cancels set)
// In resulting map keys 3 and 2 are swapped, because we respect the key ordering of the replacement map
// but what about the 1?
// If we cancel the `set` operation should we "move" 1 at the end of the array?
// or should we keep it at the beginning?
```
Similary with delete:
```javascript
{ 1:1, 2:2, 3:3, 4:4 } // original
{ 5:5, 2:20 } // replacement
// Consider  there is a `delete` interceptor on key 1 and 4, which returns false (cancels delete)
// 5 is added
// 2 is moved and modified
// 3 is omitted (deleted)
// 1 and 4 must remain in in map ... where?
```
So to continue further we need to agree on this key handling first.

mweststrate:
@urugator I think 'insertion order' here basically means that the exiting and remaining keys will preceed the new ones (it is the complete inserts that gets cancelled, updates can be cancelled or not, but don't affect the map order). So I think the result should be 1,2,3 and 1,2,4,5 repectively. 

That being said, I think we might consider that a different bug that could be filed if not correct, and accept this one even when it fixes only a limited use case? But if @lmzd wants to combine the patches that would be appreciated of course!

urugator:
That would mean to reintroduce #1980, in which case I am not sure about correctness of the current impl. 
I also think the current PR doesn't handle correctly key change detection for cancelled additions:
If the key isn't in original map, `set` gets cancelled, `keyHasBeenSet` is false, `keysChanged` is set tu `true` even though, the keys are not modified.

mweststrate:
I think .replace is different from the others, in that sense, that it
should act I think as a full clear + all new items chained, rather than as
one operation. So this always results in the insertion order of the new
data (which was what that issue was about IIRC), and intercepting a single
key beying re-added would effectively result in the loss of that key in the
map?

Maybe the best way to tackle this is by starting a PR with (.skip-ed) unit
tests so that we have a cohesive overview of all the cases and past issues?
(I noticed I tend to scanning comments to quickly and sometimes
misunderstand them, where unit tests leave little room for interpretation)

urugator:
> intercepting a single key beying re-added would effectively result in the loss of that key in the map?

Seems reasonable to me. 
If I follow correctly, given the examples above, the results will be:
```javascript
{ 3:3, 2:2 }
{ 1:1, 4:4, 5:5, 2:20 }
```

@Lmzd does that work for you or your use case?

For the next major I would consider using single `replace` interception event, rather than invoking `set/delete` for individual elements.
Currently there is a problem that the interceptor can access a map which is in inconsistent/unknown state ... so eg. you can't decide about interceptor outcome based on current map elements.
Potentially cou can even break the operation by calling map methods during interception.
Same is probably true for other "complex" operations (clear/merge/...)

Lmzd:
@urugator I implement those 2 test in `test/base/map.js`, unsurprisingly they fail.

Can you check if the test has the good logics please ?
https://github.com/Lmzd/mobx/blob/83f48bde95ac4f495abf596015dee9185da30fe3/test/base/map.js#L837-L883

test/base/map.js
```diff
@@ -789,3 +790,94 @@ test("#1858 Map should not be inherited", () => {
...
+            return null
+        }
+        return change
+    })
+    observedMap.replace(data)
+
+    expect(interceptCount).toEqual(2)
+    expect(observedMap.size).toEqual(4)
+    expect(observedMap._keys.slice()).toEqual(["1", "2", "4", "5"])
+    expect(observedMap.toJSON()).toEqual({ 1: 1, 2: 5, 4: 4, 5: 20 })
```
urugator: `{ 1:1, 4:4, 2:5, 5:20 }`

Lmzd: Ok thanks

urugator: And also note that the interceptor will be called 4 times ... 2x for delete, 2x for set

test/base/map.js
```diff
@@ -789,3 +790,94 @@ test("#1858 Map should not be inherited", () => {
...
+
+    intercept(observedMap, change => {
+        if (change.name === "1") {
+            return null
+        }
+        return change
+    })
+    observedMap.replace(data)
+
+    expect(interceptCount).toEqual(1)
```
urugator: 2 ... as noted above (1 for delete, 1 for set)

test/base/map.js
```diff
@@ -789,3 +790,94 @@ test("#1858 Map should not be inherited", () => {
...
+    expect(observedMap.size).toEqual(2)
+    expect(observedMap._keys.slice()).toEqual(["3", "2"])
+    expect(observedMap.toJSON()).toEqual({ 3: 3, 2: 2 })
```
urugator: `{ 1:1, 3:3, 2:2 }`
delete is cancelled on 1, so you end up with `{ 1:1 }`, then only 3,2 are added, because 1 is cancelled once again during set

urugator:
Omg, I haven't realized that `set` are actually 2 different events, so you can actually prevent addition/update individually... that makes it even more complicated, I suppose? Eg. previously cancelled `delete` (during "clear"), can change the event of `set` (from `add` to `update`)...?

FredyC:
@Lmzd Are you able to finish this soon? Thing is I want to move ahead with #2248 which means a lot of files will change place so I don't want to do that with a lot of elaborate PR open.

Lmzd:
Hi @FredyC , I've got some local changes but this is not good at this point (it breaks some of the #1980  observableMap tests) so IMO you can  move ahead with #2248. I will rebase my work on the future "real" master branch.

FredyC:
@Lmzd The just went through so you need to rebase your work onto a `master` and there `src/v4` and `test/v4` where you should do changes.

Lmzd:
I think `master` does not have `master-v4` commit history so that's why I have all those commit. It could not have commit reconciliation ...

Lmzd:
Changes on `*.md` files are due to linter

FredyC:
Yea, I don't think it's a generally good idea to rebase from V4 branch. History is way too different. It might be easier to just start on current master and cherry-pick changes, but with those files moved it will be painful anyway.

FredyC:
I am kinda surprised you went this way, V4 branch history is totally different, but looks like you got it sorted out, so it's fine. We will squash when merging, so it won't matter.

FredyC:
So where are we with this? Can we move it forward somehow? The test is failing :/

A new issue #2274 for the `replace` has been filled and I wonder if this PR might fix it or break it even more. It might be worth adding that test case in this PR.

mweststrate:
Just did a quick check, this branch also fixes #2274,  and it passes all `v5` tests for me. @lmzd could you share if there are any open items from your side?

FredyC:
Based on the CI output, `v4` tests are failing, `v5` is fine because there are no changes to it in this PR.

mweststrate:
Oh lol, was blindly assuming the changes where made to the v5 one because
that tests weren't failing, and v4 tests needed updating. Too early monday
morning, lol.

mweststrate:
To avoid double work: I think this one can be put [on hold] as the same issues will be investigated by @urugator, including tests / changes / fixes by this PR

urugator:
Closing in favor #2255

PR #2256, opened 2020-01-10
extendObservable: remove constraint for existing property for mobx4

davefeucht:
Same as https://github.com/mobxjs/mobx/pull/2252 but for mobx4.


davefeucht:
Not sure what to do about the failing checks, let me know if they matter and if I need to do something to make them pass :)

FredyC:
Build fails with docs which do not exists in this branch. Don't worry about it.

In fact, I am waiting for your work to get merged and then the whole mobx4-master will be gone ... #2248

PR #2261, opened 2020-01-23
Fix broken mixed-versions test

FredyC:
As part of combining V4 as V5 into a "master" branch, I've found out that these tests weren't part of the CI before and are failing now. Hopefully, someone can make some sense of it and come up with a fix.


FredyC:
I suppose we can ditch this with the upcoming V6.

PR #2262, opened 2020-01-24
Define action name to be as the function name

nadavkaner:

PR checklist:

* [x] Added unit tests
* [x] Updated changelog
* [x] Updated `/docs`. For new functionality, at least `API.md` should be updated

### DIFF
--- a/src/v5/core/action.ts
+++ b/src/v5/core/action.ts
@@ -17,7 +17,6 @@ import { allowStateReadsStart, allowStateReadsEnd } from "./derivation"
 // mobx versions
 let currentActionId = 0
 let nextActionId = 1
-const isProduction = process.env.NODE_ENV === "production"
 const functionNameDescriptor = Object.getOwnPropertyDescriptor(() => {}, "name")
 const isFunctionNameConfigurable = functionNameDescriptor && functionNameDescriptor.configurable
 
@@ -26,7 +25,7 @@ export interface IAction {
 }
 
 export function createAction(actionName: string, fn: Function, ref?: Object): Function & IAction {
-    if (!isProduction) {
+    if (process.env.NODE_ENV !== "production") {
         invariant(typeof fn === "function", "`action` can only be invoked on functions")
         if (typeof actionName !== "string" || !actionName)
             fail(`actions should have valid names, got: '${actionName}'`)
@@ -35,8 +34,10 @@ export function createAction(actionName: string, fn: Function, ref?: Object): Fu
         return executeAction(actionName, fn, ref || this, arguments)
     }
     ;(res as any).isMobxAction = true
-    if (!isProduction && isFunctionNameConfigurable) {
-        Object.defineProperty(res, "name", { value: actionName })
+    if (process.env.NODE_ENV !== "production") {
+        if (isFunctionNameConfigurable) {
+            Object.defineProperty(res, "name", { value: actionName })
+        }
     }
     return res as any
 }
--- a/src/v5/core/action.ts
+++ b/src/v5/core/action.ts
@@ -17,13 +17,16 @@ import { allowStateReadsStart, allowStateReadsEnd } from "./derivation"
 // mobx versions
 let currentActionId = 0
 let nextActionId = 1
+const isProduction = process.env.NODE_ENV === "production"
+const functionNameDescriptor = Object.getOwnPropertyDescriptor(() => {}, "name")
+const isFunctionNameConfigurable = functionNameDescriptor && functionNameDescriptor.configurable
 
 export interface IAction {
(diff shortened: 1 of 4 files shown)


nadavkaner:
Do I need to make these changes also in v4 folder?

test/v5/base/action.js
```diff
@@ -561,3 +561,18 @@ test("out of order startAction / endAction", () => {
 
     mobx._endAction(a1)
 })
+
+test.only("given actionName, the action function name should be defined as the actionName", () => {
```
FredyC: Probably an oversight, but removes these `.only` flag please :)

FredyC:
> Do I need to make these changes also in v4 folder?

Once the `v5` is polished, the `v4` will be needed as well for sure.

src/v5/core/action.ts
```diff
@@ -17,13 +17,16 @@ import { allowStateReadsStart, allowStateReadsEnd } from "./derivation"
...
+const functionNameDescriptor = Object.getOwnPropertyDescriptor(() => {}, "name")
+const isFunctionNameConfigurable = functionNameDescriptor && functionNameDescriptor.configurable
 
 export interface IAction {
     isMobxAction: boolean
 }
 
 export function createAction(actionName: string, fn: Function, ref?: Object): Function & IAction {
-    if (process.env.NODE_ENV !== "production") {
+    if (!isProduction) {
```
mweststrate: best keep the condition here, otherwise it kills the dead code elimination when making production builds

FredyC: Yea, eventually I want to use TSDX here as well so we could stick to a simple `__DEV__` condition instead of this wordy thing, but for now, it needs to be like this.

test/v5/base/action.js
nadavkaner: Oops :)

src/v5/core/action.ts
nadavkaner: @FredyC I make the changes, let me know if this is your intent.

nadavkaner:
How I can change functions `name` to be non-configurable inside the test?

FredyC:
I suppose you cannot make that particular property non-configurable, but let's "cheat" and use `Object.seal` which makes everything ... well, sealed :)

nadavkaner:
With `Object.seal` I can seal only specific function instance and not every function instance that being created which this is what I need in this case, am I missing something here?

FredyC:
You not missing anything, I haven't realized how is the check done. I suppose it would have to be enough. Complicating the code further for the sake of tests is probably not worth it.

nadavkaner:
I agree :)

FredyC:
Oh, btw, next time please don't tick the box that you have updated changelog. Clearly you did not and I forgot to check. Nevermind now, I will add it there directly

elektronik2k5:
This is awesome! BTW, why not do this in production mode too?

FredyC:
@elektronik2k5 That would work only if the action name is passed as a string in the first arg. Function names are usually minified. It would be inconsistent. Besides, this is intended for debugging purposes mostly.

elektronik2k5:
@FredyC, digging a bit further into the code, I see it isn't that simple. I understand that and am not challenging it any further right now - and definitely not as part of this PR.

However, I still want to address some of your assertions and perhaps we can improve it in the future:

> Function names are usually minified.

You're right: function names are usually mangled. Practically always in browsers and often in node packages which go through a build/compilation step.

However, that is not always the case:
 - I always configured the minifiers in my projects to not mangle function names cause I value readable stack traces in production much more than the negligible size decrease which mangling produces. You can have minification and keep readable names thanks to gzip/brotli. And it doesn't negate all other minification tricks. :)
 - Thanks to all modern browsers supporting ESM, dynamic `import()` hitting stage 4, HTTP >= 2 and Brotli we have things like https://www.snowpack.dev/, where we're able to go back to working like once: via `<script>` tags and without bundling. It is a valid approach and a breath of fresh air in light of our currently tooling-heavy workflows.
 - In node, there's absolutely no need to minify anything, and we know mobx is used in node too. 
 - Due to the fact that every application's target environment is different and package sources are written in different languages using different standards, there's talk of reversing the current standard of shipping ES5 packages and instead:
    - Shipping different target environment bundles: ES5, ES2015 without ESM, ESM + latest. Some packages already do this and bundler support is in place.
    - Ignoring bundled package artifacts altogether and having the bundler take care of compilation, polyfilling, minification, etc of raw source code. React Native does this, for example.

> It would be inconsistent.

Only if we let it be inconsistent.

> this is intended for debugging purposes mostly

In my view, debugging isn't a nice to have, but a first class feature of any tool/package I consider using.

PR #2265, opened 2020-01-27
Fix #2264

### LINKED ISSUE #2264: [5.15.3] Error "process is not defined" in production if `node` is set to `false` in webpack config
If webpack doesn't mock `process` then it only replaces exact strings defined in `DefinePlugin` or `EnvironmentPlugin` etc, so `process.env.IGNORE_MOBX_MINIFY_WARNING` doesn't get replaced and causes a "process is not defined" error in production. This is the offending line.

https://github.com/mobxjs/mobx/blob/d5488a180a34701ed90c4a25ccecef3619a42ea8/src/v5/mobx.ts#L43

Workarounds on user's side for now are:
- roll back to 5.15.2
- add `IGNORE_MOBX_MINIFY_WARNING` to Define/Environment plugin in webpack config

le0nik:


src/v4/mobx.ts
```diff
@@ -35,7 +35,7 @@ try {
     if (
         testCodeMinification.name !== "testCodeMinification" &&
         process.env.NODE_ENV !== "production" &&
-        process.env.IGNORE_MOBX_MINIFY_WARNING !== "true"
+        typeof process !== 'undefined' && process.env.IGNORE_MOBX_MINIFY_WARNING !== "true"
```
mweststrate: let's put it as first or second condition, so that the other one can't died either (there have been reports of that in the past)

FredyC:
I wonder why the trick for defining a shim for `process` does not work for you.

https://github.com/mobxjs/mobx/blob/d5488a180a34701ed90c4a25ccecef3619a42ea8/src/v5/mobx.ts#L28

Not that your fix is wrong, but I would like to know more about your setup and why is it a problem in 5.15.3 but not 5.15.2.

**5.15.2 mobx.js**
image

**5.15.3 mobx.js**
image

src/v4/mobx.ts
FredyC: Not needed, the first one is correctly replaced in latest version, see pictures in my comment.

FredyC:
Btw, regarding why there is a `development`, it's because `mobx.js` is a development bundle. For production there is a [`mobx.min.js`](https://unpkg.com/mobx@5.15.3/lib/mobx.min.js) which has only a single occurrence of `process` for that shim. Everything else is eliminated as dead code as it should be.

The problem though is that `main` field in `package.json` refers to `mobx.js`, it should probably use a minified version instead.

le0nik:
@FredyC 

| I wonder why the trick for defining a shim for process does not work for you.
It doesn't work anywhere.

mobx

As you can see pre-built mobx.module.js(and other bundles) have `process.env.NODE_ENV` replaced inside the try/catch block. So entire `try` block is always empty and is stripped out when minified.

You can check yourself by looking through the bundles here https://unpkg.com/browse/mobx@5.15.3/lib/

FredyC:
Ah I see, that makes total sense now, haven't noticed that at a first glance.

Anyway, I think the shim shouldn't be required at all. For a regular `NODE_ENV` it will be replaced and for that specific runtime env, adding `typeof process` makes sense for sure. So I am going to merge this as it won't harm anything. Thanks.

mweststrate:
See also #2266 

I think we need to take a closer look into this. The problem is that fields like `main`, `next`, `module` etc describe the _module_ strategy, but nothing about prod versus dev. So that is why we always have issues on either side if we already subtitute those environment variables; bundlers don't pick different _entry_ points based on the kind of build (dev vs prod) they are making. So I think we should live this as it were, leave the responsibility to the bundler of substituting, and just make sure that without substitution nothing breaks. I'm a bit swamped atm due to a meetup,  bit I can take a further look  on wednesday

PR #2267, opened 2020-01-27
Handle process.env correctly in build

### LINKED ISSUE #2266: False warning message starting from version 5.15.3 (`you are running a minified build, but 'process.end.NODE_ENV'...`)
* [x] Issue:
  * [x] Did you check this issue wasn't filed before?
  * [x] Elaborate on your issue. What behavior did you expect?
  * [x] State the versions of MobX and relevant libraries. Which browser / node / ... version?

## The Problem

On version `mobx@5.15.3` the bundle `mobx.module.js` throws an error which is a false positive. It says that the bundler was not configured for production which in fact it was.

> [mobx] you are running a minified build, but 'process.end.NODE_ENV' was not set to 'production' in your bundler. This results in an unnecessarily large and slow bundle

This is a degradation from `mobx@5.15.2` where there was no such message.

## Explanation

Digging into the build process of mobx i've found [this change](https://github.com/mobxjs/mobx/commit/582a51691ebdbb5b8cd628ee286919a6dafc480c#diff-ea1c4abd09127cfb80f898b2bf12b0a4R72) which did the following replacement

```diff
- process.env.NODE_ENV !== "production"
+ "development" !== "production"
``` 

https://github.com/mobxjs/mobx/blob/d5488a180a34701ed90c4a25ccecef3619a42ea8/src/v5/mobx.ts#L42

It makes mobx send this message in case the code passed minification regardless of the `process.env.NODE_ENV` that was used during build time. Sending a false message to the user.

## Possible Solution

While the replacement is good for optimizing production bundles, for development it is not critical. Removing [this line](https://github.com/mobxjs/mobx/commit/582a51691ebdbb5b8cd628ee286919a6dafc480c#diff-ea1c4abd09127cfb80f898b2bf12b0a4R72) should remove this warning. If we would like to optimize the development bundle we can later on see how we can do it without sending the false warning.

FredyC:
Fixes #2266 #2264 

Here is the build output it creates, looks ok to me, but I could use a second pair of eyes.

[v5.zip](https://github.com/mobxjs/mobx/files/4117954/v5.zip)

cc @mweststrate @ranyitz @le0nik

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,11 +2,11 @@
   "name": "mobx",
   "version": "0.15.3",
   "description": "Simple, scalable state management.",
-  "main": "lib/mobx.js",
+  "main": "lib/index.js",
   "umd:main": "lib/mobx.umd.js",
   "module": "lib/mobx.module.js",
   "browser": {
-    "./lib/mobx.js": "./lib/mobx.js",
+    "./lib/mobx.js": "./lib/mobx.min.js",
     "./lib/mobx.module.js": "./lib/mobx.module.js"
   },
   "unpkg": "lib/mobx.umd.min.js",
@@ -29,6 +29,7 @@
     "_prepublish": "yarn small-build",
     "quick-build": "tsc --pretty",
     "small-build": "node scripts/build.js",
+    "build": "node scripts/build.js",
     "release": "node scripts/publish.js",
     "publish-script": "yarn release",
     "docs:build": "yarn --cwd website build",
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -55,23 +55,16 @@ function runTypeScriptBuild(outDir, version, target, declarations) {
     }
 }
 
-async function generateBundledModule(inputPath, version, outputFile, format, env = "development") {
+async function generateBundledModule(inputPath, version, outputFile, format, env) {
     console.log(`Generating ${version} ${outputFile} ${env} bundle.`)
+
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "_prepublish": "yarn small-build",
     "quick-build": "tsc --pretty",
     "small-build": "node scripts/build.js",
+    "build": "node scripts/build.js",
```
ranyitz: This the same as `"small-build"` I would remove one of them

scripts/build.js
```diff
@@ -121,6 +114,19 @@ function writePackage(versionPath, version) {
     )
 }
 
+function writeIndex(versionPath) {
+    fs.writeFileSync(
+        path.resolve(versionPath, "lib", "index.js"),
```
ranyitz: This is not critical, but IMHO it's better to write this file somewhere in the codebase and just copy it as part of the build process. It will make sure that we're not making any mistake and makes the code a bit clearer.

le0nik:
Can confirm that it also solves #2264

package.json
FredyC: I want to standardize to `build` but not to remove `small-build` in case someone still uses that for something.

scripts/build.js
FredyC: Yea, was thinking about it but did want to come up with some artificial filename or folder where to keep it. It's such a small thing it's ok.

FredyC:
Thanks, guys. I will merge and release tomorrow, stick to X.15.2 till then, please.

PR #2275, opened 2020-02-10
Fix map replace for v4, fixes #2274

mweststrate:

PR #2277, opened 2020-02-10
ObservableMap - replace _keys array with _keysAtom

urugator:


src/types/observablemap.ts
```diff
@@ -261,16 +260,18 @@ export class ObservableMap<K = any, V = any>
...
     keys(): IterableIterator<K> {
-        return (this._keys[iteratorSymbol()] as any)()
+        this._keysAtom.reportObserved()
+        return this._data.keys()
     }
 
     values(): IterableIterator<V> {
         const self = this
         let nextIndex = 0
+        const keys = Array.from(this.keys())
```
mweststrate: use `iteratorToArray` as well here for consistency?

src/types/observablemap.ts
```diff
@@ -83,9 +82,7 @@ export class ObservableMap<K = any, V = any>
     $mobx = ObservableMapMarker
     private _data: Map<K, ObservableValue<V>>
     private _hasMap: Map<K, ObservableValue<boolean>> // hasMap, not hashMap >-).
-    private _keys: IObservableArray<K> = <any>(
-        new ObservableArray(undefined, referenceEnhancer, `${this.name}.keys()`, true)
-    )
+    private _keysAtom = createAtom(`${this.name}.keys()`)
```
mweststrate: this is gonna be a nice memory saver! 💯

src/types/observablemap.ts
```diff
@@ -279,10 +280,11 @@ export class ObservableMap<K = any, V = any>
     entries(): IterableIterator<IMapEntry<K, V>> {
         const self = this
         let nextIndex = 0
+        const keys = Array.from(this.keys())
```
mweststrate: use iteratorToArray as well here for consistency?

src/types/observablemap.ts
```diff
@@ -294,7 +296,7 @@ export class ObservableMap<K = any, V = any>
     }
 
     forEach(callback: (value: V, key: K, object: Map<K, V>) => void, thisArg?) {
-        this._keys.forEach(key => callback.call(thisArg, this.get(key), key, this))
+        for (const [key, value] of this) callback.call(thisArg, value, key, this)
```
mweststrate: Let's double check that this results in sane code for ES5, I think I avoided using `for..of` loops in this code base, due to issues in the past on IE 11, not entirely sure about that anymore

src/types/observablemap.ts
```diff
@@ -320,45 +322,40 @@ export class ObservableMap<K = any, V = any>
     clear() {
         transaction(() => {
             untracked(() => {
-                this._keys.slice().forEach(key => this.delete(key))
+                for (const key of this.keys()) this.delete(key)
```
mweststrate: See above comment

test/base/map.js
```diff
@@ -573,6 +573,14 @@ test("issue 940, should not be possible to change maps outside strict mode", ()
             m.set("x", 1)
```
mweststrate: Would be great if we added the tests from #2275, #2255, and missing v5 map tests here as well

src/types/observablemap.ts
```diff
@@ -320,45 +322,40 @@ export class ObservableMap<K = any, V = any>
...
             })
         })
     }
 
     replace(values: ObservableMap<K, V> | IKeyValueMap<V> | any): ObservableMap<K, V> {
         transaction(() => {
             const replacementMap = convertToMap(values)
-            const oldKeys = this._keys
+            const orderedData = new Map()
+            const oldKeys = Array.from(this.keys())
```
mweststrate: See previous comment, IE doesn't have Array.from, so that would be problematic for mobx 4

src/types/observablemap.ts
```diff
@@ -368,9 +365,10 @@ export class ObservableMap<K = any, V = any>
      */
     toPOJO(): IKeyValueMap<V> {
         const res: IKeyValueMap<V> = {}
-        this._keys.forEach(
-            key => (res[typeof key === "symbol" ? <any>key : stringifyKey(key)] = this.get(key)!)
-        )
+        for (const [key, value] of this) {
```
mweststrate: See previous

src/types/observablemap.ts
```diff
@@ -379,9 +377,7 @@ export class ObservableMap<K = any, V = any>
      * Note that the values migth still be observable. For a deep clone use mobx.toJS.
      */
     toJS(): Map<K, V> {
-        const res: Map<K, V> = new Map()
-        this._keys.forEach(key => res.set(key, this.get(key)!))
-        return res
+        return new Map(this)
```
mweststrate: 💯

src/types/observablemap.ts
```diff
@@ -393,7 +389,9 @@ export class ObservableMap<K = any, V = any>
         return (
             this.name +
             "[{ " +
-            this._keys.map(key => `${stringifyKey(key)}: ${"" + this.get(key)}`).join(", ") +
+            Array.from(this.keys())
```
mweststrate: See previous

src/types/observablemap.ts
urugator: I think we can actually avoid the array here and reuse the iterator returned from keys().
However I first want it to match v5 as close as possible...
And I also want to take a look at how native Map handles modification of keys during iteration.

src/types/observablemap.ts
urugator: I think it relies on iteration protocol as well, will it work in IE?

mweststrate:
Yeah, iterators should be fine afaik

FredyC:
@urugator Why is this based on mobx4-master branch? We should probably delete that one to avoid confusion. I assume #2279 is a replacement for this PR?

PR #2283, opened 2020-02-13
Clarify flow usage in 'actions.md'

dwnste:
Hello! Although it may be obvious for some, but I think it makes sense to show that you need to bind context while using `flow`.

### DIFF
--- a/docs/best/actions.md
+++ b/docs/best/actions.md
@@ -204,7 +204,7 @@ class Store {
         } catch (error) {
             this.state = "error"
         }
-    })
+    }).bind(this)
 }
 ```
 


FredyC:
Thanks for the contribution, but this is not correct. Sure, it assumes the use of class properties which surprisingly is still a proposal, but see quick example that it indeed works.

https://codesandbox.io/s/fast-dust-fllcg (check the console)

Closing as this would introduce even more confusion :)

FredyC:
Although this really depends. As long as you are calling `store.fetchProjects()` it will be always correctly bound. However, in case, you would pass around `fetchProjects` reference only and call directly then it won't be bound.

I think it's generally bad practice to be passing methods only in the class-based world. If you are worried about doing `() => store.fetchProjects()` which creates an extra function, then my advice is to measure first before you start optimizing something like that.

PR #2287, opened 2020-02-15
fix: map.replace #2253 (interceptors) #1980 (key order V5) 

### LINKED ISSUE #1980: ObservableMap.replace() - breaks entities order!

### LINKED ISSUE #2253: Change of behavior in ObservableMap.replace when using intercept in Mobx v4

urugator:
Fixed #2253 
Fixed #1980 for V5 (it was fixed for V4)
Synced V4/V5 map tests.
Fixed deprecation warnings in V4 test.
Removed invalid use of `Object.entries` in V4 (`convertToMap`)

### DIFF
--- a/src/v4/utils/utils.ts
+++ b/src/v4/utils/utils.ts
@@ -1,11 +1,4 @@
-import {
-    ObservableMap,
-    globalState,
-    IObservableArray,
-    isObservableArray,
-    IKeyValueMap,
-    isObservableMap
-} from "../internal"
+import { globalState, IObservableArray, isObservableArray, isObservableMap } from "../internal"
 
 declare const Symbol: any
 
--- a/src/v4/utils/utils.ts
+++ b/src/v4/utils/utils.ts
@@ -108,13 +108,17 @@ export function isPlainObject(value) {
     return proto === Object.prototype || proto === null
 }
 
-export function convertToMap(dataStructure) {
+export function convertToMap(dataStructure: any): Map<any, any> {
     if (isES6Map(dataStructure) || isObservableMap(dataStructure)) {
         return dataStructure
     } else if (Array.isArray(dataStructure)) {
         return new Map(dataStructure)
     } else if (isPlainObject(dataStructure)) {
-        return new Map(Object.entries(dataStructure))
+        const map = new Map()
+        for (const key in dataStructure) {
+            map.set(key, dataStructure[key])
+        }
+        return map
     } else {
         return fail(`Cannot convert to map from '${dataStructure}'`)
     }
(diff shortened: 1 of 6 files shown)


src/v4/utils/utils.ts
```diff
@@ -108,13 +101,17 @@ export function isPlainObject(value) {
...
-export function convertToMap(dataStructure) {
+export function convertToMap(dataStructure: any): Map<any, any> {
     if (isES6Map(dataStructure) || isObservableMap(dataStructure)) {
         return dataStructure
     } else if (Array.isArray(dataStructure)) {
         return new Map(dataStructure)
     } else if (isPlainObject(dataStructure)) {
-        return new Map(Object.entries(dataStructure))
+        const map = new Map()
+        for (const key in dataStructure) {
```
mweststrate: use `forOf` here?

urugator: `forOf` is just a workaround for `Iterator`s - converting them to array is waste of resources and using `while` is PITA ... the function doesn't even accept/work with array/map/object.
Also I would prefer to keep this in sync with V5 (which intentionally doesn't have `forOf` util).

urugator: Basically the idea is: If you use `for of` in V5, use native `.forEach` (array/map/set) or `forOf` for iterators in V4... `forOf/for of` for `object` doesn't make sense...

Issue #2282, opened 2020-02-12
Getting Started - Cannot run demo code


berrybell:
I have a:
2. [x] Issue: 

Whenever I am trying to run any code on the [https://mobx.js.org/getting-started](Getting Started) page I am greeted by `Code editors not yet loaded` message in the `console.log` section.

I see the following error messages in my console:
```
Failed loading 'https://mobx.js.org/assets/getting-started-assets/javascripts/codemirror/lib/codemirror.js'.
ReferenceError: CodeMirror is not defined
```

mweststrate:
Fix has been published. Thanks for reporting, and @lgordey thanks for the fix!

PR #2289, opened 2020-02-16
Fix Getting Started #2282: add missing codemirror files

lgordey:
I’m not sure how correct this decision is. 😅
But I noticed issue #2282  and decided to help return the lost packages.

So both files are taken from https://github.com/codemirror/CodeMirror

### DIFF
--- a/docs/assets/getting-started-assets/javascripts/codemirror/lib/codemirror.css
+++ b/docs/assets/getting-started-assets/javascripts/codemirror/lib/codemirror.css
@@ -1,0 +1,350 @@
+/* BASICS */
+
+.CodeMirror {
+    /* Set height, width, borders, and global font properties here */
+    font-family: monospace;
+    height: 300px;
+    color: black;
+    direction: ltr;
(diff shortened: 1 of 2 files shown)


FredyC:
Thanks a lot for contribution.

To be honest, I am not exactly sure where is the issue happening. I don't see any runnable code in the docs.

Either way, the preview of this PR is available here: https://deploy-preview-2289--mobx-docs.netlify.com/readme#getting-started

@berrybell can you please verify if the issue is fixed in that preview?

lgordey:
@FredyC 

This PR fixes this page https://deploy-preview-2289--mobx-docs.netlify.com/getting-started
And the original one is https://mobx.js.org/getting-started

If you open the developer tools and look in the console, you will see errors about lost files.

FredyC:
Alright, I see now. I was looking at the wrong page. It's so confusing documentation :)

Looks like it has indeed fixed the issue, so let's roll with it. Thanks again.

PR #2291, opened 2020-02-18
Github action to run prettier on master

FredyC:
I've been checking out GitHub actions lately and this is a first small attempt to utilize that.

Running prettier for each commit with husky is often fairly slow. The outcome of this change is that prettier will be executed automatically after a push to the `master` branch. Usually, that's after merging PR, but also some direct quick commits. With correctly setup IDE which runs Prettier during coding, no commit will be made if there is nothing "unpretty".

You might ask why not to run it inside the PR. It feels like a logical step, but it means that more complex PR would be struggling with conflicts where code would be prettified after each commit and if the author does not pull before doing follow up changes.

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -1,0 +1,19 @@
+name: "Run Prettier on master"
+on:
+    push:
+        branches: [master]
+
+jobs:
+    prettier:
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Cloning the repository
+              uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


FredyC:
Just found out

```
Can't find Husky, skipping prepare-commit-msg hook
You can reinstall it using 'npm install husky --save-dev' or delete this hook
```

Husky won't remove itself from Git hooks and it has to be done manually. That means going to `.git/hooks` directory and unless you have some custom specific hooks there, you can basically delete everything in there.

PR #2294, opened 2020-02-19
Upgrade bunch of dev deps

Bnaya:
Upgrade bunch of dev deps


package.json
```diff
@@ -58,41 +58,41 @@
...
+    "@babel/plugin-proposal-decorators": "^7.8.3",
+    "@babel/plugin-transform-runtime": "^7.8.3",
+    "@babel/preset-env": "^7.8.4",
+    "@babel/runtime": "^7.8.4",
+    "@rollup/plugin-node-resolve": "^7.1.1",
+    "@size-limit/preset-big-lib": "^4.0.2",
     "@types/jest": "^24.0.11",
-    "@types/node": "^11.11.3",
-    "@typescript-eslint/eslint-plugin": "^1.4.2",
+    "@types/node": "^12",
```
Bnaya: 12 is LTS

package.json
```diff
@@ -58,41 +58,42 @@
...
+    "rollup": "^1.31.1",
+    "rollup-plugin-filesize": "^6.2.1",
     "rollup-plugin-replace": "^2.1.0",
-    "rollup-plugin-terser": "^4.0.4",
-    "semver": "^7.1.1",
+    "rollup-plugin-terser": "^5.2.0",
+    "semver": "^7.1.3",
     "serializr": "^1.5.1",
     "shelljs": "^0.8.3",
+    "size-limit": "^4.0.2",
```
FredyC: Not a big deal, but the `@size-limit/preset-big-lib` has it as dependency already and actually warns about it when you run it.

Bnaya: it was moved to peerDeps there

Bnaya: https://github.com/ai/size-limit/blob/master/packages/preset-big-lib/package.json#L15-L23

PR #2311, opened 2020-03-13
Proof of concept, microtasks to batch batches.

sliftist:
(This isn't a serious pull request, just a proof of concept).

`mobx` uses actions to batch observable changes, but I was thinking it could have an option to use microtasks to batch changes, at least when changes are made outside of an action.

I tried it out, and it seems to work, I deployed a highly experimental build at [slift/mobx](https://github.com/sliftist/mobx#0548ddc4179aa87452fd46f621ebf114f854b7da) (`yarn add sliftist/mobx#0548ddc4179aa87452fd46f621ebf114f854b7da`), and am going to start using that branch in projects to get additional testing.

My implementation isn't meant to be a serious implementation, it basically just does this:
```diff
public reportChanged() {
    startBatch()
    propagateChanged(this)
    endBatch()
+   if (globalState.queueReportChangedEndBatchAsMicrotask) {
+       Promise.resolve().then(() => {
+           endBatch()
+       })
+   } else {
        endBatch()
+   }
}
```

If this is something `mobx` is interested in pursuing, I would need to make changes to ensure microtasks are only used outside of actions, instead of in all reportChanged calls.

Batching changes via a microtask can yield a large performance increase. For example:

```tsx
@observable lookup = {};
constructor() {
    this.runLoop();
    autorun(this.reactionHandler.bind(this));
}
async runLoop() {
    while(true) {
        await new Promise((resolve) => setTimeout(resolve, 1000));
        for(let key in this.lookup) {
            this.lookup[key]++;
        }
        this.lookup[Date.now()] = 0;
    }
}
reactionHandler() {
    let sum = 0;
    for(let key in this.lookup) {
        sum += this.lookup[key];
    }
    console.log("Keys", Object.keys(this.lookup).length, "Sum", sum);
}
```

This case scales `O(N^2)` without batching, due to every lookup change triggering the reactionHandler, but `O(N)` with batching.

I'm unsure on the full ramifications of this change; if there are parts of `mobx` that are incompatible with this approach, or if instead it would be a relatively stable change (behind a flag), so guidance would be greatly appreciated!

Related:
- [is @action really necessary?](https://github.com/mobxjs/mobx-react/issues/505)
- [Opt-in variant of batched updates](https://github.com/mobxjs/mobx-react/pull/787)

### DIFF
--- a/src/v5/api/configure.ts
+++ b/src/v5/api/configure.ts
@@ -7,7 +7,7 @@ import {
 } from "../internal"
 
 export function configure(options: {
-    enforceActions?: boolean | "strict" | "never" | "always" | "observed"
+    enforceActions?: boolean | "strict" | "never" | "always" | "observed" | "useMicrotasks"
     computedRequiresReaction?: boolean
     /**
      * (Experimental)
@@ -48,6 +48,7 @@ export function configure(options: {
                 ea = true
                 break
             case false:
+            case "useMicrotasks":
             case "never":
                 ea = false
                 break
@@ -57,11 +58,12 @@ export function configure(options: {
                 break
             default:
                 fail(
-                    `Invalid value for 'enforceActions': '${enforceActions}', expected 'never', 'always' or 'observed'`
+                    `Invalid value for 'enforceActions': '${enforceActions}', expected 'never', 'always', 'observed' or 'useMicrotasks'`
                 )
         }
         globalState.enforceActions = ea
         globalState.allowStateChanges = ea === true || ea === "strict" ? false : true
+        globalState.queueReportChangedEndBatchAsMicrotask = true
     }
     if (computedRequiresReaction !== undefined) {
         globalState.computedRequiresReaction = !!computedRequiresReaction
--- a/src/v5/api/configure.ts
+++ b/src/v5/api/configure.ts
@@ -49,6 +49,7 @@ export function configure(options: {
                 break
             case false:
             case "useMicrotasks":
+                globalState.queueReportChangedEndBatchAsMicrotask = true
             case "never":
                 ea = false
                 break
@@ -63,7 +64,6 @@ export function configure(options: {
         }
         globalState.enforceActions = ea
         globalState.allowStateChanges = ea === true || ea === "strict" ? false : true
-        globalState.queueReportChangedEndBatchAsMicrotask = true
     }
     if (computedRequiresReaction !== undefined) {
         globalState.computedRequiresReaction = !!computedRequiresReaction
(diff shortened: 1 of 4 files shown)


mweststrate:
As documented one of the design principles is that MobX is synchronous in
all circumstances, which makes it behavior very predictable and much easier
to debug (a change causing a reaction is always in the same call stack for
example). Doing things async is a valid design choice with it's own merits
as you mentioned. However, in a product like MobX that is already shipped
in thousands of products, such a significant fundamental semantic change
would result in a totally unmanageable support load so it is definitely not
an option to change that at this point imho, even if (which I wouldn't
agree upon) the benefits would outweigh the costs in case this was a
greenfield library.

For some background:
https://hackernoon.com/the-fundamental-principles-behind-mobx-7a725f71f3e8

mweststrate:
also note that reactions support having a custom scheduler, so that you can
batch individual reactions in the way you want, so what you want can
basically be achieved by creating a small userland utility around reaction,
as far as I can see.

mweststrate:
Beyond that, creating abstractions to better support (and thereby
encourage) anti-patterns (" at least when changes are made outside of an
action") isn't something I would aim for in the first place. If you are
concerned about perf, rather enable strict mode first.

spion:
@mweststrate One of the things I've been thinking about is that strict mode don't fully solve the problem. You could have an action in a class, then loop over a list of those classes calling their actions. Strict mode will not complain, but reactions will definitely be slow.

Some sort of warning in development mode where > 10 (or N=?) separate actions are detected within one microtask loop would definitely be useful for debugging these cases.

sliftist:
@mweststrate , ah that makes sense, keeping side effects in the same call stack is indeed very useful. I didn't realize that was specifically a design principle of mobx, in which case I'll just close this pull request. Thanks!

For anyone that wants this kind of behavior (as in, batching all reactions), the userland utility to accomplish this is:
```tsx
configure({ reactionScheduler: (callback) => Promise.resolve().then(callback) })
```
However as @mweststrate mentioned, this will detach the call stack from the original action, and so debugging what line of code triggered a reaction will be much more difficult. Devtools will give the async call stack, so you will be able to find the original action, however the programmatic stack trace (ie, the one you send to the server) from `new Error()` will only contain the synchronous stack trace.

Also this impacts all actions, not just code that is not inside of an action, so existing actions will begin to trigger reactions differently (as in later).

@spion , I'm not 100% sure, but it appears that nested actions are 'flattened' to some degree. At least in my limited testing if your outer loop is inside of an action (or wrapped with runInAction) side effects are only triggered after the outer most action finishes. That doesn't help the case if the actions are triggered in an independent fashion though.

I agree with @spion that a warning when too many reactions are triggered within a single microtask loop would be very useful.

sliftist:
I created a pull request to add a warning if too many reactions are triggered with a microtask here: https://github.com/mobxjs/mobx/pull/2312 .

Issue #2288, opened 2020-02-15
[breaking change] Get rid of field initializers (and legacy decorators)


Kukkimonsuta:
Enabling `useDefineForClassFields` in TypeScript will prevent decorators from working (both transpiled and using `mobx.decorate`).

This flag will be enabled by default for `ESNext` once class fields go stage 4: https://github.com/microsoft/TypeScript/issues/34787

Possibly related to https://github.com/mobxjs/mobx/issues/1969

**Intended outcome:**

`autorun` using `objectState` is executed upon clicking on "increment" button.
`autorun` using `classState` is executed upon clicking on "increment" button.
`autorun` using `classStateNoDecorator` is executed upon clicking on "increment" button.

**Actual outcome:**

✅  `autorun` using `objectState` is executed upon clicking on "increment" button.
⛔️  `autorun` using `classState` is NOT executed upon clicking on "increment" button.
⛔️  `autorun` using `classStateNoDecorator` is NOT executed upon clicking on "increment" button.

**How to reproduce the issue:**

https://codesandbox.io/s/fragrant-frog-x2487

**Versions**

TypeScript 3.7+
MobX - all tested versions (4.x, 5.x)

FredyC:
Thank you for a nice heads up. So basically we can just be explicit and set that flag `false` to avoid that default change, right? Would you mind creating a PR for that?

Or are there some other implications of not using that flag?

FredyC:
Oh wait, I got that all wrong (kinda sleepy). This isn't about mobx building, but users running TS 3.8 will have this problem... Well that goes beyond my expertise why is it a problem.

Basically this...

```js
class State {
    constructor() {
        this.value = 0;
    }
}
```

Becomes...

```js
class State {
    constructor() {
        Object.defineProperty(this, "value", {
            enumerable: true,
            configurable: true,
            writable: true,
            value: 0
        });
    }
}
```

And MobX cannot handle that for some reason.

mweststrate:
Nope, I protested heavily when TC 39 proposed this, for reasons like this,
but to no avail.... Field initializers will be non-interceptible once this
finalizes, so the assignments will have to move to the constructor instead

Kukkimonsuta:
So if I understand this correctly to support classes with this we will need to resort to something like:

```ts
import { observable, extendObservable } from 'mobx';

class State {
    constructor() {
        extendObservable(this, {
            value: this.value,
        });
    }

    value = 0;
}
```

or maybe

```ts
[...]

PR #2312, opened 2020-03-14
Warn if too many reactions are triggered inside of a single microtask loop

sliftist:
As suggested by @spion in #2311 , warns if too many reactions occur in a single microtask loop.

Hopefully this helps the learning curve with mobx, as in https://github.com/mobxjs/mobx-react/issues/505 sometimes programmers may not see a reason to use actions, as mobx works without actions.

However as an application grows and the number of reactions grows, the application may run into performance issues from lack of batching, at which point this warning will fire.

### Code change checklist

-   [x] Updated changelog

### DIFF
--- a/src/v4/core/reaction.ts
+++ b/src/v4/core/reaction.ts
@@ -211,6 +211,9 @@ export function onReactionError(handler: (error: any, derivation: IDerivation) =
  */
 const MAX_REACTION_ITERATIONS = 100
 
+const REACTIONS_IN_MICROTASK_LOOP_WARN_THRESHOLD = 20
+let curMicrotaskReactionCount = 0
+
 let reactionScheduler: (fn: () => void) => void = f => f()
 
 export function runReactions() {
@@ -223,6 +226,22 @@ function runReactionsHelper() {
     globalState.isRunningReactions = true
     const allReactions = globalState.pendingReactions
     let iterations = 0
+
+    // Calculate how many times the reaction helper is triggered
+    //  inside a single microtask, and warn if it is too many.
+    if (allReactions.length > 0) {
+        if (curMicrotaskReactionCount === 0) {
+            Promise.resolve().then(() => {
+                if (curMicrotaskReactionCount > REACTIONS_IN_MICROTASK_LOOP_WARN_THRESHOLD) {
+                    console.warn(
+                        `MOBX WARNING, a sigle microtask loop had ${curMicrotaskReactionCount} reactions (which are caused by changes to observables). Batch these actions with @action or runInAction to gain up to a ${curMicrotaskReactionCount}X performance boost!`
+                    )
+                }
+                curMicrotaskReactionCount = 0
+            })
+        }
+        curMicrotaskReactionCount++
+    }
 
     // While running reactions, new reactions might be triggered.
     // Hence we work with two variables and check whether
--- a/src/v4/core/reaction.ts
+++ b/src/v4/core/reaction.ts
@@ -212,7 +212,7 @@ export function onReactionError(handler: (error: any, derivation: IDerivation) =
 const MAX_REACTION_ITERATIONS = 100
 
 const REACTIONS_IN_MICROTASK_LOOP_WARN_THRESHOLD = 20
-let curMicrotaskReactionCount = 0
+let currentMicrotaskReactionCount = 0
 
 let reactionScheduler: (fn: () => void) => void = f => f()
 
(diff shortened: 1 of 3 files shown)


FredyC:
Thanks for the contribution, but can you please follow outlined checklist?

sliftist:
I didn't add any unit tests, but I don't think this change requires unit testing (as it just adds a warning), so I just marked that as done.

One thing to note, is that ironically enough the warning IS triggered by unit tests, and the performance test suite. I'm unsure on how this should be handled, as although it doesn't look good, it appears to be that running many reactions in a row without a microtask loop (so without user interaction) is almost always bad. Except of course, for tests.

src/v4/core/reaction.ts
```diff
@@ -211,6 +211,9 @@ export function onReactionError(handler: (error: any, derivation: IDerivation) =
  */
 const MAX_REACTION_ITERATIONS = 100
 
+const REACTIONS_IN_MICROTASK_LOOP_WARN_THRESHOLD = 20
+let curMicrotaskReactionCount = 0
```
FredyC: Let's keep a full name `currentMicrotaskReactionCount`, no need to be stingy :)

src/v4/core/reaction.ts
```diff
@@ -224,6 +227,22 @@ function runReactionsHelper() {
...
+            Promise.resolve().then(() => {
+                if (curMicrotaskReactionCount > REACTIONS_IN_MICROTASK_LOOP_WARN_THRESHOLD) {
+                    console.warn(
+                        `MOBX WARNING, a single microtask loop had ${curMicrotaskReactionCount} reactions (which are caused by changes to observables). Batch these actions with @action or runInAction to gain up to a ${curMicrotaskReactionCount}X performance boost!`
+                    )
+                }
+                curMicrotaskReactionCount = 0
+            })
+        }
+        curMicrotaskReactionCount++
```
FredyC: My personal nit, I like to see that number explicitly there.
```suggestion
        curMicrotaskReactionCount += 1
```

sliftist:
@FredyC , thanks for taking a look, I just pushed the style changes, and made it so it only checks when not in production.

sliftist:
Would there be opposition to setting a global flag that indicates if we are current in a test? I don't imagine there would be many cases when it would be used, but for example, in this case, the warning is indicating abnormal behavior, which isn't wrong because tests have rather abnormal behavior.

This warning is similar to how Chrome warns in the console if an event handler takes too long, because a slow handler is generally bad for user experience. However a slow event handler warning would not be relevant if you were running unit tests in the browser, as you might have tests that intentionally take a long time.

It just looks bad when one of the last lines of the performance tests is a warning.

urugator:
I would like to see a test case with strict mode enabled, not complaining, while warning being triggered.
I think it will also warn for long reaction chains (reaction leading to reaction leading to reaction etc...)?
Lots of reactions isn't problem by itself, but effects tend to be expensive. 
It's not necessarily an indicator of abnormal behavior per se.
We would need a config option for this, default being current behavior (no warning), the value can be a number of reactions. That should also settle the problem with tests.
I would also prefer this to be "experimental" until we find out there are no false positives.

sliftist:
Hi @urugator , by `not complaining, while warning being triggered` do you mean a test where some instrumentation function is called to indicate the warning would have fired, but without the warning being logged to the console?

I believe the `while (allReactions.length > 0)` loop will continue to run until all reaction chains are exhausted, but I could be wrong. So I believe `currentMicrotaskReactionCount` will only increment once no matter how long a reaction chain is, unless the chain is asynchronous, in which case the `currentMicrotaskReactionCount = 0` will run which will reset the count anyway.

It seems like this warning is directed at new users to mobx, who if they knew enough to turn the warning on, probably wouldn't need it. It could be useful in a codebase with many programmers to ensure new programmers use actions, in which case the warning will probably be enabled (hopefully).

urugator:
> by not complaining, while warning being triggered do you mean

Strict mode is already supposed to prevent you from forgetting actions. Another precaution would only make sense for cases that are not already covered by strict mode, so I think we should firstly identify these cases, in order to decide whether it makes sense to introduce another check.
So I mean a test with `configure({ enforceAction: "always" })`, that triggers only the newly introduced warning. Sorry if I confused you with the *strict mode*, it's a legacy term.

> I believe the while (allReactions.length > 0) loop will continue

Yea, I think you're correct

> who if they knew enough to turn the warning on, probably wouldn't need it

I understand, but personally I think we shouldn't start producing warning out of the blue in existing codebases, unless it always means an error, that has to be fixed. Others may feel differently about this.
If you're interested, [there is a proposal](https://github.com/mobxjs/mobx/issues/2288#issuecomment-595834171) (among others), which would hopefully make things more fool proof in the next major.

spion:
> Strict mode is already supposed to prevent you from forgetting actions. Another precaution would only make sense for cases that are not already covered by strict mode, so I think we should firstly identify these cases, in order to decide whether it makes sense to introduce another check.

Strict mode doesn't prevent you from forgetting actions. Here is an example

```typescript
class Counter {
  @observable private count = 0;
  @action increment() { this.count++; }
  get value() { return this.count; }
}

class Counters {
  @observable counters = new Array(10).fill(0).map(() => new Counter());

  incrementAll() {
    this.counters.forEach(c => c.increment());
  }
}
let cnts = new Counters();

autorun(() => {
  console.log("Total count", cnts.counters.reduce((acc, el) => acc + el.value, 0));
});
```

This will trigger the autorun 10 times: 

https://codesandbox.io/s/mobx-react-typescript-example-is75z

urugator:
@sliftist Could you add it as test?

Btw `configure({ observableRequiresReaction: true })` can also coincidentally reveal missing `action`, like in this example.

urugator:
Would it be correct to state:
It doesn't count the number of reactions, but number of batches.
If I use `enforceAction: "observed/always"`, then I always expect a single batch during microtask.
Therefore we could make it part of `enforceAction`, with a threshold of 2?

sliftist:
Hi @urugator , that sounds good! I'm going to be busy all this week, and maybe over the weekend, so if you're interesting I would have no problem with you taking over this pull request (by making a new one). Otherwise it might be a bit before there's I can update the code. Thanks!

sliftist:
Sorry for the long delay, I have been swamped with non-mobx related projects for the past few weeks.

After thinking about it more, I think the warning would actually be best accomplished with a custom reaction scheduler. Something like this (untested code):
```js
const REACTIONS_IN_MICROTASK_LOOP_WARN_THRESHOLD = 20;
let currentMicrotaskReactionCount = 0;
configure({
    reactionSchedule(f: () => void) {
        if (currentMicrotaskReactionCount === 0) {
            Promise.resolve().then(() => {
                if (currentMicrotaskReactionCount > REACTIONS_IN_MICROTASK_LOOP_WARN_THRESHOLD) {
                    console.warn(
                        `MOBX WARNING, a single microtask loop had ${currentMicrotaskReactionCount} reactions` +
                        ` (which are caused by changes to observables). Batch these actions with @action or ` +
                        `runInAction to gain up to a ${currentMicrotaskReactionCount}X performance boost!`
                    );
                }
                currentMicrotaskReactionCount = 0;
            });
        }
        currentMicrotaskReactionCount++;
        f();
    }
});
```

As it isn't very much code (a lot of it is just the warning message), I think it is almost as easy as manually specifying the warning, and keeps code that likely be seldomly used out of the main mobx repo.

Thanks for the quick replies though, everyone!

PR #2320, opened 2020-03-27
Converted the 4 most simple test files

gilamran:
### Code change checklist

-   [X] Added/updated unit tests
-   [ ] Updated changelog


FredyC:
Thanks. I think it would be probably best to keep v4 and v5 in sync. Most tests are copy & paste, so please do the same changes in the v4 folder.

gilamran:
Sure, will do

FredyC:
Thanks, keep em coming if you like :)

gilamran:
Cool, I'll move on to bigger and bigger tests.
While I'm in the process of conversion, I'll also upgrade the code to be more es6 "friendly"

Issue #356, opened 2016-06-23
Computed function invoked regardless of whether or not the underlying params have changed.


pdavidow:
The last test below is failing. The computed function (tickCount) seems to be invoked regardless of whether or not the underlying params have changed. 
````
import {computed, observable} from 'mobx';
import {lcm} from 'mathjs';
import test from 'tape';

class Beat {
    constructor (rh, lh) {
        this.observed = observable({
            rh: rh,
            lh: lh
        });

        this.count = 0;

        this.tickCount = computed(() => {
            this.count++;
            return lcm(this.observed.rh, this.observed.lh);
        });
    }
};

test('Beat tick count', (assert) => {
    const beat = new Beat(3, 1);

    let before = beat.count;
    beat.tickCount.get();
    let after = beat.count;
    assert.ok(before != after, 'tickCount should be called'); // pass

    before = beat.count;
    beat.tickCount.get();
    after = beat.count;
    assert.ok(before == after, 'tickCount should NOT be called'); // FAIL

    assert.end();
});
````

**Comment by mweststrate** [comment]
@pdavidow as long as a computed value is not used by a reaction, it is not memoized and so it just like a normal eager evaluating function. If you would would use the tickCount in an autorun this behavior will change and you won't see unnecessary computations. 

E.g. simply adding:

```
var disposer = autorun(() => beat.tickCount.get())
.... tests
disposer()
assert.end()
```

**Comment by pdavidow** [comment]
It would be nice to not have to use the autorun, but to rather to simply use something like 
````
this.observed = memoizedObservable({
            rh: rh,
            lh: lh
        });
````

**Comment by pdavidow** [comment]
It would be nice to not have to wrap with autorun, but rather to simply have something like
````
this.observed = **memoizedObservable**({
            rh: rh,
            lh: lh
        });
````

**Comment by mweststrate** [comment]
@pdavidow the reason MobX works this way is that as long as a computed value is not in use by some reaction, it can simply be ignored. MobX doesn't recompute it all, and the computation doesn't keep any other computation alive.

To keep the computation alive you can simply do the following:

```javascript
var c = computed(() => expr)
var suspend = c.observe(() => noop)
c.get() // will now memoize
suspend() // stops memoizing
```

you could even build a simple abstraction around that:
```javascript
function memoize(expr) {
  var c = computed(expr)
  var v;
  var d = c.observe(newValue => v = newValue)
  return {
    get: () => return v
    stop: () => d()
  }
}

[...]

Issue #2309, opened 2020-03-11
KeepAlive computed and onBecome(Un)Observed inconsistencies 


melnikov-s:
Given this code
```javascript

import {
  observable,
  autorun,
  computed,
  onBecomeUnobserved,
  onBecomeObserved
} from "mobx";


let count = 0;
const o = observable.box(1);
const c = computed(
  () => {
    return o.get() * 2;
  },
  { keepAlive: true }
);

onBecomeObserved(o, () => count++);
onBecomeUnobserved(o, () => count++);
onBecomeObserved(c, () => count++);
onBecomeUnobserved(c, () => count++);

const u = autorun(() => c.get());
u();


console.log(count);
```

When `keepAlive` is set to `true` the final count is `3` while when it's set to `false` the final count is `4`. It seems like `onBecomeUnobserved(c, () => count++);` is not being executed with `keepAlive: true`.  

This is with the latest version of 4.x and 5.x. Trying different versions produces different results for `keepAlive: true`. I've seen `0` , `2` , `3` and `4` being logged out depending on the version.

If `3` is the desired output here instead of `4` then why? Should it be documented?

FredyC:
Please follow the issue template and supply a reproduction.

melnikov-s:
Here it is: https://codesandbox.io/s/happy-dawn-vr6ke

urugator:
Modified your snippet like this:
```javascript
onBecomeObserved(o, () => console.log('o observed'));
onBecomeUnobserved(o, () => console.log('o unobserved'));
onBecomeObserved(c, () => console.log('c observed'));
onBecomeUnobserved(c, () => console.log('c unobserved'));
/* 
output
c observed 
o observed 
c unobserved 
*/
```
Which I think is expected behavior?...
The `o` won't become unobserved because the computed *keeps* observing `o` for changes - if `o` changes it will invalidate the computed's cache.
The version diferences are mostly caused by #1560 and #1799

melnikov-s:
Thanks, so that means observable dependencies for keep alive computed methods are considered "observed" even if not part of a reaction? Won't that effectively mean that once an observable is read by a keepAlive computed it will never be unobserved? Or does mobx provide a mechanism to clear out a keepAlive ? 

Even with the above heuristic I played around with the sandbox a bit more and still saw inconsistencies: 
https://codesandbox.io/s/crazy-ives-iwhv7
```javascript
import {
  observable,
  autorun,
  computed,
[...]

PR #2329, opened 2020-04-14
Add memoization and getter access documentation

kevzettler:
This PR adds some paragraphs around computed caching and memoization that seem to be a stumbling block for new users of Mobx. I have been using Mobx for a few years coming from other derived data layers and this was never explicitly clear to me.  The use of the `get` keyword in example code with intention of not using the method as a getter is unintuitive.

I had to ask for clarification about this on stackoverflow: https://stackoverflow.com/questions/61129530/mobx-how-to-cache-computed-values

and then pieced together tidbits from https://github.com/mobxjs/mobx/issues/356

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2329"></a>

### DIFF
--- a/docs/refguide/computed-decorator.md
+++ b/docs/refguide/computed-decorator.md
@@ -97,7 +97,20 @@ const orderLine = observable.object({
 ```
 
 ## Computed values are not getters
-The previous computed examples use the `get` keyword however, they generally should not be accessed directly as a getter. This can be a source of confusion to users new to Mobx from other derived cascading data layers like Reselect. As long as a computed value is not used by a reaction, it is not memoized and so it executes everytime it is accessed just like a normal eager evaluating function. This can cause performance degredation if a computed value is read high in a frequency loop like `requestAnimationFrame`.  MobX can be configured to report an error when computeds are being access directly by using the `computedRequiresReaction` value
+The previous computed examples use the `get` keyword however, they generally should not be accessed directly as a getter. This can be a source of confusion to users new to Mobx from other derived cascading data layers like Reselect. The following code demonsrates issue.
+
+```
+const Ol = new OrderLine(2.00)
+
+// don't do this.
+// avoid accessing Ol.total directly
+// it will recompute everytime.
+setInterval(() => {
+  console.log(Ol.total);
+}, 60);
+```
+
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for the docs! I think it would be good to mention the `keepAlive` flag as well (https://stackoverflow.com/a/61234019/1983583)

kevzettler:
@mweststrate I have added a `keepAlive` section in https://github.com/mobxjs/mobx/pull/2329#issuecomment-614149408

However i'm not sure I fully grok the functionality. When would you use keepAlive, instead of the `autorun` pattern I have documented? Seems like they both have the same behavior?

mweststrate:
@kevzettler it does the same as the autorun in your example, but it does it a lot more efficient (it can for example keep the computed alive, but defer computation until somebody actually reads the value, something the autorun can't do). 

The only case where `autorun` would be beneficial, is the case where you call the returned `disposer` to nicely clean up the computed value if it is no longer used (typically you would do that in a destructor of a class for example)

kevzettler:
thanks for that clarification @mweststrate I have added some sections expanding on KeepAlive and the conversation here

urugator:
> it does the same as the autorun in your example, but it does it a lot more efficient 

@mweststrate Have you noted https://github.com/mobxjs/mobx/issues/2309#issuecomment-598707584? It seems that currently the `keepAlive` won't keep the nested computeds "alive", while autorun/reaction will.

mweststrate:
@urugator yeah I did, but I am not sure how much of a problem that is of now, as the effect underlying computed values is not really specified. I think it'll be good address that, depending on how complicated that would be. If there is a simple fix that would be great, but I don't think I'll manage to look into it on the short term, but feel free to PR something!

docs/refguide/computed-decorator.md
```diff
@@ -96,6 +96,85 @@ const orderLine = observable.object({
...
+    constructor(price) {
+        this.price = price
+        // When computed total changes
+        // cache value to this.total
+        autorun(() => {
+            this.total = this.computedTotal
+        })
+    }
+
+    @computed get computeTotal() {
```
seivan, line 140: Typo? 
`/s/computeTotal/computedTotal/`?

mweststrate:
PRs rather than issues for docs please :)

FredyC:
Let's merge this and fix possible mistakes later. Thanks, @kevzettler for contribution.

Issue #1956, opened 2019-04-28
Reaction delay throttling instead of debouncing


Annihil:
Hi,

As stated in the documentation for `reaction`

> Options
> `delay`: Number in milliseconds that can be used to debounce the effect function. If zero (the default), no debouncing will happen.

The `delay` option should debounce the reaction, but in fact it throttles it.
Is the documentation incorrect or is it a bug?

mweststrate:
Feel free to update the docs. Note that a custom scheduler can be set to customize the behavior

Annihil:
I'm failing to use scheduler to achieve debouncing
```js
let id;
...
delay: 0,
scheduler: f => {
    clearTimeout(id);
    id = setTimeout(f, 1000);
}
```

It's still throttling and not clearing the timeout if a reaction has been done in those 1000ms timeout

mweststrate:
well, you would have to wrap the reaction function itself as well, and
introduce some state in the closure to take care of the throttling process.

Annihil:
That's what I ended up doing yes.  
I think it'd be great to have a `debounceDelay` option that provides this out of the box :)

koffmoff:
Agree that there is some confusion between throttling/debouncing. This is my take on achieving true debouncing.

```ts
function debounceEffect<T>(effect: (arg: T, r: IReactionPublic) => void, debounceMs: number) {
    let timer: NodeJS.Timeout
    return (arg: T, r: IReactionPublic) => {
        clearTimeout(timer)
        timer = setTimeout(() => effect(arg, r), debounceMs)
    }
}

// Use it like so:
reaction(
    () => this.searchTextField,
    debounceEffect(s => this.searchString = s, 1000)
)
```

FredyC:
Docs fixed in #2337

pkhodaveissi:
> ```ts
> function debounceEffect<T>(effect: (arg: T, r: IReactionPublic) => void, debounceMs: number) {
>     let timer: NodeJS.Timeout
>     return (arg: T, r: IReactionPublic) => {
>         clearTimeout(timer)
>         timer = setTimeout(() => effect(arg, r), debounceMs)
>     }
> }
> ```

@koffmoff I tried to use this function to run a function which returns Promise<void> instead of an assignment inside my reaction, but couldn't use your code, what am I missing?

koffmoff:
@pkhodaveissi should work. Msg your code + error to me if you want.

pkhodaveissi:
[...]

PR #2337, opened 2020-04-19
replace word debounce with throttle in the documentation

ivandotv:
### Code change checklist

-   [ ] Updated changelog

Default scheduler for `reaction` and `autorun` is actually **throttling** the execution even though in the documentation it says it's **debouncing**. There is a big difference between the two, and I think the documentation should reflect that.
There is an issue opened because of this confusion: https://github.com/mobxjs/mobx/issues/1956
[throttle vs debounce example reminder](http://demo.nimius.net/debounce_throttle/)

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2337"></a>

Fixes #2337

### DIFF
--- a/docs/refguide/api.md
+++ b/docs/refguide/api.md
@@ -368,7 +368,7 @@ Returns a disposer function to cancel the side effect.
 **options**
 
 -   **`name?: string`**: A name for easier identification and debugging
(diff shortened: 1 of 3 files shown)


FredyC:
Thanks!

PR #2343, opened 2020-04-25
Mobx6 - Add action name in prod too

### LINKED ISSUE #2263: Minification & debugging
_Originally posted by @elektronik2k5 in https://github.com/mobxjs/mobx/pull/2262#issuecomment-578387728_

---

digging a bit further into the code, I see it isn't that simple. I understand that and am not challenging it any further right now - and definitely not as part of this PR.

However, I still want to address some of your assertions and perhaps we can improve it in the future:

> Function names are usually minified.

You're right: function names are usually mangled. Practically always in browsers and often in node packages which go through a build/compilation step.

However, that is not always the case:
 - I always configured the minifiers in my projects to not mangle function names cause I value readable stack traces in production much more than the negligible size decrease which mangling produces. You can have minification and keep readable names thanks to gzip/brotli. And it doesn't negate all other minification tricks. :)
 - Thanks to all modern browsers supporting ESM, dynamic `import()` hitting stage 4, HTTP >= 2 and Brotli we have things like https://www.snowpack.dev/, where we're able to go back to working like once: via `<script>` tags and without bundling. It is a valid approach and a breath of fresh air in light of our currently tooling-heavy workflows.
 - In node, there's absolutely no need to minify anything, and we know mobx is used in node too. 
 - Due to the fact that every application's target environment is different and package sources are written in different languages using different standards, there's talk of reversing the current standard of shipping ES5 packages and instead:
    - Shipping different target environment bundles: ES5, ES2015 without ESM, ESM + latest. Some packages already do this and bundler support is in place.
    - Ignoring bundled package artifacts altogether and having the bundler take care of compilation, polyfilling, minification, etc of raw source code. React Native does this, for example.

> It would be inconsistent.

Only if we let it be inconsistent.

> this is intended for debugging purposes mostly

In my view, debugging isn't a nice to have, but a first class feature of any tool/package I consider using.

elektronik2k5:
## Changes:
- Define `action.name` unconditionally. Fixes #2263 
- Update readme to reflect correct scripts
- Fix `prettier` script to overwrite files
- Fix `.prettierignore` to ignore generated assets
- Run `prettier` on repo: 2 changed files

### Code change checklist

-   [x] Updated changelog

I'm not sure about how to test for perf regressions: I `yarn build && yarn perf` before and after with these results:

Before:
```shell
yarn test:performance proxy
Completed performance suite in 60.257 sec.
64.66user 1.62system 1:00.32elapsed 109%CPU (0avgtext+0avgdata 1092404maxresident)k
1136inputs+16outputs (1major+476988minor)pagefaults 0swaps

yarn test:performance legacy
Completed performance suite in 61.295 sec.
66.08user 1.66system 1:01.36elapsed 110%CPU (0avgtext+0avgdata 1033992maxresident)k
0inputs+16outputs (0major+477497minor)pagefaults 0swaps
```

After:
```shell
test/perf/index.js proxy
Completed performance suite in 61.343 sec.
64.80user 1.74system 1:01.41elapsed 108%CPU (0avgtext+0avgdata 1065372maxresident)k
0inputs+16outputs (0major+484730minor)pagefaults 0swaps

test/perf/index.js legacy
Completed performance suite in 60.269 sec.
65.21user 1.68system 1:00.33elapsed 110%CPU (0avgtext+0avgdata 1062128maxresident)k
0inputs+16outputs (0major+475698minor)pagefaults 0swaps
```
Looks pretty much the same.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2343"></a>


elektronik2k5:
Should I do a few more changes (is separate PRs)? Here's a list:
- Upgrade all deps
- Add spell checker
- Run tests against built package (not for perf)

FredyC:
Thanks for doing it. I am not sure you should even do other unrelated changes you did here. @mweststrate is working on that branch and if he has some unpublished changes it could cause unnecessary conflicts.

I will leave merging up to him in case there are something he wants to do differently.

FredyC:
I guess I will do merging then :)

elektronik2k5:
Thanks guys! :)

> Should I do a few more changes (is separate PRs)? Here's a list:
> 
> * Upgrade all deps
> * Add spell checker
> * Run tests against built package (not for perf)

@mweststrate what about this list?

mweststrate:
@elektronik2k5 

1. go ahead!
2. can you clarify? 
2. yes that would be awesome, not sure how to deal with obfuscated / not-in-prod errors. Maybe those tests need to be in an if block?

mweststrate:
Unrelated, I was thinking making `observableRequiresReaction` the default, how'd you feel about that?

elektronik2k5:
> 2. can you clarify?

Yes, here's a story: as a renown quality and grammar Nazi :smirk:, I review all of my client's code and catch a lot of typos. However, I once had to work with a team member who was particularly bad at spelling and got sick of pointing out obvious typos. So I decided to add a spell checker as a git hook and it worked well :). However, I also discovered that the codebase still has loads of typos which I missed.

The point is: a computer can spell check much better than any person. I've been using [cspell](https://github.com/streetsidesoftware/cspell) to catch spelling errors in codebases for a few years now.
It works exactly like any other linter for files staged for a commit: fail if non English words are found. Then you can fix the spelling, ignore it once or whitelist the offending word.
cspell also has a VSCode extension which picks up the configuration from a project and integrates with the "problems" panel, just like eslint, typescript and others.

> 3. yes that would be awesome, not sure how to deal with obfuscated / not-in-prod errors. Maybe those tests need to be in an if block?

Perhaps, I'll look into it.

elektronik2k5:
> Unrelated, I was thinking making `observableRequiresReaction` the default, how'd you feel about that?

Sorry, I have no experience with any of the newer configuration options. Before MST I always worked in strict mode (is that `enforceActions: 'always'` nowadays?), but ever since switching to MST I haven't worked with bare mobx. :(

It is probably better to consult @Bnaya, @benjamingr and others.

FredyC:
I think spellchecking in open source lib is a bit stretch. It could annoy contributors as any linter does and the benefit is very small imo. Who cares if there is a typo in a comment? Most people will understand it anyway.

elektronik2k5:
Upgrade to all deps PR is ready: #2344

PR #2348, opened 2020-05-05
Fix #2326  pass correct array to callbacks

### LINKED ISSUE #2326: Mobx Array when using push
I get this error when i use array.push

mobx.module.js:3138 Uncaught Error: [mobx] Modification exception: the internal structure of an observable array was changed.
at ObservableArrayAdministration.updateArrayLength (mobx.module.js:3138)
at ObservableArrayAdministration.spliceWithArray (mobx.module.js:3176)
at Proxy.push (mobx.module.js:3302)
at AppState.js:127

The Source code is

```js
@observable gData = [{
    "title": "1",
    "key": "1",
    "children": [
      {
        "title": "2",
        "key": "2",
        "children": [
          {
            "title": "3",
            "key": "3"
          },
          {
            "title": "34",
            "key": "33"
          },
          {
            "title": "32",
            "key": "32"
          }
        ]
      },
      {
        "title": "21",
        "key": "21"
      }
    ]
  },

  {
    "title": "11",
    "key": "11"
  }];

onDrop = info => {
    console.log(info);
    const dropKey = info.node.props.eventKey; 
    const dragKey = info.dragNode.props.eventKey; 
    const dropPos = info.node.props.pos.split('-');
    const dropPosition = info.dropPosition - Number(dropPos[dropPos.length - 1]);

    const loop = (data, key, callback) => {
      data.forEach((item, index, arr) => {    // I think the cause of the problem***************
        if (item.key === key) {
          return callback(item, index, arr);
        }
        if (item.children) {
          return loop(item.children, key, callback);
        }
      });
    };
    
    const data = [...this.gData];
    // Find dragObject
    let dragObj;
	
    loop(data, dragKey, (item, index, arr) => {
      arr.splice(index, 1); 
      dragObj = item;
    });

    if (!info.dropToGap) {
      // Drop on the content
      loop(data, dropKey, item => {
        item.children = item.children || [];
        // where to insert     
        item.children.push(dragObj); // The Error location***************
      });
    } else if (
      (info.node.props.children || []).length > 0 && // Has children
      info.node.props.expanded && // Is expanded
      dropPosition === 1 // On the bottom gap
    ) {
      loop(data, dropKey, item => {
        item.children = item.children || [];
        // where to insert 
        item.children.unshift(dragObj);
      });
    } else {
      let ar;
      let i;
      loop(data, dropKey, (item, index, arr) => {
        ar = arr;
        i = index;
      });
      if (dropPosition === -1) {
        ar.splice(i, 0, dragObj);
      } else {
        ar.splice(i + 1, 0, dragObj);
      }
    }

    this.gData = data;
  };
```

I think the reason is ' data.forEach '. The item in the loop is not a proxy object ,but 'item.children' is a proxy object
So I replaced ‘data.forEach’ with the following function ,and no errors have occurred since

```
const myforEach = (arr,fn) =>{
      let i
      for(i=0; i<arr.length; i++){
        fn(arr[i],i,arr); 
      }  
    }
```

**Is it a problem with foreach or am I using it wrong ?**

the version i use is

```
"mobx": "^5.15.4",
"mobx-react": "^6.1.8",
```

urugator:
additionally:
Possibly fixed `find`, `findIndex`, `includes` not being reactive in V5
Added support for `flatMap`/`flat` in V5

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2348"></a>

### DIFF
--- a/src/v4/types/observablearray.ts
+++ b/src/v4/types/observablearray.ts
@@ -642,8 +642,6 @@ addHiddenProp(ObservableArray.prototype, toStringTagSymbol(), "Array")
 
 /**
  * Make sure callbacks recieve correct array arg #2326
- * Impl assumes that callback's last 3 arguments are:
- * element, index, array
  */
 ;[
     "every",
@@ -653,8 +651,6 @@ addHiddenProp(ObservableArray.prototype, toStringTagSymbol(), "Array")
     //"flatMap", // not supported
     "forEach",
     "map",
-    "reduce",
-    "reduceRight",
     "some"
 ].forEach(funcName => {
     const baseFunc = Array.prototype[funcName]
@@ -665,13 +661,20 @@ addHiddenProp(ObservableArray.prototype, toStringTagSymbol(), "Array")
     addHiddenProp(ObservableArray.prototype, funcName, function(callback, thisArg) {
         const adm = this.$mobx
         adm.atom.reportObserved()
-        return adm.values[funcName]((...args) => {
-            // replace last arg (the array) with observable array
-            args[args.length - 1] = this
-            // replace last but two arg (the element) with dehanced value
-            args[args.length - 3] = adm.dehanceValue(args[args.length - 3])
-            return callback.apply(thisArg, args)
+        return adm.values[funcName]((element, index) => {
+            element = adm.dehanceValue(element)
+            return callback.call(thisArg, element, index, this)
         }, thisArg)
+    })
+})
+;["reduce", "reduceRight"].forEach(funcName => {
+    addHiddenProp(ObservableArray.prototype, funcName, function(callback, initialValue) {
+        const adm = this.$mobx
+        adm.atom.reportObserved()
+        return adm.values[funcName]((accumulator, currentValue, index) => {
+            currentValue = adm.dehanceValue(currentValue)
+            return callback(accumulator, currentValue, index, this)
+        }, initialValue)
     })
 })
 
--- a/src/v4/types/observablearray.ts
+++ b/src/v4/types/observablearray.ts
@@ -629,21 +629,7 @@ addHiddenProp(ObservableArray.prototype, toStringTagSymbol(), "Array")
 /**
  * Wrap function from prototype
  */
-;[
-    //"every",
-    //"filter",
-    //"forEach",
-    "indexOf",
-    "join",
-    "lastIndexOf",
-    //"map",
-    //"reduce",
-    //"reduceRight",
-    "slice",
-    //"some",
-    "toString",
-    "toLocaleString"
-].forEach(funcName => {
+;["indexOf", "join", "lastIndexOf", "slice", "toString", "toLocaleString"].forEach(funcName => {
     const baseFunc = Array.prototype[funcName]
     invariant(
         typeof baseFunc === "function",
--- a/src/v4/types/observablearray.ts
+++ b/src/v4/types/observablearray.ts
@@ -663,7 +663,6 @@ addHiddenProp(ObservableArray.prototype, toStringTagSymbol(), "Array")
         `Base function not defined on Array prototype: '${funcName}'`
     )
     addHiddenProp(ObservableArray.prototype, funcName, function(callback, thisArg) {
-        //return baseFunc.apply(this.peek(), arguments)
         const adm = this.$mobx
         adm.atom.reportObserved()
         return adm.values[funcName]((...args) => {
--- a/src/v4/types/observablearray.ts
+++ b/src/v4/types/observablearray.ts
@@ -545,9 +545,7 @@ export class ObservableArray<T> extends StubArray {
                 return impl.dehanceValue(impl.values[index])
             }
             console.warn(
-                `[mobx.array] Attempt to read an array index (${index}) that is out of bounds (${
-                    impl.values.length
-                }). Please check length first. Out of bound indices will not be tracked by MobX`
+                `[mobx.array] Attempt to read an array index (${index}) that is out of bounds (${impl.values.length}). Please check length first. Out of bound indices will not be tracked by MobX`
             )
         }
         return undefined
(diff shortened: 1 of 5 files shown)


src/v5/types/observablearray.ts
```diff
@@ -474,27 +472,48 @@ const arrayExtensions = {
...
         const adm: ObservableArrayAdministration = this[$mobx]
         adm.atom.reportObserved()
-        const res = adm.dehanceValues(adm.values)
-        return res[funcName].apply(res, arguments)
+        return adm.values[funcName]((...args) => {
+            // replace last arg (the array) with observable array
+            args[args.length - 1] = this
+            // replace last but two arg (the element) with dehanced value
+            args[args.length - 3] = adm.dehanceValue(args[args.length - 3])
+            return callback.apply(thisArg, args)
```
melnikov-s, line 515: very subtle issue here that for `reduce` and `reduceRight` the context will always be the `initialValue` instead of `null`/`global` .

src/v5/types/observablearray.ts
```diff
@@ -474,27 +472,48 @@ const arrayExtensions = {
      * Without this, everything works as well, but this works
      * faster as everything works on unproxied values
      */
+;["concat", "indexOf", "join", "lastIndexOf", "slice", "toString", "toLocaleString"].forEach(
```
melnikov-s, line 475: missing `flat`, `includes` and  `copyWithin` (?) which will force it to go through the slow proxy path. Also prototype check

urugator: Good points
`flat` - I will add this one to V5 in a similar manner to `flatMap` (with prototype check) 
`includes` - probably missing for the same reason as `find/findIndex` were missing - method list copy pasted from V4 (I think it would be nice to provide own impl in V4, but I don't want to introduce new API atm...same for `flatMap` or whatever is missing in V4...)
`copyWithin` - that would require some additional work I suppose, probably out of scope of this issue

FredyC:
@urugator Are you able to continue working on this?

urugator:
It should be done, it's just waiting for approval :)

FredyC:
Ah right, I've seen unresolved review comments and though it's not ready yet. Let's roll with it then. It will be slightly annoying to merge with the V6 branch. Perhaps you might want to do that so Michel doesn't need to deal with it?

urugator:
I may try to port it (and perhaps improve a bit) manually later (when v6 impl stabilizes) - currently I am not sure how to approach some things regarding feature detection.

FredyC:
Well, the problem at hand is that Michel often merges master into V6 branch so he will have to deal with conflicts. I am not sure what do you mean by feature detection.

urugator:
It's not clear which array methods we should "polyfill" when the native impl isn't available.
In V4, which targets ES5, we implement some methods that are part of ES>5 for convinience.
On the other hand, in this PR I added support for `flatMap`, `flat` for V5, but only if native impl is also available.
And finally there are method(s) (eg `copyWithin`) that are not supported at all and will not work as expected - ideally we should throw in such case (probably the proxy trap should default to error rather than delegating to target - unless we are sure that the target will work as intended - this way we can also be sure we didn't forget to implement something).

Also I have to take a look at how to implement these target dependent features, so they're stripped from sources during compilation if not needed.

I am also not yet familiar with how V6 handles difference between V4/V5 array impl (V4 array is object with numerical props)

FredyC:
Hm, isn't it enough to just check if `Array.prototype.flatMap` (_and others_) is available? I mean it's not our job to make sure those exist. People should polyfill that on their own (eg. with polyfill.io) and then we can inject our logic if it's discovered. It's, of course, tricky race condition depending on what happens first. I am not sure how to handle that properly.

I don't think you will be able to "strip from sources" because there will be only one bundle for all environments so any checking will have to be done in runtime.

mweststrate:
For context, I consider ES5 support to mean: It *can* work on ES 5, given
that the necessary polyfills are provided, rather than: we'll make sure
everything is polyfilled. So I think it is fine if the implementation
assumes / asserts those methods exist, and take it from there.

urugator:
What if some implementation is a complete replacement and doesn't require the native impl (or polyfilled one) to be present - such implementation can also be stripped from bundle when native isn't available, because we can assume that user won't be using it anyway.
However in order to do that, the user would have to explicitely name all the methods that he is going to use - something like `enableArrayMethod1(); enableArrayMethod2(); // etc`, is that correct?

> I don't think you will be able to "strip from sources"

I was refering to that "`immer` technique" that should reduce bundle size, mentioned in v6 proposal issue.

urugator:
Bundling aside ...  should the implementation of any method throw if it's not available on `Array.prototype` or only if the `Array.prototyp.method` is actually required for the impl?

mweststrate:
I don't think we will be using the immer technique, I tried that, but the gains were too insignificant in MobX.  

I think throwing on invocation is the best, assuming we are wrapping the function for tracking purposes.

FredyC:
@urugator What methods specifically are required to be available? Besides those nifty little additions like `flatMap` or `includes` which are not necessary to be used the support is pretty good [across the spectrum](https://kangax.github.io/compat-table/es5/#test-Array_methods). I wouldn't stress much about it and adding invariants for that.

> However in order to do that, the user would have to explicitely name all the methods that he is going to use - something like `enableArrayMethod1(); enableArrayMethod2(); // etc`, is that correct?

That doesn't sound like a good DX to me just to save a few bytes of code.

src/v4/types/observablearray.ts
```diff
@@ -631,28 +629,52 @@ addHiddenProp(ObservableArray.prototype, toStringTagSymbol(), "Array")
...
  * Wrap function from prototype
  */
+;["indexOf", "join", "lastIndexOf", "slice", "toString", "toLocaleString"].forEach(funcName => {
+    const baseFunc = Array.prototype[funcName]
+    invariant(
+        typeof baseFunc === "function",
+        `Base function not defined on Array prototype: '${funcName}'`
+    )
+    addHiddenProp(ObservableArray.prototype, funcName, function() {
+        return baseFunc.apply(this.peek(), arguments)
```
mweststrate, line 639: I'd moved the invariant here indeed, so that people that never call the method aren't killed by it

PR #2359, opened 2020-05-25
Enhancement to observable documentation

jgarboza813:
I have been learning MobX and using it for projects at work for about a year now. I struggled with this section of the documentation when I first came across it.  Specifically, I did not understand which aspect of spreadsheet cells using `observable` was like. My enhancement adds some context to the metaphor so that it is clear *how* `observable` is like a spreadsheet cell.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2359"></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,8 +122,8 @@ class Todo {
 }
 ```
 
-Using `observable` is like turning a property of an object into a spreadsheet cell.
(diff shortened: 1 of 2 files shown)


FredyC:
Thank you, I like it :)

PR #2382, opened 2020-06-23
[WIP] Mobx6 docs

faassen:
I'm going through the documentation to modify it with MobX 6 changes in mind:

* Philosophy: one way to do things. Have one through-line and have alternatives (if needed) described elsewhere.

* Use makeObservable instead of decorators. Decorator support is assumed to be in a package 'mobx-decorators'.

* Assume actions are enforced by default in the docs.

* Since MobX6 brings back non-Proxy support, I also can clean up discussions of MobX 5 vs 4, assuming the old documentation will become less relevant. (of course it may need to still be available somewhere)

There are a whole bunch of questions that arose when I was editing the docs. I've added them in notes-docs.md, and I'd love to see some answers that help guide me forward. Writing documentation can also help clarify APIs, so hopefully my efforts can contribute to that as well.

To see a live version of the docs, go here:

https://deploy-preview-2382--mobx-docs.netlify.app/

(or click on the 'details' link in netlify/mobx-js/deploy-preview in the built status at the bottom)
 
<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2382"></a>

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2382"></a>

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2382"></a>

PR #2388, opened 2020-06-30
Run prettier

FredyC:
@mweststrate Merge it when you want. I noticed in #2382 that @faassen has a bunch of styling changes. It's probably better to do it separately as that PR might take a while.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2388"></a>

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,7 +4,7 @@
     🚨🚨🚨 IMPORTANT NOTICE 🚨🚨🚨
 
     Next major of MobX V6 is under development. Try to direct all changes toward `mobx6` branch instead of `master`.
-    
+
     🚨🚨🚨 IMPORTANT NOTICE 🚨🚨🚨
 
     👋 Are you making a change to documentation only? Delete the rest of the template and go ahead.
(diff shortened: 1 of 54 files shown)


mweststrate:
Could you drop de changes to files in `docs/`? otherwise the merge conflicts will be horrible (git isn't smart enough to follow along even if that branch is prettified, if there are other changes as well). 

Thinking of it, I think we could just work on the docs on the same branch v6, that is probably more straightforward @faassen? And makes also these kind of changes easier.

FredyC:
I don't think it's a problem, these are just whitespace and @faassen already have these changes in that branch. But yea, working on docs directly in this branch would be probably better long term.

FredyC:
Ah and another PR https://github.com/mobxjs/mobx/pull/2389 that has some Prettier styling. I wonder @mweststrate, do you have Prettier installed in your IDE? It seems like it's not formatting it for you. Although it should be done in commit phase unless you are skipping it deliberately.

mweststrate:
@FredyC yeah that is correct, I had disabled it for a while because it messed up snapshots tests every time. But in the end it turned out that was caused by vscode's trim trailing whitespace setting, so it should be good from know on again.

PR #2389, opened 2020-06-30
Add spy typings

phiresky:
This PR adds typings to the spy() function. The changes are 99% compile-time only, except that I added two keys (especially observableKind: "box"|"array"|"map"|...) to the spy events to make it possible to fully discriminate between them at runtime.

# Code change checklist

-   [x] Added/updated unit tests - not needed, everything is compile time 
-   [x] Updated changelog

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2389"></a>


mweststrate:
Please open PR's against the mobx6 branch which is currently the default
branch

mweststrate:
Apologies, misread, it is. Thanks!

mweststrate:
Merged! Thanks for the effort :)

mweststrate:
Reverted the merge for several reasons:

1. tests weren't updated
2. types for map and set were missing

Feel free to reopen a fresh PR with the above changes.

phiresky:
> not all tests were updated

Could you elaborate @mweststrate ? Test behaviour didn't change, since everything I added was typed as any before (so completely unverified), and it is now verified by the typescript compiler. The snapshots did change somewhat but I updated those in 18264bd. 

> 2. The types for map and set are there:

I'm not sure what you mean here - for map and set i use / adjust the existing IMapDidChange etc types.

```ts
type PureSpyEvent = 
// ....
    | (Omit<IMapDidChange<unknown, unknown>, "name"> & { name: string; key: unknown })
    | (Omit<ISetDidChange<unknown>, "name"> & { name: string })
```

and if they weren't the rest of the code wouldn't pass the typescript compiler :)


> 3. docs for spy weren't updated

I can do that

mweststrate:
1. Tests are failing https://app.circleci.com/pipelines/github/mobxjs/mobx/709/workflows/6bc7c1cf-9464-4801-bd7b-6717e2f32f87/jobs/3881/steps. I tried merging it into mobx6 as I assumed it was some transient issue and assumed they were updated given the last commit, but a significant amount failed on the mobx6 branch as well so I reverted. Looking at the job these seem legit failures, so I recommend to use `yarn jest --watch` locally until everything passes. Test behavior did change, since you introduced the `observableKind` field, so this isn't purely a type change (which is fine, but does affect tests). For some tests this could be updated automatically thanks to snapshots, but for others you will need to update tests by hand.
2. sorry, missed overlooked those since the other ones have explicit type definitions. Looks good
3. 👍

PR #2398, opened 2020-07-14
Strongly type spy and unify observe/spy interfaces

phiresky:
I've updated the spy docs, and fixed the tests.

I've also changed the semantics from the previous PR, such that the obserableKind property is only emitted for spy() events not for observe() or intercept() ones. I'd have preferred to have it in observe/intercept too, but I think there's some inconsistency(?) in that events such as IArrayWillChange are reported to observe() but not to spy, so I didn't feel comfortable adding or changing those too much.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2398"></a>


mweststrate:
Feel free to unify them, I think it removes the cognitive burden /
confusion in the api by having those subtle differences between spy and
observe. If you are interested I think it'd be fine to even literally pass
the same object to the callbacks in both cases. Makes it cheaper as well :)

phiresky:
Ok. I've unified the APIs of spy() to be the same as observe(). spyReportStart() currently still has to duplicate the object to add the spyReportStart: true property. Didn't change that since it might be confusing to have that property in observe() events.

PR #2401, opened 2020-07-16
Update README.md

tee:
Update README to reflect correct dollar amount for silver sponsors

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2401"></a>


tee:
This is a mear assumption. It seemed strange to me that silver was at 100pm and bronze was at 500pm. Hopefully not too far off base here.

mweststrate:
The 100$ / 500$ is correct. But the 100$ is monthly, while the 500$ is in total; either one time donations or past but ended sponsorships.

PR #2405, opened 2020-07-22
Skip redundant comparison

MoonBall:
### Code change checklist

-   [x] Updated changelog

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2405"></a>

### DIFF
--- a/src/core/derivation.ts
+++ b/src/core/derivation.ts
@@ -223,12 +223,12 @@ function bindDependencies(derivation: IDerivation) {
             dep.diffValue_ = 1
             if (i0 !== i) observing[i0] = dep
             i0++
-        }
 
-        // Upcast is 'safe' here, because if dep is IObservable, `dependenciesState` will be undefined,
-        // not hitting the condition
-        if (((dep as any) as IDerivation).dependenciesState_ > lowestNewObservingDerivationState) {
-            lowestNewObservingDerivationState = ((dep as any) as IDerivation).dependenciesState_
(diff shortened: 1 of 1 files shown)


FredyC:
@MoonBall Is there some actual issue for this PR or just your assumption about performance? You haven't added any tests or explained anything.

Also note, that you are contributing to the upcoming V6 which you are probably not using yet.

MoonBall:
There is no actual issue. It's only about performance. 

Should I make the pr for master branch?

FredyC:
@mweststrate Probably better for you to decide this.

mweststrate:
@MoonBall, no this branch is fine. I guess the correctness of the change cannot be proven beyond that all tests still pass :) But it is years since this code is written, so a little rationale why the truth branch is always taken, to be able to think about the correctness without fully diving what is happening here again😅. Especially since the branch isn't taken always, I want to make sure that I understand that making the update always, doesn't slow anything down anywhere else:

Screen Shot 2020-07-23 at 20 24 42

MoonBall:
@mweststrate Hi, how do you get the execution count, the 3908x and 14x in your image.

mweststrate:
@MoonBall  yeah, run `yarn test:coverage` and than open `coverage/lcov-report/index.html`

MoonBall:
@mweststrate Thanks. I find that the amount of duplicated observing is only 4 times when I run `yarn test`.
So I close the PR.

PR #2410, opened 2020-07-26
cache getKeys() of ObservableObjectAdministration

MoonBall:
### Code change checklist

-   [ ] Updated changelog

I made the PR to cache getKeys() of ObservableObjectAdministration. 
When I run `yarn test`, the program hit the cache 91 times.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2410"></a>

### DIFF
--- a/src/types/observableobject.ts
+++ b/src/types/observableobject.ts
@@ -80,6 +80,8 @@ export class ObservableObjectAdministration
     interceptors_
     proxy_: any
     private pendingKeys_: undefined | Map<PropertyKey, ObservableValue<boolean>>
+    private keysValue_: PropertyKey[] = []
+    private isStaledKeysValue_ = true
 
     constructor(
         public target_: any,
@@ -220,7 +222,7 @@ export class ObservableObjectAdministration
             const oldValue = oldObservable && oldObservable.get()
             oldObservable && oldObservable.set(undefined)
             // notify key and keyset listeners
-            this.keysAtom_.reportChanged()
+            this.reportKeysChanged()
             this.values_.delete(key)
             if (this.pendingKeys_) {
                 const entry = this.pendingKeys_.get(key)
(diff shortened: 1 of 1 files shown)


src/types/observableobject.ts
```diff
@@ -280,15 +282,24 @@ export class ObservableObjectAdministration
...
         this.keysAtom_.reportObserved()
+        if (!this.isStaledKeysValue_) {
+            return this.keysValue_
+        }
         // return Reflect.ownKeys(this.values) as any
         const res: PropertyKey[] = []
         for (const [key, value] of this.values_) if (value instanceof ObservableValue) res.push(key)
-        return res
+        this.isStaledKeysValue_ = false
+        return (this.keysValue_ = res)
```
mweststrate, line 297: I am curious behind the rationale of this change. Did it solve a specific problem for you?

A risk of this change is that the keys are now shared amongst all readers, so if someone would `keys().sort().forEach()` for example, that would sort the keys for anybody requesting the keys. That could be protected against by using `Object.freeze(this.keysValue_)`.

MoonBall: > I am curious behind the rationale of this change. Did it solve a specific problem for you?
I read the code, and think that it can be more efficient by using cache rather than retrieving `this.values_` every time.

> A risk of this change is that the keys are now shared amongst all readers, so if someone would keys().sort().forEach() for example, that would sort the keys for anybody requesting the keys. That could be protected against by using Object.freeze(this.keysValue_)
I prefer to use `slice()` to solve this issue.

MoonBall: Yes, you're right. The `.slice()` is just a little bit faster by seeing [the bench](https://jsben.ch/ZDbO6).
I added `if(__DEV__) Object.freeze(this.keysValue_)` for avoiding consumer's modification.

PR #2411, opened 2020-07-26
docs: modify little nits

MoonBall:
### Code change checklist

-   [ ] Updated changelog

1. Add a dot at the end of sentence.
2. Use more accurate nouns.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2411"></a>

### DIFF
--- a/docs/best/react.md
+++ b/docs/best/react.md
@@ -194,7 +194,7 @@ message.likes.push("Jennifer")
 ```
 
 This will react with the above sample data, array indexers count as property access. But **only** if the provided `index < length`.
-MobX will not track not-yet-existing indices or object properties (except when using maps).
+MobX will not track not-yet-existing indices.
 So always guard your array index based access with a `.length` check.
 
 #### Correct: access array functions in tracked function
@@ -217,7 +217,7 @@ autorun(() => {
 message.likes[2] = "Jennifer"
 ```
 
-This will react as expected. All array index assignments are detected, but only if `index <= length`.
+This will react as expected. All array index assignments are detected, but only if `index < length`.
 
 #### Incorrect: "use" an observable but without accessing any of its properties
 
--- a/docs/refguide/computed-decorator.md
+++ b/docs/refguide/computed-decorator.md
@@ -119,7 +119,7 @@ configure({
 })
 ```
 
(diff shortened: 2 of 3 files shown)


docs/best/react.md
```diff
@@ -194,7 +194,7 @@ message.likes.push("Jennifer")
 ```
 
 This will react with the above sample data, array indexers count as property access. But **only** if the provided `index < length`.
-MobX will not track not-yet-existing indices or object properties (except when using maps).
```
MoonBall, line 197: The below code will print twice. So I think the not-yet-existing object properties can be tracked.

```
import { observable, autorun } from "../mobx"

const message = observable({
    likes: {},
}) as any

autorun(() => {
    console.log(message.likes[0])
})
message.likes[0] = "Jennifer"
```

docs/best/react.md
```diff
@@ -217,7 +217,7 @@ autorun(() => {
 message.likes[2] = "Jennifer"
 ```
 
-This will react as expected. All array index assignments are detected, but only if `index <= length`.
+This will react as expected. All array index assignments are detected, but only if `index < length`.
```
MoonBall, line 220: consult: 
https://github.com/mobxjs/mobx/blob/master/src/v5/types/observablearray.ts#L464

docs/best/react.md
mweststrate: yeah, this is old and only applies to non proxy environments, so the change is correct.

docs/best/react.md
mweststrate: This was correct, see https://github.com/mobxjs/mobx/blob/e9986766844c1da5cb337b8e88934ac0d4b81aa9/src/v5/types/observablearray.ts#L458, an assignment like `ar[ar.length] = x` will be interpreted as a `push`

MoonBall:
> I just noticed this pr is against master. Since the docs are being completely refactored now in the v6 branch, would you mind checking if the changes are still applicable there?

I checked document of v6. There is no case in v6.
@mweststrate Do you think the fix for v5 is necessary? If it's not necessary, I will close the PR.

mweststrate:
I think it is fine for now since v6 is underway anyway and can be closed

Issue #2386, opened 2020-06-28
Typescript 3.9 breaks getters in extendObservable


JabX:
Typescript 3.9 introduced a sneaky breaking change regarding the way getter are emitted by setting `enumerable: false` instead of `enumerable: true` (associated PR [here](https://github.com/microsoft/TypeScript/pull/32264)).

This causes `extendObservable` to ignore properties defined as getters in its second parameter, like that :

```ts
const obs = extendObservable({a: 2, b: 3}, { get c() { return a + b; });
obs.c // undefined (!!)
```

I'm a bit surprised that no one has yet reported this, it literally broke my entire app once I upgraded. 

I'm not sure how this can be mitigated : how can `extendObservable` enumerate the properties to add if they aren't actually enumerable?

FredyC:
Interesting observation. It doesn't seem we can really do much here. I see some complaint in the linked PR so hopefully, that will give it a second thought. I guess you have to stick to the previous TS till then.

mweststrate:
Non enumerable props can be iterated through Object.getOwnPropertyNames w/o
problem. I think in MobX 6 this would actually work already. MobX 4/5 might
need a patch if this is default behaviour in TS

JabX:
This is actually the only behavior that TS supports now, there is no plan to add a flag or something to control it. I ended up reverting to 3.8 for the time being.

They keep mentioning classes though in their justification (and in the existing complaints), but here I'm not using a class at all, just a plain JS object. Is the behavior really supposed to be the same here? 

I forgot to mention, but I am using MobX 4 (latest), so I don't know if it also happens in 5.

If we were to switch to using `Object.getOwnPropertyNames`, isn't it going to break something else? Like weird inheritance-based patterns or something?

mweststrate:
I think there would be little risk in updating that, as the second argument of extendObservable is supposed to be a plain object anyway.

mweststrate:
Anybody interested in making a PR for this one on v5/4? I think the unit tests from 6 could be reused directly: https://github.com/mobxjs/mobx/blob/7c5ed6ce793b262a7aa8ed5ee6243e2bdc70bc73/test/v5/base/make-observable.ts#L338-L357

mweststrate:
Fixed and released as (4/5).14.5

PR #2412, opened 2020-07-27
extendObservable can be used late and support non-enumerable getters.

MoonBall:
### Code change checklist

-   [ ] Updated changelog

Fix https://github.com/mobxjs/mobx/issues/2386 on v5.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2412"></a>

### DIFF
--- a/src/v4/api/extendobservable.ts
+++ b/src/v4/api/extendobservable.ts
@@ -64,7 +64,7 @@ export function extendObservable<A extends Object, B extends Object>(
     asObservableObject(target, options.name, defaultDecorator.enhancer) // make sure object is observable, even without initial props
     startBatch()
     try {
-        for (let key in properties) {
+        for (const key of Object.getOwnPropertyNames(properties)) {
             const descriptor = Object.getOwnPropertyDescriptor(properties, key)!
             if (process.env.NODE_ENV !== "production") {
                 if (isComputed(descriptor.value))
--- a/src/v4/api/extendobservable.ts
+++ b/src/v4/api/extendobservable.ts
@@ -64,7 +64,9 @@ export function extendObservable<A extends Object, B extends Object>(
     asObservableObject(target, options.name, defaultDecorator.enhancer) // make sure object is observable, even without initial props
     startBatch()
     try {
-        for (const key of Object.getOwnPropertyNames(properties)) {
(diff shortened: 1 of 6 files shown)


mweststrate:
Looking great! As far as I'm concerned `enableNonEnumerablePropertiesInExtendObservable` could be just the default and I don't think it has to be an option. Afaik all normal use cases of extendObservable consist of the second argument to be a locally constructed object literal (and not a class instance or something), so I expect this change to be safe.

MoonBall:
> Looking great! As far as I'm concerned `enableNonEnumerablePropertiesInExtendObservable` could be just the default and I don't think it has to be an option. Afaik all normal use cases of extendObservable consist of the second argument to be a locally constructed object literal (and not a class instance or something), so I expect this change to be safe.

okey, I remove the `enableNonEnumerablePropertiesInExtendObservable` state. PTAL.

mweststrate:
Awesome, thanks for doing this! Would you mind making the same change in the v4 folder? (from 6 onward that wouldn't be needed anymore luckily)

src/v4/api/extendobservable.ts
```diff
@@ -64,7 +64,7 @@ export function extendObservable<A extends Object, B extends Object>(
     asObservableObject(target, options.name, defaultDecorator.enhancer) // make sure object is observable, even without initial props
     startBatch()
     try {
-        for (let key in properties) {
```
MoonBall, line 67: `for ... in ...` contains enumerable properties that the `properties` inherits from its prototype chain. 
But `Object.getOwnPropertyNames` only contains enumerable properties of `properties` itself. 
@mweststrate I'm not sure the difference matters.

mweststrate: @MoonBall good one, no for extendObservable that should be match. However, If I remember correctly the downlevel compilation of `for of` loops is problematic with MobX 4, which might run in environments without iterators. Would you mind changing it to a `.forEach` or class `for(;;)` loop?

MoonBall: I fix `for of` by using `for(;;)`.

MoonBall:
>  If you want to contribute to MobX more in the future, let me know and I send a maintainers invite. 

Thanks, I'm interested in doing it.

Issue #2413, opened 2020-07-27
Mobile Responsiveness issue


Udit-singh:
# Mobx Footer Logo Alignment Fix.

**Description:** - Above the Docs section, the logo should be aligned with other section of the footer

**Screenshot:**

![Screenshot (610)](https://user-images.githubusercontent.com/55452299/88551347-63425f80-d040-11ea-9634-84f793563384.png)

**Environment:**

* OS: Windows
* Browser: Chrome
* View mode - iPhone 6/7/8  IOS
* Version: Version 81.0.4044.113 (Official Build) (64-bit)

mweststrate:
Feel free to take a shot at it (against the mobx6 branch)

PR #2420, opened 2020-08-05
fix: log doesn't align with other section of the footer(#2413)

MoonBall:
### Code change checklist

-   [ ] Updated changelog

fix https://github.com/mobxjs/mobx/issues/2413.
The following image shows the result after the fix.
result (1)

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2420"></a>


FredyC:
Thanks!

PR #2421, opened 2020-08-05
fix: BLM covers the search input in mobile

MoonBall:
### Code change checklist

-   [ ] Updated changelog

The first image below show the case that BLM covers the input search.
The second image below show the result of the fix.
before fix (1)
after fix (1)

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2421"></a>


FredyC:
Thanks!

Issue #2423, opened 2020-08-06
Breaking change in 5.15.5?


austinbutler:
**Intended outcome:**

Upgrade from 5.15.4 to 5.15.5, app still runs.

**Actual outcome:**

App runs with 5.15.4, but crashes on startup with 5.15.5. Seems many things that were previously working are now `undefined`.

Maybe side effect of https://github.com/mobxjs/mobx/issues/2386?

**How to reproduce the issue:**

If necessary I can try to put this together later. But given the extent of the breakage I assume it's not just me at the moment.

**Versions**

```
"mobx": "5.15.5",
"mobx-react-lite": "^2.0.7",
"mobx-state-tree": "^3.17.2",
"typescript": "^3.9.7"
```

MoonBall:
@austinbutler Please give a example code. So I can try to find out the cause.

MoonBall:
The https://github.com/mobxjs/mobx/issues/2386 has two modifications about `extendObservable`.
1. doesn't include properties of object's prototype.
2. include non-enumerable properties of object.

mweststrate:
@austinbutler, please do provide a reproduction. Without this issue is inactionable.

SDSLeon:
mobs-state-tree also affected.

`
export const GroupSetting = types.model('GroupSetting', {
  id: types.string,
  descript: types.string,
  name: types.enumeration([...SettingNamesArray]),
  value: types.string,
});

settigns: types.array(GroupSetting);

get serverTimezone() {
      const setting = self.settigns.find(setting => setting.name === 'TimeZone');

      console.log(setting.value);
    },
`

returns wrapped finded value

so I need to do setting.value.id to actually get id

austinbutler:
Sorry about that, seems this issue it wasn't as broadly applicable as I'd assumed (which is good!).

I've put together a quick approximation of how it happens in my app. If you toggle between `5.15.4` and `5.15.5` in `package.json` you can see on .4 it loads fine, on .5 it crashes.

https://codesandbox.io/s/empty-dawn-76v4f

mweststrate:
@urugator would you mind taking a look at this one? From the console logs in https://codesandbox.io/s/hidden-snowflake-2clmz?file=/src/Models.ts, at first sight it looks like the dehancer is not applied with for the `.map` of observable array, (it seems to be applied to the filter), so I guess the result from `filter` is not dehanced in #2348?

mweststrate:
[...]

PR #2426, opened 2020-08-07
fix array methods not dehancing

urugator:
Fix for: https://github.com/mobxjs/mobx/issues/2423

~~Missing test for `flat/flatMap`, not sure how they're supposed to behave.~~
EDIT: Dehancer is responsible for nested arrays, see the test how it's handled.

Once it's merged I will take a look at V6

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2426"></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # UNPUBLISHED (keep this here and add unpublished changes)
 
+-   Fixed [2423](https://github.com/mobxjs/mobx/issues/2423), array methods not dehancing by [@urugator](https://github.com/urugator)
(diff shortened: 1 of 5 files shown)


airhorns:
Would be great to get this in and potentially worth yanking the release with this bug in it! 💖  This is low enough level functionality that I think a lot of MST users are going to hit this issue when upgrading, could save all of 'em a bit of time tracking this down.

urugator:
The thing is Michael is on holiday and I don't want to mess anything, but I will ask on gitter whether someone can do a release or something...

PR #2436, opened 2020-08-25
Fixed Issue with Silver Sponsor Amount

chase-manning:
Just a small change to Readme :)

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2436"></a>


FredyC:
Notice that small `+`. It's on purpose like this because there is no fixed amount for sponsors. So it's correct like this.

PR #2439, opened 2020-08-26
FIxed typings (function properties instead of methods in decorators)

### LINKED ISSUE #2438: Fix typings, function properties should not be typed as methods
Hello! In typescript there is distinction between methods (which depends on `this`) and just function properties, which is can be used unbound. Currently typescript doesn't lint it (https://github.com/microsoft/TypeScript/issues/7968), but there is eslint rule, which is really helpful (https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md), especially in react codebases. Currently `@computed.struct` is defined as method and not a property, so eslint shout at me for it :) I fix it via custom typings, but I think it should be corrected at sources.

**Intended outcome:**
```ts
export interface IComputed {
    //...
    struct: (target: Object, key: string | symbol, baseDescriptor?: PropertyDescriptor) => void;
}
```

**Actual outcome:**
```ts
export interface IComputed {
    //...
    struct(target: Object, key: string | symbol, baseDescriptor?: PropertyDescriptor): void;
}
```

**Versions**
Mobx 5 latest

Amareis:
Closes #2438

### Code change checklist

-   [ ] Updated changelog

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2439"></a>

### DIFF
--- a/src/v4/api/action.ts
+++ b/src/v4/api/action.ts
@@ -6,7 +6,7 @@ import {
     executeAction,
     fail,
     invariant,
-    namedActionDecorator,
+    namedActionDecorator
 } from "../internal"
 
 export interface IActionFactory {
--- a/src/v4/api/action.ts
+++ b/src/v4/api/action.ts
@@ -6,7 +6,7 @@ import {
     executeAction,
     fail,
     invariant,
-    namedActionDecorator
+    namedActionDecorator,
 } from "../internal"
 
 export interface IActionFactory {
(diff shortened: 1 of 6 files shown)


mweststrate (changes requested):
Thanks for making this PR! However, please open a separate PR per issue. I'm fine merging the TS arrow fn changes right away, but for the computed.shallow see the issue. In any case PRs ideally address only a single issue :) to simplify issue tracking / reporting / reverting

Amareis:
I think .prettierrc should be updated on master :D

mweststrate (approved):

mweststrate:
Merging, thanks again!

mweststrate:
Released

Issue #2448, opened 2020-09-11
isPlainObject() to support objects from another window


ConneXNL:
So I've been running into an issue where I have multiple browser windows, one window holding the state, and the other windows rendering various UI referencing the main window's mobx state.

However, I am running into problems whenever I try to set values with Mobx/MST as the implementation of isPlainObject() in Mobx&MST uses the local window's Object prototype to check if it is an object.

However, when you have multiple windows the Object.prototype of window A !== Object.prototype of window B, thus the below check fails. My setup is really unique but I think apps using Election-like technology might also suffer from this.

Here are the implementation of isPlainObject MobX and MST (found in the utils.ts):
https://github.com/mobxjs/mobx/blob/6ec6499fb8b55a17fe65f42b14d1188fd7fa1ba1/src/utils/utils.ts#L82-L86

https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/utils.ts#L123-L127

So far I have been patching MobX and Mobx-State-Tree to use https://github.com/jonschlinkert/is-plain-object but I figured it might make more sense to see if you are open to change the isPlainObject implementation.

Would you be willing to change isPlainObject() implementation?

FredyC:
Might be worth investigation for the upcoming V6, but I don't we will be backporting since nobody else is interested.

Would you be willing to work on PR?

ConneXNL:
Added a pull request that fixes the issue and doesn't require an external library or much changes.

hartum:
> So I've been running into an issue where I have multiple browser windows, one window holding the state, and the other windows rendering various UI referencing the main window's mobx state.
> 
> However, I am running into problems whenever I try to set values with Mobx/MST as the implementation of isPlainObject() in Mobx&MST uses the local window's Object prototype to check if it is an object.
> 
> However, when you have multiple windows the Object.prototype of window A !== Object.prototype of window B, thus the below check fails. My setup is really unique but I think apps using Election-like technology might also suffer from this.
> 
> Here are the implementation of isPlainObject MobX and MST (found in the utils.ts):
> 
> https://github.com/mobxjs/mobx/blob/6ec6499fb8b55a17fe65f42b14d1188fd7fa1ba1/src/utils/utils.ts#L82-L86
> 
> https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/utils.ts#L123-L127
> 
[...]

PR #2461, opened 2020-09-24
Added support for plain objects from another window

ConneXNL:
Added support for plain objects from another window.

This should fix https://github.com/mobxjs/mobx/issues/2448

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2461"></a>


mweststrate:
Thanks!

PR #2462, opened 2020-09-25
Docs grammar corrections IV

zangornjak:
Evening!

I've:
- added the missing titles
- synced titles with the 1st paragraphs and sidebar_label's
- reworked `README.md`'s Further resources section
- changed "How to read ..." to "About ...", and `README.md`'s sidebar_label "Introduction" to "About MobX", avoiding having Introduction written twice
- updated `docs/refguide/computed.md` (now `docs/computeds.md`)
- updated `docs/refguide/computed-with-args.md` (now `docs/computeds-with-args.md`)
- updated `docs/refguide/autorun.md` (now `docs/reactions.md`)
- updated `docs/refguide/action.md` (now `docs/actions.md`)
- added a video about re-creating MobX with the advanced icon next to it, there's one more ~14min long video that does the same, but I can't find it anymore, it's brilliant though
- changed to a flat directory structure and added links to their new locations in the legacy documents
- updated all references to their new locations
- fixed all dead references or full url / .html / no ext ending ones
- marked all direct mentions of advanced sections with the marker
- surrounded all advanced markings with {} as it looks the cleanest all-around
- etc.

**Q:** ~~Can I also rename files to their new sidebar_label names (e.g. debugging-mobx -> analyzing-reactivity) and mirror folders to left-side headings on the docs website (intro, core, react, tips-and-tricks, configuration)? I'll also fix all references in the docs & correct the `sidebars.json`. If there's anything else I'll find out (probably api link above).~~

**Q:** `observable-sets.md`, `observable-objects.md` and `stateless-HMR.md` are never referenced to in the docs. Should I merge _set_ and _object_ with their counterparts in `api.md` and remove them? Same with `observable-modifiers.md` that's only referenced from `observable-objects.md` that you've pointed out.

**Q:** I just noticed the change count and realized I should've moved the original files to their new locations first and comitted, before creating their replacements. I stupidly did it in a single commit and they were re-created instead of moved. Do we need to fix this? I haven't reversed things in a PR before, but I guess if I follow [these steps](https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-a-pull-request/51400593#51400593) it should work.

_This time around I have local docs running and am observing the changes through them, so some new things are standing out. Previously I've been looking at them on GitHub, and an occasional PR deploy preview._

_All in all, minus the changes resulting from above Q's, only the `api.md` hasn't been updated yet._ :tada: 

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2462"></a>


website/sidebars.json
```diff
@@ -9,25 +9,25 @@
...
     ],
     "Tips & Tricks": [
+      "best/store",
       "best/what-does-mobx-react-to",
       "best/debugging-mobx",
-      "best/store",
-      "react/react-performance",
-      "refguide/mobx-utils",
       "refguide/computed-with-args",
+      "refguide/mobx-utils",
```
zangornjak, line 24: I've moved it to after Computed with args, since there are 3 other ways to do things before resorting to mobx-utils, as per computed-withargs.

website/sidebars.json
```diff
@@ -9,25 +9,25 @@
...
+      "refguide/computed",
+      "refguide/autorun"
     ],
     "MobX and React": [
-      "react/react-integration"
+      "react/react-integration",
+      "react/react-performance"
     ],
     "Tips & Tricks": [
+      "best/store",
```
zangornjak, line 20: I've moved it to the 1st place so there's a clear distinction between basic and advanced docs in the heading.

website/sidebars.json
```diff
@@ -9,25 +9,25 @@
     "MobX core": [
       "refguide/observable",
       "refguide/action",
-      "refguide/autorun",
-      "refguide/computed"
+      "refguide/computed",
```
mweststrate, line 12: Note that this will require a small forwarding note in the explanation of `computed`, as it refers to `autorun` / things being observed.

website/sidebars.json
```diff
@@ -9,25 +9,25 @@
...
+      "refguide/autorun"
     ],
     "MobX and React": [
-      "react/react-integration"
+      "react/react-integration",
+      "react/react-performance"
```
mweststrate, line 17: Makes sense to move performance here. But I think it should be marked advanced as well. It kinda is.

mweststrate:
Concerning file names, I tried to keep the amount of renames to a minimum. Although it creates some inconcistencies, they are often linked to in old issues, so that is why I kept them (otherwise the whole dir structure could actually be removed as well, as it doesn't make much sense anymore)

zangornjak:
I've figured as much and asked before doing it. Ok, crossing that one off. :smile:

website/sidebars.json
zangornjak: Understood & marked it as advanced in the menu & 1st heading. While we're at it, menus use the marking without [], whereas all Tip: and headings use [] around it. Do you want all of them to be a certain way or like this?

mweststrate:
No strong opinion on it. Whatever looks best 😊

website/sidebars.json
zangornjak: I've added it just before the example, pointing at autorun and the Reactions section. Does it resemble something you've had in mind, or was it a miss?

zangornjak:
One more question regarding the file names. Say we moved them to a new hierarchy, but kept the old files with links to their new locations? Then once you see that traffic to the old ones dies down, remove them completely as everybody is happily using the new ones?

No pressure, I'm only thinking out loud.

mweststrate:
That would be perfect 😊.

zangornjak:
Ok. I've removed that comment because I later thought it was kind of dumb. :laughing:

- the inherent extra dir mess, especially if any overlap with the new ones

What kind of directory structure did you have in mind in this case? The intro, core, react, tips-and-tricks, configuration I proposed above, or better? Note that intro & react would overlap in this case. Maybe introduction & with-react instead.

zangornjak:
I've also just tried creating symbolic links from the new locations to the old and the docs seem to render as expected. The only thing I notice is that the left-side menu doesn't show for documents that are not defined within it, so those "redirects" (or ln -s) end up not displaying it, leaving you essentially "stuck" on the landing page without a menu. A click anywhere else fixes it, but still.

So another potential solution could be a bash script that creates / removes these links on command?

mweststrate:
I'd make the dir structure just flat :) It are not so many files, and it is way easier when writing as you only need to know the page id, and not where it lives relative to the current doc. I don't know if the symbolic links work on github hosting as well, so generating a small markdown file would work as well if it doesn't

zangornjak:
Good points and I'm not sure regarding GitHub hosting. I'll move them all to the README.md's level and put notices to where they've moved in the original locations.

zangornjak:
Before I update all the references, let's agree on the new names. I went by the sidebar_label's.

```
action.md -> actions.md
api.md
autorun.md -> reactions.md
computed.md -> computeds.md
computed-with-args.md -> computeds-with-args.md
concepts.md -> the-gist-of-mobx.md
configure.md -> configuring.md
debugging-mobx.md -> analyzing-reactivity.md
decorators.md
extending.md -> custom-observables.md
how-to-read.md -> about-this-documentation.md
installation.md
migrate-to-6.md -> migrating-to-6.md
mobx-utils.md
modifiers.md -> observable-modifiers.md
object-api.md -> collection-utilities.md
object.md
observable.md -> observable-state.md
observe.md -> observe-and-intercept.md
on-become-observed.md -> lazy-observables.md
react-integration.md
react-performance.md -> react-optimizations.md
set.md
stateless-HMR.md
store.md -> defining-data-stores.md
what-does-mobx-react-to.md -> understanding-reactivity.md
```

mweststrate:
Sounds good to me. I think set, lazy and modifiers are the ones not linked
anymore? They can be dropped, the 1st and 3rd are discussed in observable,
the second in custom- observable from top of my head

zangornjak:
Ok, moving forward with those names. I've also updated the comment with a table instead for easier viewing.

`refguide/set`, `refguide/object` and `best/stateless-HMR` are never referenced to in the docs. Lazy and modifiers are under the Tips & Tricks heading (Lazy observables / Custom observables).

zangornjak:
I won't double post needlessly, but in case you're replying from the email a lot, I'm making many edits to correct any f-ups in the original comments.

mweststrate:
Ha sorry yeah mobile day today so my comments are random. Will check again
tomorrow to look what is actually there 😉

zangornjak:
The next version of Docusaurus has dark mode, which would be awesome to have. It's still in alpha, but 64 versions in and appears stable from the "quick glances" I've taken at it. It doesn't appear too much of a hassle to [upgrade to it](https://v2.docusaurus.io/docs/migrating-from-v1-to-v2) since you don't use versioning or translations.

Want me to try locally and see if everything's stable? And if it is pushing the changes so all of us dark mode people can rest our eyes. :laughing:

mweststrate:
@zangornjak I'd hold of with the docusaurus update, mostly because we have quite some customizations, which might be hard to port over. Like the collapsable / linkable summary sections, etc. 

Yeah with an org invite it should be much easier to branch etc w/o needing to sync your fork

mweststrate:
Ok, with the file moves + linking the diff becomes impossible to review. With a quick look around it looks fine, so I'll gonna merge it blindly.😳🙈

mweststrate:
> Q: observable-sets.md...

I think all info in there is already covered, so they can be deleted. 

> Q: I just noticed the change count ...

Oh yes good point, that would have fixed it. Merged to quickly here... It's okay.

PR #2470, opened 2020-10-01
Fix docs link in v6 CHANGELOG

leader22:
Links were pointed to 404. 👀 

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2470"></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,12 @@
 
 ## New features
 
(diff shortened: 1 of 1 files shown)


zangornjak:
Yikes! Thank you for catching that. Blame me as I've pushed for docs name updates, but then didn't think of checking out this file as well.

PR #2472, opened 2020-10-01
Add shell and change directory path of mobx-undecorate

janpieterz:
No tests or any other changes affected ( I think).
Cannot verify this works ok cross-platform though.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2472"></a>

### DIFF
--- a/packages/mobx-undecorate/cli.js
+++ b/packages/mobx-undecorate/cli.js
@@ -3,7 +3,7 @@ const { spawn } = require("child_process")
 
 // this is pretty lame, probably better make a .cmd and .sh file...
 spawn(
-    "node_modules/.bin/jscodeshift",
+    "node_modules\\.bin\\jscodeshift",
     [
         "--extensions=js,jsx,ts,tsx",
         ...process.argv.filter(arg => arg.startsWith("--")),
@@ -13,6 +13,7 @@ spawn(
     ],
     {
         cwd: __dirname,
-        stdio: "inherit"
+        stdio: "inherit",
+        shell: true
     }
 )
--- a/packages/mobx-undecorate/cli.js
+++ b/packages/mobx-undecorate/cli.js
@@ -4,7 +4,7 @@ const path = require("path")
 
 // this is pretty lame, probably better make a .cmd and .sh file...
 spawn(
-    `node_modules${path.sep}.bin${path.sep}jscodeshift`,
+    path.join("node_modules", ".bin", "jscodeshift"),
     [
         "--extensions=js,jsx,ts,tsx",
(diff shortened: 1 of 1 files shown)


packages/mobx-undecorate/cli.js
```diff
@@ -3,7 +3,7 @@ const { spawn } = require("child_process")
 
 // this is pretty lame, probably better make a .cmd and .sh file...
 spawn(
-    "node_modules/.bin/jscodeshift",
+    "node_modules\\.bin\\jscodeshift",
```
mweststrate, line 6: this will break on any unix system. Could you use path.separator or path.join instead?

packages/mobx-undecorate/cli.js
```diff
@@ -13,6 +13,7 @@ spawn(
     ],
     {
         cwd: __dirname,
-        stdio: "inherit"
+        stdio: "inherit",
+        shell: true
```
mweststrate, line 17: can you elaborate on why `shell` is needed here?

mweststrate (changes requested):

packages/mobx-undecorate/cli.js
janpieterz: @mweststrate I have no clue except that it didn't work without that. Found it on a StackOverflow as a comment somewhere while trying to figure it out. 
https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options

janpieterz:
@mweststrate I'm actually trying to replicate this in a clean repo and can't get it to work even with the above fixes. Will do some more digging 👀

packages/mobx-undecorate/cli.js
janpieterz: Tested this with normal cmd, powershell, git-bash, neither worked without the shell: true set.

packages/mobx-undecorate/cli.js
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env node
 const { spawn } = require("child_process")
+const path = require("path")
 
 // this is pretty lame, probably better make a .cmd and .sh file...
 spawn(
-    "node_modules/.bin/jscodeshift",
+    `node_modules${path.sep}.bin${path.sep}jscodeshift`,
```
vkrol, line 7:
```suggestion
    path.join("node_modules", ".bin", "jscodeshift"),
```

vkrol (changes requested):

vkrol (approved):

mweststrate:
Thanks! I'll test it and release soon

mweststrate:
Published as 1.0.1

Issue #2487, opened 2020-10-06
Update documentation for useAsObservableSource hook


marko-hologram:
Hello,

Sorry in advance if this is something that is already planned or in progress, but reading through the React integration part in the docs (https://mobx.js.org/react-integration.html) I noticed few small "problems" with the part of the documentation that mentions `useAsObservableSource` hook (https://mobx.js.org/react-integration.html#computed-props).

1. Link pointing to `useAsObservableSource` doesn't point to the proper place. It points to this https://github.com/mobxjs/mobx-react#useasobservablesource-hook. This section doesn't exist on `mobx-react` github page, but it does exist on `mobx-react-lite` (https://github.com/mobxjs/mobx-react-lite/#useasobservablesourcetsource-t-t-deprecated)
2. `useAsObservableSource` is deprecated as indicated here https://github.com/mobxjs/mobx-react-lite/tree/master#useasobservablesourcetsource-t-t-deprecated and here https://github.com/mobxjs/mobx-react-lite/blob/master/src/useAsObservableSource.ts#L8

I know that having a warning in development is a good way to get notified about this, but maybe adding some info about deprecation in docs directly would just make it more obvious this isn't really a recommended hook to use anymore.

Also, I wasn't sure if I should open this issue in `mobx-react` repo because the topic is this React hook, but it's also about docs and those are located in this main repo here.

I didn't just want to randomly open a PR because I didn't know if this is something you would want addressed and how should it actually be addressed it if you want it to be addressed.

mweststrate:
Hmm that looks like some changes in the docs were list :/. It should be
using useEffect instead to sync the prop into the observable.

marko-hologram:
> Hmm that looks like some changes in the docs were list :/. It should be using useEffect instead to sync the prop into the observable.

You want me to maybe create a PR to update this? I can probably do it today/tomorrow.

If I understand it correctly, it should just use `offset` inside `useLocalObservable` and update it inside a `useEffect`, is that correct? And remove `useAsObservableSource` entirely from that example.

Also then, the link and any reference to `useAsObservableSource` should probably be removed from that part of docs.

mweststrate:
Yes, all correct. PR would be great!

mweststrate:
Thanks for the PR!

PR #2499, opened 2020-10-08
Docs: update React integration computed props tip

marko-hologram:
Updated computed props part of React integration docs. It should address issue https://github.com/mobxjs/mobx/issues/2487

- Cleaned up the example a bit by updating imports
- Removed mention and usage of deprecated [`useAsObservableSource`](https://github.com/mobxjs/mobx-react-lite/tree/master#useasobservablesourcetsource-t-t-deprecated) hook
- Updated demo timer to make it a working example. Not really a MobX issue, but it's ok to have actual working examples in docs

---

As a sidenote, I noticed that some examples might have unused imports in them or even a slight mistake like the one with the timer example function `increaseTimer` not actually being called so it makes the example not really a working one.

I might try to catch some time to go through more example in the next few days, just to clean them up a bit. I would at least try to remove unused imports in order to make examples as clean as possible and easy to use use by copy-pasting.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2499"></a>

### DIFF
--- a/docs/react-integration.md
+++ b/docs/react-integration.md
@@ -50,6 +50,7 @@ setInterval(() => {
     myTimer.increaseTimer()
 }, 1000)
 ```
+
 **Hint:** you can play with the above example yourself on [CodeSandbox](https://codesandbox.io/s/minimal-observer-p9ti4?file=/src/index.tsx).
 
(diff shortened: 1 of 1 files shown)


mweststrate:
Awesome, thanks for the update!

PR #2502, opened 2020-10-09
docs: add "Advanced feature" tooltip to 🚀

jheeffer:
As discussed in https://github.com/mobxjs/mobx/discussions/2465

Add clarifying tooltip to rockets emojis in documentation through span element with title attribute.

For content pages simple html is added to markdown.
For the sidebar a small javascript function adds the wrapping span element. Docusaurus encoded any html tags in the side bar (`&lt;`,`&gt;`), so it was not possible to just add the span to the markdown `sidebar_label` header.

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2502"></a>

### DIFF
--- a/website/static/js/scripts.js
+++ b/website/static/js/scripts.js
@@ -9,14 +9,19 @@ function openTarget() {
         }, 150)
     }
 }
-function addTooltipToSideBar() {
-    var items = document.getElementsByClassName('navItem');
-    for (var item of items) {
-        item.innerHTML = item.innerHTML.replace('🚀', '<span title="Advanced feature">🚀</span>');
+function addTooltipToRockets() {
+    var classNames = ['navGroups', 'onPageNav', 'post'];
+    var rocketRegex = /🚀/g;
+
+    for (var className of classNames) {
+        var els = document.getElementsByClassName(className);
+        for (var el of els) {
+            el.innerHTML = el.innerHTML.replace(rocketRegex, '<span title="Advanced feature">🚀</span>');
+        }
     }
 }
 window.addEventListener('hashchange', openTarget);
 window.addEventListener('DOMContentLoaded', function() {
     openTarget();
-    addTooltipToSideBar();
+    addTooltipToRockets();
 });
--- a/website/static/js/scripts.js
+++ b/website/static/js/scripts.js
@@ -8,8 +8,15 @@ function openTarget() {
             details.scrollIntoView();
         }, 150)
     }
-  }
-  window.addEventListener('hashchange', openTarget);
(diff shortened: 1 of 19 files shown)


website/static/js/scripts.js
```diff
@@ -8,8 +8,15 @@ function openTarget() {
...
         }, 150)
     }
-  }
-  window.addEventListener('hashchange', openTarget);
-  window.addEventListener('DOMContentLoaded', function() {
+}
+function addTooltipToSideBar() {
+    var items = document.getElementsByClassName('navItem');
+    for (var item of items) {
+        item.innerHTML = item.innerHTML.replace('🚀', '<span title="Advanced feature">🚀</span>');
```
FredyC, line 15: Doesn't this mean when you run it again it will double wrap it? That would be good :)

jheeffer: When would it be run again? It's only run on DOMContentLoaded. Isn't that only once [when the initial DOM is loaded](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event)?

FredyC: Wait, so all those occurrences are not the result of this script? You have manually replaced all these? So what is this script good for then? Won't it double wrap in runtime?

jheeffer: From the PR:
> For content pages simple html is added to mark down.
> For the sidebar a small javascript function adds the wrapping span element. Docusaurus encoded any html tags in the side bar (&lt;,&gt;), so it was not possible to just add the span to the markdown sidebar_label header.

The script is for the sidebar, where manual changes are not possible.
The rest was a "manual" search & replace all (and undo 2 unrelated rocket emoji replaces) indeed.

jheeffer: Would you rather have the script wrap all rocket emoji's, including the page content and right side menus? 
I thought having it explicit in the markdown where possible is preferred. But a case can be made for a consistent wrapping method of course.

FredyC: Ah sorry, somehow I missed the PR description 😅 I am not sure why would fine-grained control be necessary here. Anyone adding something _with rocket_ in the future will have to remember to add that span there. So yea, if it's possible, I would be for runtime modification of everyone only

jheeffer: Alright, updated PR.
Scriptwrapping aaaaall the rockets ;).

FredyC:
Thanks, looks great :)

jheeffer:
New PR to improve this one.

https://github.com/mobxjs/mobx/pull/2503

PR #2503, opened 2020-10-10
Docs: fix broken openTarget and add prev/next button to tooltip

jheeffer:
Fix bug introduced in https://github.com/mobxjs/mobx/pull/2502 and add one overlooked rocket emoji location

- `openTarget` should be executed after `addTooltipToRockets` to not break `openTargets`'s async scroll on refresh. Currently, the `details` element in `openTarget` is replaced by `addTooltipToRocket`'s innerHTML reassignment.
- add bottom prev/next doc page buttons to `addTooltipToRockets`

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2503"></a>

### DIFF
--- a/website/static/js/scripts.js
+++ b/website/static/js/scripts.js
@@ -10,7 +10,7 @@ function openTarget() {
     }
 }
 function addTooltipToRockets() {
-    var classNames = ['navGroups', 'onPageNav', 'post'];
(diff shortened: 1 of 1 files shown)


FredyC:
Thanks for checking it and fixing so quickly.

PR #2514, opened 2020-10-14
Docs: Derivations should be pure

nilshartmann:
I think derivations should be pure? 

<a href="https://gitpod.io/#https://github.com/mobxjs/mobx/pull/2514"></a>


mweststrate:
Nice catch, thanks!

PR #2548, opened 2020-10-21
Fix #2542 makeAutoObservable not respecting deep option

### LINKED ISSUE #2542: The `deep` option does not work as documented.
The `deep` option works only in combination with `defaultDecorator`. In contrast to the documented behavior.

The documentation **"Creating observable state"** states the following about the `deep` option:

> `deep: false` uses `observable.ref` by default, rather than `observable` to create new observable members.

Unfortunately setting the `deep: false` option has no effect in combination with `makeAutoObservable()`.

In the code I have found the following:

https://github.com/mobxjs/mobx/blob/e5d91c4eea7d16f74628d1e00d2937df9c225d97/src/api/makeObservable.ts#L258-L260

If in addition to `deep: false` also `defaultDecorator: observable.ref` is set, **it works.**

Either the documentation should be adapted accordingly, or the behavior is an unwanted bug?

**Versions**

MobX 6

urugator:
Fixes #2542
`options.deep` takes a precedence over `options.defaultDecorator`

### DIFF
--- a/.changeset/pink-apricots-sparkle.md
+++ b/.changeset/pink-apricots-sparkle.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fixed [2542](https://github.com/mobxjs/mobx/issues/2542), makeAutoObservable not respecting deep option [@urugator](https://github.com/urugator)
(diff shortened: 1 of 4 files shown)


FredyC:
I suppose this is safe to merge with added tests and everything passing. I don't see anything bad here.

PR #2555, opened 2020-10-24
Fix <Observer> example in react-integration.md

PaulMorris:
Just a few tiny documentation corrections.

### DIFF
--- a/docs/react-integration.md
+++ b/docs/react-integration.md
@@ -295,7 +295,7 @@ Since `onRender` is part of the rendering cycle of `GridRow`, rather than `TodoV
 Or, we can create an in-line anonymous observer using [`<Observer />`](https://github.com/mobxjs/mobx-react#observer):
 
 ```javascript
-const TodoView = observer(({ todo }: { todo: Todo }) =>
+const TodoView = observer(({ todo }: { todo: Todo }) => {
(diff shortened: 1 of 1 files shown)


docs/react-integration.md
```diff
@@ -295,16 +295,16 @@ Since `onRender` is part of the rendering cycle of `GridRow`, rather than `TodoV
 Or, we can create an in-line anonymous observer using [`<Observer />`](https://github.com/mobxjs/mobx-react#observer):
 
 ```javascript
-const TodoView = observer(({ todo }: { todo: Todo }) =>
+const TodoView = observer(({ todo }: { todo: Todo }) => {
```
FredyC, line 298: Well, the `observer` shouldn't be here at all if understand correctly what this example is trying to demonstrate.

urugator: Imo it tries to show that even though `todo.title` appears in `observer`, it's not actually evaluated during it's execution and therefore either `GridRow` must be observer or if thats not possible, the callback must be wrapped in `<Observed>`.

FredyC: Ah right, I missed the whole point with callback here :) Cool.

FredyC:
Thanks @PaulMorris

PaulMorris:
Sure thing!

PR #2563, opened 2020-10-26
Fix #2558 Undecorate removes non-mobx decorators

### LINKED ISSUE #2558: [mobx-undecorate] Removes not mobx decorators
Code:

```
@withRouter
@observer
export class Component extends React.PureComponent {
```

**Intended outcome:**

```
export const Component = withRouter(observer(class Component extends React.PureComponent {}))
```

or throw an error

**Actual outcome:**

at the moment silently changes the code to this
```
export const Component = observer(class Component extends React.PureComponent {})
```

run `npx mobx-undecorate`

**Versions**

mobx-undecorate: 1.0.2

urugator:
Fix for #2558

### DIFF
--- a/.changeset/many-cars-do.md
+++ b/.changeset/many-cars-do.md
@@ -1,5 +1,0 @@
----
-"mobx": patch
----
-
-Fix [#2558](https://github.com/mobxjs/mobx/issues/2558) Undecorate removes non-mobx decorator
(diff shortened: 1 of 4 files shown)


FredyC:
Same here, everything green, so let's push it forward, although release has to be handled manually (@mweststrate).

PR #2573, opened 2020-10-29
infra: Intro yarn 2

Bnaya:
Use yarn 2


mweststrate:
@Bnaya are there any benefits to using yarn 2? Honestly my experience with it so far hasn't been to good, so I'm curious what problem adopting yarn 2 would solve or benefits it would unlock? yarn is doing a pretty fine job so far imho :)

FredyC:
To me what made sense about Yarn 2 was the local cache of installed modules that are committed to the repo. That ensured stability no matter what happens in the NPM world. I see you haven't done that. If you use nodeLinker, why to even bother? I don't even understand what is the new `publishConfig` useful for, you haven't changed that in the PR.

Bnaya:
Not feeling strongly about it :)
In this PR I've made the minimal changes to use yarn 2 as a drop in replacement for yarn 1

The publish config thing will be useful, for example on monorepo, when you want to to point package.json main field to the `src/index.ts` and on publish to `dist/index.js`

But you are right that we can wait for it until we feel concrete  need.

PR #2575, opened 2020-10-30
add throwIfNoDecoratorFound option

xaviergonz:
While adding compatibility for mobx6 to mobx-keystone I found out a case where I need to call makeObservable(this), but with no idea if `this` actually contains mobx decorators or not (since it is third party code).

That made me write this kind of hacky code:
```
      try {
        ;(mobx as any).makeObservable(instance)
      } catch (err) {
        // TODO: perhaps add an option to makeObservable so it will skip this errors
        if (
          err.message !==
          "[MobX] No annotations were passed to makeObservable, but no decorator members have been found either"
        ) {
          throw err
        }
      }
```

This PR adds an option named "throwIfNoDecoratorFound" that when false will not throw in such case.

### DIFF
--- a/.changeset/short-dodos-act.md
+++ b/.changeset/short-dodos-act.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
+Added throwIfNoDecorator option to makeObservable(this)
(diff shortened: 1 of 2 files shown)


mweststrate:
I prefer hacks to remain on your side, rather than having more one-off
options in MobX. I think if you want you can also use reflection to see if
there are any decorators. And skip the call.

xaviergonz:
Hmm, I think I'll skip doing the reflection part since reflecting in the end like makeObservable does (over the prototype chain) would mean it needs to be done 2x and therefore probably slower than catching the exception itself, plus also it would need access to mobxDecoratorsSymbol, which would probably couple it too much to the internals (not like catching a specific error message doesn't though...)

Ah well, was worth a try :)

FredyC:
So I suppose I am closing this :)

PR #2576, opened 2020-10-31
Fix typo in undecorate.ts

tomduncalf:
I know this is a totally unimportant change but I noticed it while looking at the code :)

### Code change checklist


FredyC:
Thanks 😆

PR #2582, opened 2020-11-01
support running in web-worker

tomenden:
In a web-worker environment, there is no `global` or `window`, and therefore currently mobx explodes when running in a web-worker.
This PR replaces `window` with self, as in a browser environment `self === window`.
Please let me know if something additional is needed.

Thanks!

### DIFF
--- a/.changeset/tame-pens-bathe.md
+++ b/.changeset/tame-pens-bathe.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Support running in a web-worker
--- a/src/utils/global.ts
+++ b/src/utils/global.ts
@@ -1,4 +1,4 @@
 export function getGlobal(): any {
     // @ts-ignore
-    return typeof global !== "undefined" ? global : window
+    return typeof self !== "undefined" ? self : global
 }
--- a/src/utils/global.ts
+++ b/src/utils/global.ts
@@ -1,4 +1,17 @@
-export function getGlobal(): any {
-    // @ts-ignore
-    return typeof self !== "undefined" ? self : global
+declare const window: any
+declare const self: any
+
+const mockGlobal = {}
+
+export function getGlobal() {
+    if (typeof window !== "undefined") {
+        return window
+    }
+    if (typeof global !== "undefined") {
+        return global
+    }
+    if (typeof self !== "undefined") {
+        return self
+    }
+    return mockGlobal
 }


FredyC:
Tests are failing, it's definitely not that easy because `self` doesn't exist in other environments. I believe we had some robust checks in V5, so you can try to dig in history.

tomenden:
Thanks @FredyC !
What tests are failing? For some reason I could not access `CircleCI` to see it (do I need some special permissions?) and locally I checked `npm test` and everything seemed to pass.
Regarding the changeset - is there a guideline on how to generate it? I followed [this](https://github.com/atlassian/changesets/blob/master/docs/adding-a-changeset.md#i-am-in-a-single-package-repository) and I get:
```
error Command "changeset" not found.
```
on `npx changeset` and `yarn changeset`

tomenden:
Regarding circleCI - I did manage to access it now, looking into it

tomenden:
It looks like the gitpod url I was using for some reason forked an old version of `mobx` which is the reason for the missing changeset command and for the failing build. I will update it soon

FredyC:
Well, can you please dig into V5 history either way? I am sure there are some gotchas and `global` is needed elsewhere (possibly NodeJS).

tomenden:
@FredyC - Note that `global` is still being used, the difference is using `self` instead of `window`.

Regarding v5, I looked and it seemed like this is the relevant part of code:
https://github.com/mobxjs/mobx/blob/5.15.7/src/v5/core/globalstate.ts#L150-L159

It is actually the same as my implementation as far as I can tell, since `self === window` in all major browsers ([source](https://developer.mozilla.org/en-US/docs/Web/API/Window/self))
image

The only difference in the code between this and v5 is that there is no default empty object used as `mockGlobal` in case there is no `global` or `self`. I did not see the mock fallback in v6 implementation, so I assumed this is not needed. I can add it if you think it is necessary, though I doubt `mobx` will ever run in an environment in which both `global` and `self` are non-existent (are you aware of such an environment?)

FredyC:
Right, can you please use the same code? I am not sure why it got lost in the transition to V6. I remember several people worked on that piece to support all environments, so let's not elaborate on what is supposed to be working. I am sure there might be browsers that need `window` specifically

tomenden:
@FredyC OK. Changed to the v5 implementation

FredyC:
Thank you!

Issue #2564, opened 2020-10-26
browser breakage in mobx@6 esm: "process is not defined"


chase-moskal:
**Intended outcome:**  
i tried to upgrade from mobx@5 to mobx@6 in my esm app

**Actual outcome:**  
browser throws error in [mobx.esm.js](https://unpkg.com/mobx@6.0.1/dist/mobx.esm.js) upon encountering reference to node-specific `process.env`

**How to reproduce the issue:**  
1. in this [codepen example](https://codepen.io/ChaseMoskal/pen/jOrBXjB?editors=1001), open your browser's js console, run the codepen, and you'll see the error `process is not defined`
2. now switch to mobx@5 by replacing the import map:
    ```js
    {
      "imports": {
        "mobx/": "https://unpkg.com/mobx@5.15.7/",
        "mobx": "https://unpkg.com/mobx@5.15.7/lib/mobx.module.js"
      }
    }
    ````
3. run again and you'll see the problem is gone in v5, and the mobx module is logged to console

**My thoughts:**  
- the problem appears when i move from [@5.15.7 mobx.module.js](https://unpkg.com/mobx@5.15.7/lib/mobx.module.js) to [@6.0.1 mobx.esm.js](https://unpkg.com/mobx@6.0.1/dist/mobx.esm.js)
- i see `process.env` in both of those files.. it looks like v5 does some magic to write a fake `process` object when it's missing globally
- i think it would be preferable if mobx didn't write to the global process object at all. sometimes, other programs might check the existence of that object to detect node environment, so if the solution is like mobx@5 here, mobx@6 could accidentally interfere with another system's environment detection, causing tricky bugs

mweststrate:
If you want to consume MobX directly in the browser, I recommend to use the
.min.js or .umd.js builds instead. I still don't know what the expected
behavior of libraries is here in the current JS eco system, as bundlers use
process.env.NODE_ENV (or some arbiratrily other global) to optimize their
builds to conditionally include / exclude debug information and such.

chase-moskal:
@mweststrate 

okay hold on — i'm certain i accidentally freaked you out in my example with usage of community tooling like `es-module-shims`... it probably looks to you like i'm doing something hacky or experimental and deserve to be left unsupported... however this is a mistake: please, let me reframe the issue, stripping away the community-tooling, to highlight the issue against ratified standards, as i should have from the start

- all modern browsers support es modules
- mobx@6 publishes a module which **fails to load** in web browsers.  
[...]

Issue #2530, opened 2020-10-18
🚧 Moving to monorepo


FredyC:
We want to move together [mobx-react](https://github.com/mobxjs/mobx-react) & [mobx-react-lite](https://github.com/mobxjs/mobx-react-lite) (more later maybe) into this repo to ease on issue duplication and to provide better idea of where to find help.

* [x] Transfer existing issues and disable issue reporting in those repos.
* [ ] Move API documentation from respective repos to mobx.js.org.
* [x] Sort out pending PRs (merge or close).

----
* [x] Restructure this repo to monorepo.
* [x] Modify changesets tool to work in monorepo.
* [x] Figure out how to set up CI for the monorepo to run tests (help wanted).
* [x] Build [ESM dev/prod bundles](https://github.com/mobxjs/mobx/issues/2564#issuecomment-717683710) for the browser consumption
* [x] Move mobx-react codebase to this repo and archive it.
* [x] Move mobx-react-lite codebase to this repo and archive it.

spion:
Does this include closed issues? I was looking for a past discussion in mobx-react-lite and i was not able to read or find it.

FredyC:
Yea. I suppose it will block access to closed issues as well. Not the best, but what can you do :) I've transferred that issues and converted to discussion for you: https://github.com/mobxjs/mobx/discussions/2531

FredyC:
If anyone is feeling on working on the documentation migration, help is certainly wanted. I will be able to get to this in a week or more.

jeremy-coleman:
I would like to see the 3 versions align if possible. It would make keeping compatible versions a non issue.

FredyC:
> I would like to see the 3 versions align if possible. It would make keeping compatible versions a non issue.

Nah, that doesn't make much sense. These packages are not that tightly coupled. The `mobx` is the main one here and when a major version happens, it might affect others. But not the other way around. Changes to binding packages hardly warrant publishing `mobx` even without changes.

Bnaya:
related proposal: move to yarn2
I've been tracking yarn 2 progress, and when using nodeLinker (and not pnp) it's a drop-in replacement for yarn 1,
And actively maintained, and support additional goodies

FredyC:
@Bnaya Can you elaborate on what would be the benefit of that? I tried to toy with Yarn2 shortly after its release and was a bit confused about it.

jeremy-coleman:
[...]

PR #2589, opened 2020-11-03
Transform to monorepo

FredyC:
The hardest part of #2530 seems to be working now. Namely...

 - all `mobx` related stuff moved to `packages/mobx`
 - defined packages in root `package.json`
 - mobx tests are in `packages/mobx/__tests__` to unify some convention
 - ESM bundles without NODE_ENV are available too as per #2564 (should be documented? 🤷)
 - publishing should be done exclusively through _changesets_ now
 - updated Jest to v26
 - coverage is handled by GitHub action instead of slowing down CI

Everything is executed through workspaces now, so there are very little commands in the root package. For example, to run mobx tests, you do `yarn mobx test`. There is no command to run tests from all packages, I wasn't able to figure that out.


mweststrate:
Does changeset still support collecting changes per project and releasing them separately?

Note that for the tests in mobx-react / mobx-react-lite the simplest way to run them against 'current' rather them a published version might be to link them together using https://github.com/mweststrate/relative-deps

FredyC:
> Does changeset still support collecting changes per project and releasing them separately?

Yes, of course. When you run `yarn changeset` after this is merged, it either should detect if you did changes to a single package. If you changed multiple ones, it will ask for which you want to write the changeset.

Thanks for the tip about  https://github.com/mweststrate/relative-deps, will look into that later.

FredyC:
Let's roll with it and deal with issues later :)

PR #2594, opened 2020-11-04
Fix #2579 observable does not ignore class instances

### LINKED ISSUE #2579: mobx-react(-lite) useLocalObservable will wrap a class instance in observable if make(Auto)Observable is not used
### Context
I am migrating my boilerplates 
* from mobx v4/v5 mobx-react v6 
* to mobx v6, mobx-react v7

### Intended outcome

I switched to `mobxReact.useLocalObservable()`.

For some complex components, I have classes for local state. 
So I have adapted the code to have calls like `mobxReact.useLocalObservable(() => new MyState())`

While most of the time, `make(Auto)Observable` will be used, I also had the case where a class "in WIP" was just passing non observable data.

```js
import * as mobx from 'mobx'

export default class AppSettingsVM {
  constructor ({rootStore}) {
    mobx.makeAutoObservable(this) // <- WITHOUT this line, the instance is turned into an observable and the app crashes
    this.settings = rootStore.coreStore.settings
  }
}

```

**Actual outcome:**

From the docs, if useLocalObservable returns a class instance, it should be kept as is, which is not what is happening here and leads to a crash.

Use `useLocalObservable` where it returns a class instance that is not using mobx make(Auto)Observable.

**Versions**
mobx v6, mobx-react v7

urugator:
Fixes #2579


FredyC:
I guess this is fine to merge with tests passing. I don't see anything wrong about it.

PR #2612, opened 2020-11-11
Update README.md

urugator:
Updated links to point to current docs and repos

Some topics in the README  (provider/inject/reactive props/hooks)  are not covered by current docs, only by [https://mobx-react.js.org/](https://mobx-react.js.org/),  so I've left these unchanged.


FredyC:
LGTM, thanks.

PR #2696, opened 2021-01-05
fix: mobx should be one of the dependencies.

iam-frankqiu:
### Code change checklist

Because the `mobx` involved by `mobx-react`.So it's should be one of the dependencies. Not devDependencies.

### DIFF
--- a/packages/mobx-react/package-lock.json
+++ b/packages/mobx-react/package-lock.json
@@ -1,0 +1,18 @@
+{
+    "name": "mobx-react",
+    "version": "7.0.5",
+    "lockfileVersion": 1,
+    "requires": true,
+    "dependencies": {
+        "mobx": {
+            "version": "6.0.4",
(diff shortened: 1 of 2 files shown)


mweststrate:
MobX is a peer dependency, just like React, having it as direct dependency will cause multiple, probably conflicting, copies of MobX to be included in to a project, leading to bugs and increasing bundle size.

iam-frankqiu:
> MobX is a peer dependency, just like React, having it as direct dependency will cause multiple, probably conflicting, copies of MobX to be included in to a project, leading to bugs and increasing bundle size.

But when I only involved `mobx-react` in my repository without `mobx`. It led to a bug.

FredyC:
You should follow peer dependency recommendation ;) I know, it's hard sometimes, because many packages don't maintain it properly and we tend to ignore it altogether. That's a problem of the tool, not of this package.

PR #2717, opened 2021-01-15
Export `CreateObservableOptions` type

ahoisl:
### Code change checklist

-   [ ] **IMO unnecessary** - Added/updated unit tests 
-   [ ] **IMO unnecessary** - Updated `/docs`. For new functionality, at least `API.md` should be updated
-   [X] Verified that there is no significant performance drop (`npm run perf`)

Fixes issue #2716

### DIFF
--- a/.changeset/olive-spies-cry.md
+++ b/.changeset/olive-spies-cry.md
@@ -2,4 +2,4 @@
 "mobx": patch
 ---
 
-`CreateObservableOptions` is now exported
+The TypeScript type `CreateObservableOptions` is now exported.
--- a/.changeset/olive-spies-cry.md
+++ b/.changeset/olive-spies-cry.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+`CreateObservableOptions` is now exported
--- a/packages/mobx/src/mobx.ts
+++ b/packages/mobx/src/mobx.ts
@@ -73,6 +73,7 @@ export {
     transaction,
     observable,
     IObservableFactory,
+    CreateObservableOptions,
     computed,
     isObservable,
     isObservableProp,


FredyC:
Thanks for your contribution. Please add the changeset as per the comment above.

urugator:
`CreateObservableOptions` defines undocumented fields that shouldn't be relied on.

ahoisl:
> 
> 
> `CreateObservableOptions` defines undocumented fields that shouldn't be relied on.

Sorry, I don't really understand. The options are documented here, so what do you mean by "undocumented"?
https://mobx.js.org/observable-state.html#options-

These are the default options used by makeObservable - are we talking about different things?

urugator:
https://github.com/mobxjs/mobx/blob/398078777801f2f54de051ca88072a142039a985/packages/mobx/src/api/observable.ts#L39-L46
`defaultDecorator`, `equals`, `proxy` shouldn't be part of the public API

ahoisl:
@urugator That's a different problem then in my opinion, because these options already are part of the public API. If you use makeObservable, the will already show up as valid options. We probably need a sub-type then? That would be a breaking change to the API, though, to my understanding.

urugator:
@mweststrate what's the intention/reasoning here?
`equals` is only used by `observable.box` ...
`proxy` seems to be used only by (2) tests (to force non-proxy impl)
`defaultDecorator` doesn't seem to be used at all, perhaps backward compatibility?

mweststrate:
@urugator `equals` and `defaultDecorator` are internally be used by `.struct`, `.ref` and `.shallow`? If not I think it has been accidentally left there by a past cleanup. `proxy` should definitinely be documented and is an omission of mine in https://mobx.js.org/observable-state.html#options-; it is pretty useful as it allows you to skip using Proxies if the object shape is stable, which is (probably) faster, compatible with IE and a lot debug friendlier (it is still a shame debuggers don't treat proxies nicer).

mweststrate:
see also: https://mobx.js.org/observable-state.html#avoid-proxies

ahoisl:
@urugator  I don't think I can do much more for the issue with the probably "left-over" props in the options `equals` and `defaultDecorator.` In any way, people could have already used those options and as I understand it, they would have an impact for some cases. So changing or removing the props would be a breaking change.

As these props are already public and exporting the `CreateObservableOptions` does not make a difference in regards to these props already being public/exported, can we close this PR and handle the props in a separate issue (if necessary)?

mweststrate:
I think it is fine to merge this

mweststrate (approved):

.changeset/olive-spies-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"mobx": patch
+---
+
+`CreateObservableOptions` is now exported
```
FredyC, line 5: Just a small nit :)
```suggestion
The type `CreateObservableOptions` is now exported.
```

PR #2727, opened 2021-01-18
Do not create generic arguments if keepDecorators passed

vkrol:
### Code change checklist

If `keepDecorators` flag is passed to `mobx-undecorate` there is no need to create generic arguments for `makeObservable`.

### DIFF
--- a/.changeset/gorgeous-coats-rest.md
+++ b/.changeset/gorgeous-coats-rest.md
@@ -1,6 +1,5 @@
 ---
 "mobx-undecorate": patch
-"mobx": patch
 ---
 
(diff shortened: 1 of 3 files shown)


mweststrate (approved):
Looks good! Thanks for noticing and fixing this!

.changeset/gorgeous-coats-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+"mobx-undecorate": patch
+"mobx": patch
```
FredyC, line 3:
```suggestion
```

Issue #2670, opened 2020-12-07
"Parameter implicitly has an 'any' type." error when passing action-wrapped function with implicit argument types to optional property


seanchas116:
**Intended outcome:**

`action` -wrapped function with implicit argument types can be assigned as an optional property of a object, and the types are correctly inferred.

```ts
import { action } from "mobx";

// Both should be OK

const tes1: {
  a: (x: number) => void;
} = {
  a: action((x) => console.log(x.toFixed()))
};

const test2: {
  a?: (x: number) => void;
} = {
  a: action((x) => console.log(x.toFixed()))
};
```

**Actual outcome:**

Implicit argument types in `action` -wrapped function assigned as an optional property of a object becomes `any` and causes an error in struct environments. 

```ts
import { action } from "mobx";

// OK
const tes1: {
  a: (x: number) => void;
} = {
  a: action((x) => console.log(x.toFixed()))
};

// OK in MobX 5, Error in MobX 6 (x becomes any)
const test2: {
  a?: (x: number) => void;
} = {
  a: action((x) => console.log(x.toFixed()))
  // ↑ "Parameter 'x' implicitly has an 'any' type. ts(7006)" error in this line
};
```

Looks like MobX 5 works fine because `undefined` is added to `T` in [this line](https://github.com/mobxjs/mobx/blob/c194a0adf360cc46cae0dbd806a51887fa6f35b2/src/v5/api/action.ts#L14). Adding `undefined` to `T` in Mobx6 `action.d.ts` fixed the problem in my environment.

**How to reproduce the issue:**

https://codesandbox.io/s/mobx-action-error-gnd8v

**Versions**

Occurs in MobX 6. MobX 5 works fine.

kubk:
It's better to add both `undefined` and `null` to support promises: https://github.com/mobxjs/mobx/pull/2213
Mobx 5 also had unit tests for such cases: https://github.com/mobxjs/mobx/pull/2217
But they were removed in Mobx 6

kubk:
Can make a PR to fix it if anyone interested

mweststrate:
Should be fixed in 6.0.5

PR #2740, opened 2021-01-22
Infer optional / promise action args type correctly

iChenLei:
### Code change checklist

### PR Background
seanchas116 create this issue #2670, seanchas116 found mobx can't infer `action` callback arguments types correctly.
Previous pr #2213 #2217 (by @kubk ) fix it, but mobx6 remove it (I don't know why).
I recover it and add more unit test.

### 🚨 Attention🚨 
```js
// Promises
// TODO: re-enable
// Promise.resolve(1).then(
//     action(arg => {
//         assert<IsExact<typeof arg, number>>(true)
//     })
// )
```
I noticed this comment, so maybe we need check why this unit test be disabled previously.
@FredyC @urugator 

> If this pr isn't useful, just close it 😃

### DIFF
--- a/.changeset/calm-files-nail.md
+++ b/.changeset/calm-files-nail.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Infer optional / promise `action` args type correctly
--- a/packages/mobx/__tests__/v5/base/typescript-tests.ts
+++ b/packages/mobx/__tests__/v5/base/typescript-tests.ts
@@ -2086,19 +2086,40 @@ test("type inference of the action callback", () => {
     )
 
     // Promises
-    // TODO: re-enable
-    // Promise.resolve(1).then(
-    //     action(arg => {
-    //         assert<IsExact<typeof arg, number>>(true)
-    //     })
-    // )
+    Promise.resolve(1).then(
+        action(arg => {
+            assert<IsExact<typeof arg, number>>(true)
+        })
+    )
 
-    // // Promises with names actions
-    // Promise.resolve(1).then(
-    //     action("named action", arg => {
-    //         assert<IsExact<typeof arg, number>>(true)
-    //     })
-    // )
+    // Promises with names actions
+    Promise.resolve(1).then(
+        action("named action", arg => {
+            assert<IsExact<typeof arg, number>>(true)
+        })
+    )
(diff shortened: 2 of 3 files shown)


FredyC (approved):
Thanks for the contribution, adding tests, and doing research. I am not sure either why it was removed in V6, perhaps Michel can provide some insight. Or it could have been an oversight.

kubk (approved):

mweststrate (approved):
I'm highly confused that this works :) I can see how `action<T>( ` without constraint could fix the problem, but I find it very remarkable that the union with `null | undefined` does so as well. But playground results are consistent, so let's go with it! 

I think I accidentally deleted in a clean up, so thanks for bringing it back! An extensive comment over there linking back to this PR, playground would be appreciated!

[Playground](https://www.typescriptlang.org/play?#code/DYUwLgBAHgXBBiBXAdgYzASwPbINwCh8oIBeCACgEpSA+CAbwF9DRIBPOK2iANywwAmBfG1IQAZinTZkXJoQD0CiAKwgAzsgDkkAO5YATgGtOGZOJAGQaEBABGVgIZH1lfEolTMOCLoOOABwAeABUIECgwawF1BC8ZGnJxZDgQylSGRWV9Y3UAGntESEcIYAwwMFAIMCxfQVtyR1RUEACwWOQfSTRvZFjHWN0QYGB092Vu6R8-QNDE5NT0iDD6LN9DFwKAK0R1SBR1RCtfAAtRGDdJ3t9-YLCIqOQYuJ6ZCAAfCBQBEHEzEAEHwgyEQI3mKWWSxW+AgsIgVjAR2QEjw+GY+FQOD21Q0YAATHBVrDHAB+TiwYGIAC2dks1BIdD4ggIjDERIgHnUJywoMBZgsBmqJ0ckGIGFiWHE1TYAVsIJplhhEEccBmAXI5Cg9LomL6WFAADpgFgAOaag01eAYKAAqiUNywjyARMIIAAiAAKjn8VPAlggWigWggGCpATKqHKwFEwv6yK0jmQbCDYBlIAN1XU5AA7AAGHMANkorvCBgMhmDyLAJ3FpX+aOEavIAGZqJzubzgVhis1WpAm0A)

Issue #2721, opened 2021-01-18
displayName not working in React 17 with observer


codeBelt:
**Intended outcome:**

In this old issue (https://github.com/mobxjs/mobx/issues/141) it states the below should work. Also In the [docs](https://mobx.js.org/react-integration.html#displayname) it states **it will be fixed in React 17**. 

```js
export const MyComponent2 = observer(() => "hello2!");
MyComponent2.displayName = "MyComponent2";
```

Is this an issue with `observer` now and/or should the [docs](https://mobx.js.org/react-integration.html#displayname) be updated to say `displayName` does not work with `observer`?

**Actual outcome:**
Viewing "React DevTools" in CodeSandbox notice `_c3` for the MyComponent2 name:

<img width="1197" alt="Screen Shot 2021-01-18 at 8 37 06 AM" src="https://user-images.githubusercontent.com/2893563/104928559-66236a80-5968-11eb-939c-1bb5c2c542d8.png">

**How to reproduce the issue:**
https://codesandbox.io/s/mobx-displayname-reat-17-z61pb?file=/src/MyComponent.jsx

**Versions**

- mobx: 6.0.4
- mobx-react: 7.0.5
- react: 17.0.1
- react-dom: 17.0.1

mweststrate:
Looks like a bug indeed, setting a displayName on a memo should work nowadays: https://codesandbox.io/s/react-17-memo-displayname-forked-do95u?file=/src/index.js

urugator:
Swap these?
https://github.com/mobxjs/mobx/blob/680ad86d76ab9dd9fbaeddcc73409047e08febca/packages/mobx-react-lite/src/observer.ts#L75-L76

I think it's because

iChenLei:
## TL;DR
React v17 support set memo displayName, but only can set display once!

## Code analysis
> packages/react/src/ReactMemo.js

```js
  if (__DEV__) {
    let ownName;
    Object.defineProperty(elementType, 'displayName', {
      enumerable: false,
      configurable: true,
      get: function() {
        return ownName;
      },
      set: function(name) {
        ownName = name;
        if (type.displayName == null) {  👈🏻   // This is why !
          type.displayName = name;
        }
      },
    });
  }
```
You can only set displayName once. 

```diff
      set: function(name) {
-        ownName = name;
-        if (type.displayName == null) {
-          type.displayName = name;
+        if (typeof name === 'string') {
+          ownName = name;
+          type.displayName = name; 
+        } else {
[...]

PR #2741, opened 2021-01-23
replace the deprecated react type definition with recommended type def

iChenLei:
### Code change checklist

### PR Backgroud
When I try to solve this problem #2721 , I noticed that mobx-react(and lite) source code use 
deprecated react type definition, Such as `RefForwardingComponent ` `React.Props`
and `StatelessComponent `, so I created a PR to replace these with official recommended type def.

#### StatelessComponent
```ts
  /**
   * @deprecated as of recent React versions, function components can no
   * longer be considered 'stateless'. Please use `FunctionComponent` instead.
   *
   * @see [React Hooks](https://reactjs.org/docs/hooks-intro.html)
   */
  type StatelessComponent<P = {}> = FunctionComponent<P>;
```
#### RefForwardingComponent
```ts
  /**
   * @deprecated Use ForwardRefRenderFunction. forwardRef doesn't accept a
   *             "real" component.
   */
  interface RefForwardingComponent <T, P = {}> extends ForwardRefRenderFunction<T, P> {}
```
#### React.Props
```ts
  /**
   * @deprecated. This was used to allow clients to pass `ref` and `key`
   * to `createElement`, which is no longer necessary due to intersection
   * types. If you need to declare a props object before passing it to
   * `createElement` or a factory, use `ClassAttributes<T>`:
   *
   * ```ts
   * var b: Button | null;
   * var props: ButtonProps & ClassAttributes<Button> = {
   *     ref: b => button = b, // ok!
   *     label: "I'm a Button"
   * };
   * ```
   */
  interface Props<T> {
      children?: ReactNode;
      key?: Key;
      ref?: LegacyRef<T>;
  }
```

### DIFF
--- a/.changeset/brown-camels-smash.md
+++ b/.changeset/brown-camels-smash.md
@@ -1,0 +1,6 @@
+---
+"mobx-react": patch
+"mobx-react-lite": patch
+---
+
+replace the deprecated react type definition with recommended type definition
--- a/packages/mobx-react-lite/src/observer.ts
+++ b/packages/mobx-react-lite/src/observer.ts
@@ -8,7 +8,7 @@ export interface IObserverOptions {
 }
 
 export function observer<P extends object, TRef = {}>(
-    baseComponent: React.RefForwardingComponent<TRef, P>,
+    baseComponent: React.ForwardRefRenderFunction<TRef, P>,
     options: IObserverOptions & { forwardRef: true }
 ): React.MemoExoticComponent<
     React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<TRef>>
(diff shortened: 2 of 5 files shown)


FredyC:
Thanks, I will keep it open for now. In my opinion, as long as the last package version has those types, even deprecated, it's no harm to use them. The other way around we would be forcing users to type package versions that include new types and that would be a breaking change.

Or did you also discover that old types would be causing some issues? In that case, I would ask you to include test cases for that.

iChenLei:
Yeah, You consider much more code change impact as a mobx maintianer. mobx-react(and lite) are very popular react ecosystem library and have many real user around the world,  we need care any change even it's type defintion. `on hold` is a good choice, I will refocus on this PR when @types/react remove these deprecated type def code or old types would be causing some issues(will add unit test). Anyway , thanks for your advices. 😄

may1393 (approved):

lastthyme (approved):

akphi:
Hi, brought here by https://github.com/mobxjs/mobx/issues/3370. With `react 18` is this the right time to do this upgrade now?

spamforhope:
Yes, react 18 made removed deprecated parts and now mobx-react-lite typings doesn't work at all. 
Please, merge those changes and make new patch release ASAP

kubk:
@spamforhope Consider using [patch-package](https://www.npmjs.com/package/patch-package) as a temporary solution. The adaptation to React 18 is being worked on.

PR #2742, opened 2021-01-23
Docs: remind users to confirm that @babel/preset-env disabled loose t…

franckchen:
I found mobx6 unable to work properly when @babel/preset-env enable loose transformation. so i think its necessary to inform other users.

### Code change checklist

### DIFF
--- a/docs/enabling-decorators.md
+++ b/docs/enabling-decorators.md
@@ -85,6 +85,8 @@ Install support for decorators: `npm i --save-dev @babel/plugin-proposal-class-p
 }
 ```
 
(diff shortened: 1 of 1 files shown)


FredyC:
I am not entirely sure about the composition of the whole `babel-preset-env`, but setting `loose` for the whole preset will change it for all included plugins and that's not what everyone might want. Besides, not everyone might be using that preset, so it's misleading and some users might get the impression they need it here.

For MobX we only care about the `@babel/plugin-proposal-class-properties` which needs to be `false`. If you are saying the `loose` is disabled by default, that's exactly what we want. That would mean you have set it `true` in your setup perhaps?

Also, note this is already covered in the [Migration guide](https://mobx.js.org/migrating-from-4-or-5.html#getting-started).

I am going to close this, I don't think this should be in the docs.

PR #2743, opened 2021-01-23
Remove sideEffects section in mobx-react-lite package.json

vkrol:
`sideEffects: false` is unsafe, because `mobx-react-lite/index.ts` contains a side effect https://github.com/mobxjs/mobx/blob/main/packages/mobx-react-lite/src/index.ts#L9. It was fixed in 2.2.2 https://github.com/mobxjs/mobx-react-lite/pull/314, but was reintroduced in some 3.x version.

### Code change checklist

### DIFF
--- a/.changeset/sweet-shrimps-float.md
+++ b/.changeset/sweet-shrimps-float.md
@@ -1,0 +1,7 @@
+---
+"mobx-react-lite": patch
+"mobx-react": patch
+"mobx": patch
+---
+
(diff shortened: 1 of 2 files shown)


FredyC:
Please include the changeset as per the comment above.

vkrol:
@FredyC thanks, done.

FredyC:
Yea, that's correct because it depends on each other. Upon release, it will also update peerDeps automatically.

Issue #2752, opened 2021-01-28
Mobx 6.1 breaks spying on actions in unit test


heath-freenome:
**Intended outcome:**

In our unit tests we are using the `jest.spyOn()` function to ensure that certain actions are called under certain conditions (with expected values) in our components.

**Actual outcome:**

With the 6.1 release changes all of our spys are failing in the same manner as below
```
setUsernameSpy = jest.spyOn(authStore, 'setUsername');
```
results in:
```
TypeError: Cannot assign to read only property 'setUsername' of object '[object Object]'
```

**How to reproduce the issue:**
I'm pretty sure it's obvious how to reproduce the issue. My question is HOW can I verify that an action on an observable is called with the proper arguments now that 6.1 is available? Is there an option I can set to turn off making actions read-only for tests?

**Versions**
6.0.4 - working
6.1 - not working

urugator:
>  Is there an option I can set to turn off making actions read-only for tests?

Not yet, but we would like to adress this somehow and we are open to suggestions.
We would like to avoid a global config flag as that could cause some actually broken code to pass the tests.

urugator:
We would like to know if it's only about writability of actions or whether we should also expect demand for configurability of actions or other fields (computed/observable). 
So if anyone reading this has an experienced in testing mobx based codebases, we want to hear from you!
We would like to keep things constrained to keep user in the safe zone, but we also don't want to throw obstacles in the way for legitime use cases.

vkrol:
@urugator 
> We would like to know if it's only about writability of actions or whether we should also expect demand for configurability of actions or other fields (computed/observable).

It might be helpful:
> Jest should be able to mock non-writable but configurable properties.
> Currently jest will throw an error when trying to replace non-writable (but configurable) properties.

https://github.com/facebook/jest/issues/8656

heath-freenome:
Are actions still configurable? If so, then maybe before applying the spy I can make it temporarily writable?

urugator:
Depends if it's plain object or not. All annotated fields of non-plain objects (classes) are non-configurable as well.
We will probably end up with `configure({ safeDescriptors: false })` or something like that. Actually already working on it, but still considering alternatives...

heath-freenome:
I'm using classes, so what can I expect?

urugator:
Non-configurable ... the main reason is that we want to prevent following:
```javascript
class Child extends Parent {   
   // This is like calling defineProperty with new descriptor
[...]

PR #2761, opened 2021-01-29
configure({ safeDescriptors: false })

urugator:
Added new configuration option `safeDescriptors: boolean = true` as discussed in #2752
Improved configuration section in docs - should be now clearer what are the defaults.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -72,7 +72,7 @@ Also, occassionally you will have a case where you have to supress the warnings 
 That is fine, there are good exceptions to these recommendations.
 Don't be fundamentalistic about them.
 
-#### `enforceActions`
+#### `enforceActions: "observed" | "always" | "never" = "observed"`
 
 The goal of _enforceActions_ is that you don't forget to wrap event handlers in [`action`](actions.md).
 
@@ -88,7 +88,7 @@ Since state should in principle always be created from some event handlers, and 
 
 In the rare case where you create observables lazily, for example in a computed property, you can wrap the creation ad-hoc in an action using `runInAction`.
 
-#### `computedRequiresReaction: boolean`
+#### `computedRequiresReaction: boolean = false`
 
 Forbids the direct access of any unobserved computed value from outside an action or reaction.
 This guarantees you aren't using computed values in a way where MobX won't cache them. In the following example, MobX won't cache the computed value in the first code block, but will cache the result in the second and third block:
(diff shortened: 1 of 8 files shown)


docs/configuration.md
```diff
@@ -72,7 +72,7 @@ Also, occassionally you will have a case where you have to supress the warnings
 That is fine, there are good exceptions to these recommendations.
 Don't be fundamentalistic about them.
 
-#### `enforceActions`
+#### `enforceActions: "observed" | "always" | "never" = "observed"`
```
mweststrate, line 75: Could you put the possible values in the text, something like:

```
Usage: `enforceActions: "observed" | "always" etc
```

Otherwise any link to this config item will stop working, and it will be hard to maintain links if the possible values change over time

docs/configuration.md
```diff
@@ -177,9 +177,20 @@ afterEach(() => {
 })
 ```
 
+### `safeDescriptors: boolean = true`
+
+MobX makes some fields **non**-`configurable` or **non**-`writable` to prevent you from doing things that are not supported or would most likely break your code. However this can also prevent spying/mocking/stubbing in your tests.
+`configure({ safeDescriptors: false })` disables this safety measure, making everything `configurable` and `writable`.
+Note it doesn't affect existing observables, only the ones created after it's been configured.
+<span style="color:red">Use with caution</span> and only when needed - do not turn this off globally for all tests, otherwise you risk false positives (passing tests with broken code).
```
mweststrate, line 185: I think the last sentence is a little pedantic, and I'd argue it is actually quite safe to turn it off for unit tests, as in all other cases it will fail fast, which should be covered by any superficial form of integration, e2e, manual testing

mweststrate (approved):
Approved, but please update the headers before merging :)

docs/configuration.md
urugator: Hm that didn't occur to me, what about:
`### [enforceActions: "observed" | "always" | "never" = "observed"](#enforceActions)`

docs/configuration.md
urugator: It may fail as slow as anything else (on specific IO), especially in case of non-writable actions on plain (dynamics) objects.
Redefinition in class is not the only case where it can fail, if that's what you're assuming - and you still have to create an instance, which can happen absolutely anytime.

urugator: We can always be less pedantic. For now lets be cautious, so if there is a problem, we can always say: *"We told you so"* :)

PR #2766, opened 2021-01-29
Fix: Inherited annotated fields are not assignable

urugator:
Non-writable actions and flows defined on prototype prevented initializing the prop via assigment inside constructor.
Since this PR, things defined on prototype won't use safe descriptors.
This doesn't mean that we support runtime prototype modifications, but we assume that doing so isn't common enough to justify this safety measure, while preventing other potentially valid usages.

### DIFF
--- a/.changeset/great-pants-do.md
+++ b/.changeset/great-pants-do.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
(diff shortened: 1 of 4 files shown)


mweststrate (approved):
LGTM! Thanks for the followup

PR #2767, opened 2021-01-30
fix: mobx-undecorate don't crash when node's loc is null

### LINKED ISSUE #2747: [mobx-undecorate] Lot's of "Cannot read property 'start' of null" errors
Running the mobx-undecorate utility in a fairly large react + typeascript project created using creat-react-app produces lots of these warnings:

> Users/simonstrandman/work/.../file.ts Transformation error (Cannot read property 'start' of null)
> TypeError: Cannot read property 'start' of null
>     at warn (/Users/simonstrandman/work/.../node_modules/mobx-undecorate/src/undecorate.ts:405:38)
>     at createConstructor (/Users/simonstrandman/work/.../node_modules/mobx-undecorate/src/undecorate.ts:341:17)
>     at NodePath.<anonymous> (/Users/simonstrandman/work/.../node_modules/mobx-undecorate/src/undecorate.ts:195:13)
>     at /Users/simonstrandman/work/.../node_modules/mobx-undecorate/node_modules/jscodeshift/src/Collection.js:75:36
>     at Array.forEach (<anonymous>)
>     at Collection.forEach (/Users/simonstrandman/work/.../node_modules/mobx-undecorate/node_modules/jscodeshift/src/Collection.js:74:18)
>     at transform (/Users/simonstrandman/work/.../node_modules/mobx-undecorate/src/undecorate.ts:168:37)

This happens in `mobx-undecorate/src/undecorate.ts` in the function `warn()`:

```ts
function warn(msg: string, node: Node) {
    if (process.env.NODE_ENV === "test") {
        return
    }
    const line = lines[node.loc!.start.line - 1] // <-- This line is causing the crash
    const shortline = line.replace(/^\s*/, "")
    console.warn(
        `[mobx:undecorate] ${msg} at (${fileInfo.path}:${node.loc!.start.line}:${
            node.loc!.start.column
        }):\n\t${shortline}\n\t${"^".padStart(
            node.loc!.start.column + 1 - line.indexOf(shortline),
            " "
        )}\n`
    )
}
```

Turns out `node.loc` can in fact be null. If I change the above with a null check on `node.loc` before using it, it runs properly and the actual warnings is outputed to the console. In my case the warning was:
> Generated new constructor for class  Foo. But since the class does have a base class, it might be needed to revisit the arguments that are passed to `super()`

iChenLei:
### Code change checklist

### PR Target
fix #2747 

### PR Background
`warn` util function's params `node`'s loc is null in some scenarios.
```ts
const line = lines[node.loc!.start.line - 1]  👈   //  when `loc` is null, codemod script crash
```

### Why loc will be `null`
mobx-undecorate build on top of `jscodeshift` and use `@babel/parser` as ast parser
```js
const babylon = require("@babel/parser")
```
`jscodeshift` based on `recast`, `recast` inner logic will make loc be null in some scenarios.
https://github.com/benjamn/recast/blob/master/lib/util.ts#L180-L183
```js
  } else if (node.declaration && isExportDeclaration(node)) {
    // Nullify .loc information for the child declaration so that we never
    // try to reprint it without also reprinting the export declaration.
    node.declaration.loc = null;
```
VScode debug screenshot 👇

### Solve it
add null check for node.loc
when `loc` is null, just `warn` important info(e.g. warn msg and fileinfo path), like this 👇

~~I remove the `test` mode check, for avoid regressing in the future.~~
```diff
-        if (process.env.NODE_ENV === "test") {
-            return
-        }
```
~~mute `warn` log in unit test. ( won't effect other package unit test~~
~~beforeEach(() => {~~
~~jest.spyOn(console, "warn").mockImplementation(() => {})~~ 
~~})~~
👆 I'm stupid

### DIFF
--- a/packages/mobx-undecorate/src/undecorate.ts
+++ b/packages/mobx-undecorate/src/undecorate.ts
@@ -402,15 +402,19 @@ export default function transform(
         if (process.env.NODE_ENV === "test") {
             return
         }
-        const line = lines[node.loc!.start.line - 1]
-        const shortline = line.replace(/^\s*/, "")
-        console.warn(
-            `[mobx:undecorate] ${msg} at (${fileInfo.path}:${node.loc!.start.line}:${
-                node.loc!.start.column
-            }):\n\t${shortline}\n\t${"^".padStart(
-                node.loc!.start.column + 1 - line.indexOf(shortline),
-                " "
-            )}\n`
-        )
+        if (node.loc) {
+            const line = lines[node.loc.start.line - 1]
+            const shortline = line.replace(/^\s*/, "")
+            console.warn(
+                `[mobx:undecorate] ${msg} at (${fileInfo.path}:${node.loc.start.line}:${
+                    node.loc.start.column
+                }):\n\t${shortline}\n\t${"^".padStart(
+                    node.loc.start.column + 1 - line.indexOf(shortline),
+                    " "
+                )}\n`
+            )
+        } else {
+            console.warn(`[mobx:undecorate] ${msg} at (${fileInfo.path})\n`)
+        }
     }
 }
--- a/.changeset/witty-hotels-ring.md
+++ b/.changeset/witty-hotels-ring.md
@@ -1,0 +1,5 @@
+---
+"mobx-undecorate": patch
+---
+
+mobx-undecorate don't crash when node's loc is null


packages/mobx-undecorate/src/undecorate.ts
```diff
@@ -399,18 +399,19 @@ export default function transform(
     }
 
     function warn(msg: string, node: Node) {
-        if (process.env.NODE_ENV === "test") {
```
FredyC, line 402: Why did you remove this check? It's not really useful to see all warnings when running tests.

packages/mobx-undecorate/src/undecorate.ts
```diff
@@ -399,18 +399,19 @@ export default function transform(
...
-            return
+        if (node.loc) {
+            const line = lines[node.loc!.start.line - 1]
+            const shortline = line.replace(/^\s*/, "")
+            console.warn(
+                `[mobx:undecorate] ${msg} at (${fileInfo.path}:${node.loc!.start.line}:${
+                    node.loc!.start.column
+                }):\n\t${shortline}\n\t${"^".padStart(
+                    node.loc!.start.column + 1 - line.indexOf(shortline),
+                    " "
```
FredyC, line 410: No need to include those when checking it above.
```suggestion
                    node.loc.start.column
                }):\n\t${shortline}\n\t${"^".padStart(
                    node.loc.start.column + 1 - line.indexOf(shortline),
                    " "
```

iChenLei: Sorry, I forgot it. 😃

FredyC:
It's not exactly the best approach to show all warnings in tests. Some of them are triggered on purpose and it makes it harder to navigate results.

iChenLei:
@FredyC Agree with u, thanks for fast code review. 😆  I remove the unit test cause it's not meaningful.
I'm stupid, when I `spyOn` the console's warn, We don't test the real `warn` function actually.

PR #2769, opened 2021-01-31
add warn for extending builtins

### LINKED ISSUE #2765: makeObservable crashing on class derived from ObservableMap for mobx 6.1.x
**Intended outcome:**
Should be able to derive class from `ObservableMap` and run `makeObservable(this)` in its constructor.
This code works in `mobx@6.0.5` but breaks in `mobx@6.1.x`:

```
export class ObservableMapLimitedSize<T> extends ObservableMap<string, T> {
    earliestVisible: number = 0;
    latestVisible: number = 0;

    constructor(private maxLength: number) { 
        super(); 
        makeObservable(this);
    }
    ...
```

**Actual outcome:**
In `mobx@6.1.x` it throws the error `adm.make_ is not a function` at `makeObservable.ts:38`

Run the code above.

**Versions**
Getting the crash in `mobx@6.1.x`. It worked fine in `mobx@6.0.x`.

iChenLei:
### Code change checklist

### PR Background
issue -> fix #2765 
Add friendly warn for extending builtins action.
Current 👇

After 👇

### DIFF
--- a/packages/mobx/__tests__/v5/base/make-observable.ts
+++ b/packages/mobx/__tests__/v5/base/make-observable.ts
@@ -17,7 +17,8 @@ import {
     configure,
     flow,
     override,
-    $mobx
+    ObservableSet,
+    ObservableMap
 } from "../../../src/mobx"
 
 test("makeObservable picks up decorators", () => {
@@ -221,6 +222,35 @@ test("makeObservable supports autoBind", () => {
     expect(isAction(t.bound)).toBe(true)
     expect(t.unbound.call(undefined)).toBe(undefined)
     expect(t.bound.call(undefined)).toBe(t)
+})
+
+test("Extending builtins is not support #2765", () => {
+    class ObservableMapLimitedSize extends ObservableMap {
+        limitSize = 0
+        constructor() {
+            super()
+            makeObservable(this, {
+                limitSize: observable
+            })
+        }
+    }
+
+    class ObservableSetLimitedSize extends ObservableSet {
+        limitSize = 0
+        constructor() {
+            super()
+            makeObservable(this, {
+                limitSize: observable
+            })
+        }
+    }
+
+    expect(() => {
+        new ObservableMapLimitedSize()
+    }).toThrowErrorMatchingInlineSnapshot(`"[MobX] Extending builtins is not support"`)
+    expect(() => {
+        new ObservableSetLimitedSize()
+    }).toThrowErrorMatchingInlineSnapshot(`"[MobX] Extending builtins is not support"`)
 })
 
 test("makeAutoObservable has sane defaults", () => {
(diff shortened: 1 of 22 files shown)


urugator:
Thanks for the PR, but please try to handle it like this:
https://github.com/mobxjs/mobx/blob/ddf99789daebc6a891c930bf77df132d532e090f/packages/mobx/src/types/observableobject.ts#L648
```javascript
if (hasProp(target, $mobx)) {
  if (__DEV__ && !(target[$mobx] instanceof ObservableObjectAdministration)) {
     // Something like
     die(`Cannot convert '${getDebugName(target)}' into observable object: The target is already observable of different type. Extending builtins is not supported.`)
  }
  return target
}
```
That should cover other unsupported cases like `observable/extendObservable(boxedObservable|observableArray|observableMap|observableSet)` etc
And please add tests for these other types.

iChenLei:
@urugator Yes, sir. 😃

iChenLei:
@urugator Hi, sir. Did we need test all types? I found that only `observableMap` and `observableMap` are export from `mobx` npm package.

> packages/mobx/src/mobx.ts

### extendObservable

> https://mobx.js.org/api.html#extendobservable


#### May be it's only use for function/object ? Any suggestion to test it ?

### makeAutoObservable

```ts
// https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/api/makeObservable.ts#L50-L55
if (__DEV__) {
    if (!isPlainObject(target) && !isPlainObject(Object.getPrototypeOf(target)))
        👇  // user can't extending builtin, so don't need check more
        die(`'makeAutoObservable' can only be used for classes that don't have a superclass`) 
    if (isObservableObject(target))
        die(`makeAutoObservable can only be used on objects not already made observable`)
}
```

packages/mobx/__tests__/v5/base/make-observable.ts
```diff
@@ -223,6 +224,39 @@ test("makeObservable supports autoBind", () => {
...
+            makeObservable(this, {
+                limitSize: observable
+            })
+        }
+    }
+
+    expect(() => {
+        new ObservableMapLimitedSize()
+    }).toThrowErrorMatchingInlineSnapshot(
+        `"[MobX] Cannot convert 'ObservableMap@10' into observable object: The target is already observable of different type. Extending builtins is not supported."`
```
urugator, line 251: Please pass stable name: `makeObservable(_,_, { name: "TestObject" })` otherwise the tests are fragile - they break when you move them.

iChenLei: OK, I noticed it actually. Thank you for fast reply. Haha

urugator:
> Did we need test all types?

It's not about preventing `extend Something`.
It's about inability to pass observable of different type than object to `extendObservable`/`makeObservable`/`makeAutoObservable`.
So `makeObservable(observable([]))`, `extendObservable(observable.box())` etc should fail (not neccessarily with the same error, but they must not pass).
It doesn't have to happen inside constructor when extending, thats just a special case.
Is it clear?

mweststrate (approved):
Looking good, thanks for making this change!

PR #2770, opened 2021-01-31
Doc improvements

urugator:
Mostly regarding actions and subclassing

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -208,11 +208,13 @@ class Parent {
 
     // on prototype
     action() {}
+    boundAction() {}
 
     constructor() {
         makeObservable(this, {
             arrowAction: action
             action: action,
+            boundAction: action.bound,
         })
     }
 }
@@ -222,18 +224,21 @@ class Child {
 
     // OK
     action() {}
+    boundAction() {}
 
     constructor() {
         super()
         makeObservable(this, {
             arrowAction: override,
             action: override,
+            boundAction: override,
         })
     }
 }
 ```
 
-For more information see [subclassing](observable-state.md#subclassing)
+To **bind** a single _action_ to `this`, `action.bound` can be used instead of _arrow functions_.<br>
+For more information see [**subclassing**](observable-state.md#subclassing).
 
 ## Asynchronous actions
 
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -224,6 +224,7 @@ class Child {
     action() {}
 
     constructor() {
+        super()
         makeObservable(this, {
             arrowAction: override,
             action: override,
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -197,13 +197,42 @@ Usage:
 Use this utility to create a temporarily action that is immediately invoked. Can be useful in asynchronous processes.
 Check out the [above code block](#examples) for an example.
 
-## Arrow actions and inheritance
+## Actions and inheritance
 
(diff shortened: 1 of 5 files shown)


docs/observable-state.md
```diff
@@ -56,92 +114,87 @@ class Doubler {
...
         makeObservable(this, {
-            // overriden fields
-            action1: override,
-            computed1: override,
-            // new fields
-            observable2: observable,
-            computed2: computed,
-            action2: action,
-            arrowAction2: action
+            // inherited
```
iChenLei, line 180: missing `super()` ?

docs/actions.md
```diff
@@ -198,6 +197,44 @@ Usage:
...
+        super()
+        makeObservable(this, {
+            arrowAction: override,
+            action: override,
+        })
+    }
+}
+```
+
+For more information see [subclassing](observable-state.md#subclassing)
```
mweststrate, line 236: This would be a great section to plug `action.bound` as alternative to arrow funcs?

mweststrate: Or link to what you wrote below

mweststrate: Geez I literally cant read

mweststrate, line 236: I'd make sure that you hint forward to action.bound, before people start to autoBind everywhere: "To bind a single action to `this`, `action.bound` can be used instead of arrow functions, for more information see subclassing[link]"

docs/observable-state.md
```diff
@@ -56,92 +116,88 @@ class Doubler {
...
+    *fetch() {
+        const response = yield fetch("/api/value")
+        this.value = response.json()
+    }
 }
 ```
 
+**All annotated** fields are **non-configurable**.<br>
+**All non-observable** (stateless) fields (`action`, `flow`) are **non-writable**.
+
```
mweststrate, line 129: I think our tabs become a bit overwhelming now, on my browser I cant even see them all. I think it 'd be good to move subclassing section to the end of the page as separate section (below "a note on classes") and have separate details or tabs with subclass examples there.

mweststrate:
I gave the docs a read, generally looks good, but I think we should move the subclassing section to tips & tricks. It is a level of detail that is not applicable to everyone, and people don't need to understand it before they run into some error, in which case they can proactively find the page themselves. Would you mind to move it to separate page and just link it from the classes section (& probably from limitation 4 and 7)

urugator:
Sure. The reason why I made it so prominent is because I expect that this will be one of the first problems that people will encounter with quite common usage (from the issues here I suspect ppl are unfortunatley using inheritance quite reguraly).
It also mentions things (non-configurability/non-writability) which applies even without inheritance.
Personally when I encounter some problem, I don't need to know every detail, but I prefer if I can be like *"I think, I've seen this mentioned somewhere..."*. You know, I don't need to know everything about subclassing, but I want to know that there IS a subclassing or that there is a mention of some TypeError. The biggest problem is if you don't know what to look for...
Also I think that gotchas or hard cases generally should be prominent, because well, they're the complicated parts that will most likely cause troubles.
Ideally I think it's nice to have an overview page that covers almost everything, without going into details, just so you can make a picture of what the lib does and what to look for.

urugator:
Subclassing has own page now + few changes here and there, let me know if it needs some additional adjustments.

mweststrate (approved):
Looking good. Thanks for improving the docs!

PR #2773, opened 2021-02-02
Small decorators optimization

urugator:
Avoids an annotation object copy per decorator application.
Since mixing decorators/annotations is not officialy supported (it may actually work, but there are not tests for it), annotation itself doesn't have to know that it's decorator - ~~once target has `storedAnnotationsSymbol` - meaning at least one decorator was applied - we assume that all annotations are decorators.
If one would try to mix decorators/annotation (unsupported), this change breaks check for non-existing prop (it won't throw).
The decorator detection could be made a bit more realiable by changing `make_` to accept additional `isDecorator` param and calling: `adm.make_(key, annotations![key], !annotations)` from `makeObservable`. 
Not sure if worth it ATM, but at least good to have this noted here.~~
EDIT: Changed to check for specific key, should work equally well in this case.

### DIFF
--- a/packages/mobx/src/types/actionannotation.ts
+++ b/packages/mobx/src/types/actionannotation.ts
@@ -62,7 +62,7 @@ function make_(adm: ObservableObjectAdministration, key: PropertyKey): void {
     }
     if (annotated) {
         recordAnnotationApplied(adm, this, key)
-    } else if (!adm.target_[storedAnnotationsSymbol]) {
+    } else if (!adm.target_[storedAnnotationsSymbol]?.[key]) {
         // Throw on missing key, except for decorators:
         // Decorator annotations are collected from whole prototype chain.
         // When called from super() some props may not exist yet.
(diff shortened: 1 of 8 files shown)


packages/mobx/src/types/actionannotation.ts
```diff
@@ -61,7 +62,7 @@ function make_(adm: ObservableObjectAdministration, key: PropertyKey): void {
     }
     if (annotated) {
         recordAnnotationApplied(adm, this, key)
-    } else if (!this.isDecorator_) {
+    } else if (!adm.target_[storedAnnotationsSymbol]) {
```
mweststrate, line 65: Probably I misunderstand the logic, but should this be `adm.target_[storedAnnotationsSymbol]?.[key]`?

urugator: Actually not a bad idea :) I think I was worried it would return false if `recordAnnotationApplied` is called before (it removes the annotation from `adm.target_[storedAnnotationsSymbol]`), but that can't happen here...

mweststrate (approved):

PR #2775, opened 2021-02-02
detect globalThis

pkit:
`globalThis` is a __standard__ global right now
should use it whenever available

The current code fails in sandbox environments where only standard `globaThis` exists (regression from 5.x which works fine)

### Code change checklist

### DIFF
--- a/.changeset/shy-mangos-shake.md
+++ b/.changeset/shy-mangos-shake.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+use globalThis in global detection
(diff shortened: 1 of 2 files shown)


FredyC:
Never heard of `globalThis`, do you have more information and why it should take precedence over other options?

Also, you need to rebase on the latest code and include changeset set per the comment above.

pkit:
@FredyC https://tc39.es/ecma262/#sec-global-object

FredyC:
Thanks, I see it's supported in major browsers, that's good to know. I guess it's safe to merge.

PR #2748, opened 2021-01-25
PoC Disable debug names in production builds

iChenLei:
### Code change checklist

-   [ ] Added/updated unit tests
-   [ ] Updated `/docs`. For new functionality, at least `API.md` should be updated
-   [ ] Verified that there is no significant performance drop (`npm run perf`)

### PoC for a feature

👉  Disable debug names in production builds

### Current Status
Discussion stage 👉  #2737 , I will start try to implement it when we get a definitive conclusion !

Issue #2737, opened 2021-01-21
Disable debug names in production builds


mweststrate:
See https://github.com/mobxjs/mobx/discussions/2728

iChenLei:
## Simple JS Heap Memory Test
#### wirte simple loop to create many mobx observable object
```js
import { configure, observable } from 'mobx';

configure({
  useProxies: "never"
});

const HugeRootMobx = {};
for (let i  = 0; i < 100000; i++) {  👈🏻   // Yeah, One hundred thousand loop
 HugeRootMobx[i] = observable({ val: i });
}

window.mobxTest = HugeRootMobx;
```
Add compiled js to html and open in chrome.

* Open chrome devtools
* Click Memory Tab
* Click `Take snapshot` button
* Save `HEAP SNAPSHOT`

#### Modify Mobx Source Code
Force `name_`(a.k.a `debug name`) to `void 0` everywhere.

```ts
// `Atom`、 `ComputedValue`、 `ObservableMap`、 `ObservableSet` and `ObservableValue`
name_ = void 0;
```
Run test again and save `HEAP SNAPSHOT`

#### Compare two JS Heap snapshot

![Snapshot](https://user-images.githubusercontent.com/14012511/105702163-54ccf600-5f46-11eb-99c6-9088430410b8.png)

I noticed that disable debugs name in production builds can save memory usage.(~16%)

## Simple test, not precise

If any inaccuracy, please mention it in this discussion line. 😃 

## More Think

It looks like implement this feature is meaningful. Maybe we need support mobx4 mobx5 and mobx6 both. 
But dose it break mobx-devtools user experience ? And any missing hidden problem ？

> Anyway, I think it's very meaningful for big data admin webapp which built on top of mobx. 
It's the same for 3d webapp 、Heavy js games and more.

 Look ashley-taylor's  public github activity (fork github.com/hazelcast/hazelcast),  I know why they need this feature. 😆 
```
ashley-taylor  6d ago

We have a rather large data structure.
Running the chrome heap dump we notice that the mobx debug names represent about 13% of our heap.
Would be nice if you could disable the names.
```

FredyC:
Resolved in #2780

PR #2780, opened 2021-02-04
disable debug names in production builds

iChenLei:
### Code change checklist

> Updated `/docs`. For new functionality, at least `API.md` should be updated
👆  Need add it in official document ?

### PR Background
reopen pr for #2748 

#### Why not just add `__DEV__` for `getNextId()`
```ts
this.keysAtom_ = createAtom(__DEV__ ? `${this.name_}.keys()` : "ObservableMap.keys()")  
 // 👆  e.g. why do this?
```
`getNextId()` is default debugnames gen util, and most time we will create new `Atom`, `ObservableValue` with `name` constructor params, so it also create many `cancatenated string` and `string` in JS Heap.(more details -> #2737 
#### Run simple benchmark
```js
import { observable } from 'mobx';

const HugeRootMobx = {};
for (let i  = 0; i < 100000; i++) {  👈🏻   // Yeah, One hundred thousand loop
 HugeRootMobx[i] = observable({ val: i });
}

window.mobxTest = HugeRootMobx;
```
> Test on microsoft edge 87

#### This is JS Heap compare between different logic process.

If we only deal with `getNextId()`,  not optimized enough .

@mweststrate Need your code review, maybe I missing something. Thank you, sir 😄


packages/mobx/src/core/computedvalue.ts
```diff
@@ -88,7 +88,7 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
     lastAccessedBy_ = 0
     lowestObserverState_ = IDerivationState_.UP_TO_DATE_
     unboundDepsCount_ = 0
-    mapid_ = "#" + getNextId()
+    mapid_ = __DEV__ ? "#" + getNextId() : "#"
```
mweststrate, line 91: Looking at this one, it seems that `mapid_` is used nowhere. Feel free to drop it entirely if that still passes the test

packages/mobx/src/core/reaction.ts
```diff
@@ -57,15 +57,15 @@ export class Reaction implements IDerivation, IReactionPublic {
     diffValue_ = 0
     runId_ = 0
     unboundDepsCount_ = 0
-    mapid_ = "#" + getNextId()
+    mapid_ = __DEV__ ? "#" + getNextId() : "#"
```
mweststrate, line 60: Same (probably the IDerivation interface needs to adapted as well)

packages/mobx/src/types/observablearray.ts
```diff
@@ -129,8 +129,11 @@ export class ObservableArrayAdministration
         public owned_: boolean,
         public legacyMode_: boolean
     ) {
-        this.atom_ = new Atom(name || "ObservableArray@" + getNextId())
-        this.enhancer_ = (newV, oldV) => enhancer(newV, oldV, name + "[..]")
+        this.atom_ = new Atom(
+            __DEV__ ? name || "ObservableArray@" + getNextId() : "ObservableArray"
+        )
+        this.enhancer_ = (newV, oldV) =>
+            enhancer(newV, oldV, __DEV__ ? name + "[..]" : "ObservableArray[...]")
```
mweststrate, line 136: NIT: `[...]` -> `[...]`

mweststrate (changes requested):
Thanks for looking into this. Looks awesome! Small request for changes as I think the `mapid_` can probably be dropped entirely (or it is used somewhere, and in that case I'm not sure the change is safe)

packages/mobx/src/core/computedvalue.ts
iChenLei: Hi, sir. It look likes we can delete it directly?

I search mobx codebase globally and found that nowhere is using this variable ( I am confused 😪

iChenLei: If we delete mapid_ directly, The Unit Tests will be failed.
1. keep it 
2. delete it and `jest --updateSnapshot`
Choice which one ? Sir.

mweststrate:
2 😊

mweststrate (approved):
Looking great! Thanks for making these changes. As a really nit you might want to mention in the changelog that it will reduce memory usage of MobX in production builds, which is why some projects might be interested in upgrading :)

mweststrate:
Fun, `_mapid` was already unused since 2016 :) https://github.com/mobxjs/mobx/commit/6bf6da97e497a5dead7a9b5d8a75fb3f2e682e79

iChenLei:
> Fun, `_mapid` was already unused since 2016 :) [6bf6da9](https://github.com/mobxjs/mobx/commit/6bf6da97e497a5dead7a9b5d8a75fb3f2e682e79)

That's funny, I was a sophomore college student in 2016, after 5 years I delete it. hah O(∩_∩)O~~

iChenLei:
> Looking great! Thanks for making these changes. As a really nit you might want to mention in the changelog that it will reduce memory usage of MobX in production builds, which is why some projects might be interested in upgrading :)

So you means add `reduce memory usage` info to changset ? Sir

iChenLei:
Sir, I think I should do the same work for `mobx4and5` branch. I will create anthor new pull request for it.

FredyC:
> Sir, I think I should do the same work for `mobx4and5` branch. I will create anthor new pull request for it.

Publishing older versions had to be done manually. I don't think this change can be classified as "critical bugfix". It's not worth the hassle imo and as Michel said, people might be at least motivated to upgrade to V6 which is our goal here.

iChenLei:
> > Sir, I think I should do the same work for `mobx4and5` branch. I will create anthor new pull request for it.
> 
> Publishing older versions has to be done manually. I don't think this change can be classified as "critical bugfix". It's not worth the hassle imo and as Michel said, people might be at least motivated to upgrade to V6 which is our goal here.

Ok , I got it. 😄

mweststrate:
@iChenLei yes that was about the changeset indeed :)

FredyC:
@iChenLei Perhaps be more explicit in the changeset and mention the removal of `mapid_` specifically in case someone was _crazy_ enough to use that for something.

FredyC:
Thanks, let's merge.

urugator:
Is it an intention to pass `__DEV__ ? dynmic : static` to every `createAtom`/`new ObservableValue` invocation instead of just ignoring the passed value inside `Atom` constructor?
I am asking because I think you have missed few places and also you do ignore `name` in `ComputedValue` constructor (as apposed to Atom/ObservableValue).

PR #2785, opened 2021-02-05
add parseTsAsNonJsx flag for mobx-undecorate

### LINKED ISSUE #2754: [mobx-undecorate] need option to treat parse ts file as non-jsx
One example with error
```
import { useSearch } from "./useSearch";

export function useAlias(): string {
    const parsedSearch = useSearch();
    return (<string>parsedSearch.alias || "").toUpperCase();
}
```
Error message: `Transformation error (Unterminated JSX contents (5:20))
SyntaxError: Unterminated JSX contents (5:20)
`

There is no error if type casting is removed
```
import { useSearch } from "./useSearch";

export function useAlias(): string {
    const parsedSearch = useSearch();
    return (parsedSearch.alias || "").toUpperCase();
}
```

In my project jsx code is only in files with tsx extension. There should be an option to let the parser don't parse js/ts files as jsx

iChenLei:
### Code change checklist

### PR Target
fix #2754 

### Summary
Add a new cli flag `parseTsAsNonJsx`, If add the flag we will parse ts file as `non-jsx` and other files as same as before.
```bash
npx mobx-undecorate --parseTsAsNonJsx
```


mweststrate (approved):
Looking great, thanks for picking this up!

Issue #2751, opened 2021-01-28
TypeError: Cannot convert a Symbol value to a string when making production build


markrawls:
**Intended outcome:**

I am able to compile my NextJS or other React project while `NODE_ENV=production`.

**Actual outcome:**

I receive the following error:

```
TypeError: Cannot convert a Symbol value to a string
    at b (/code/node_modules/mobx/dist/mobx.cjs.production.min.js:1:1455)
    at n.i.O (/code/node_modules/mobx/dist/mobx.cjs.production.min.js:1:40579)
    at Function.q [as u] (/code/node_modules/mobx/dist/mobx.cjs.production.min.js:1:5046)
    at n.i.u (/code/node_modules/mobx/dist/mobx.cjs.production.min.js:1:39408)
    at s (/code/node_modules/mobx/dist/mobx.cjs.production.min.js:1:46570)
    at Array.forEach (<anonymous>)
    at exports.makeAutoObservable (/code/node_modules/mobx/dist/mobx.cjs.production.min.js:1:46618)
    at new config_ConfigStore (/code/.next/server/pages/_app.js:195:49)
    at new stores_RootStore (/code/.next/server/pages/_app.js:211:19)
    at Object.sYqq (/code/.next/server/pages/_app.js:223:15)
    at __webpack_require__ (/code/.next/server/pages/_app.js:23:31)
    at Module.1TCz (/code/.next/server/pages/_app.js:113:72)
    at __webpack_require__ (/code/.next/server/pages/_app.js:23:31)
    at Object.0 (/code/.next/server/pages/_app.js:99:18)
    at __webpack_require__ (/code/.next/server/pages/_app.js:23:31)
    at /code/.next/server/pages/_app.js:91:18
ierror Command failed with exit code 1.
```

**How to reproduce the issue:**

I've created a minimal example repository on GitHub that exhibits this behavior here: https://github.com/markrawls/mobx-symbol-error-example

If you manually downgrade MobX to 6.0.5 or 6.0.4, it works as expected. If you forcibly set `NODE_ENV=development` (or any non `production` string) for the build process, the build completes as expected. This is obviously suboptimal as we want to produce a production bundle.

**Versions**

The exclusively effects MobX 6.1.0. Downgrading to 6.0.4 or 6.0.5 resolves this issue.

zhangchao828:
+1  
use `toJS`

urugator:
[...]

PR #2799, opened 2021-02-12
Fix debug names + terser `unsafe: false`

urugator:
-   fix: user provided debug names are not preserved on production
-   fix: property atom's debug name is dynamic on production
-   fix: `observable(primitive, options)` ignores `options`
-   fix: `getDebugName(action)` throws `[MobX] Cannot obtain atom from undefined`
-   [fix: terser using `unsafe: true`](https://github.com/mobxjs/mobx/issues/2751#issuecomment-778171773)


mweststrate (approved):
Thanks!

PR #2802, opened 2021-02-14
Fire observe callback immediately when computed value changes

### LINKED ISSUE #2795: Observing a computed value using observe() should trigger immediately
**Intended outcome:**

When observing a computed value using [`observe`](https://mobx.js.org/intercept-and-observe.html#observe), the listener should be called immediately and ignore transactions. This is also how `observe` behaves when observing an observable value.

**Actual outcome:**

When observing a computed value using [`observe`](https://mobx.js.org/intercept-and-observe.html#observe), the listener is only called after the outermost transaction has completed. There is no difference between using `observe` and `autorun`/`reaction` to observe a computed value. In fact, `ComputedValue.observe_` uses `autorun` internally.

I'm aware that `observe` is a low-level utility. There are good reasons to use it in some rare cases though. Besides, its inconsistent behavior is unexpected and unintuitive, and it makes `observe` for computed values redundant.

I'd be happy to work on a PR, but my first attempt worked only partly, so some guidance would be helpful.

```ts
test("observing a computed value reacts immediately during transaction", () => {
    const events: string[] = []

    const v = observable.box(0)
    const c = computed(() => v.get())

    c.observe_(e => {
        events.push(`observe ${e.oldValue} -> ${e.newValue}`)
    })

    runInAction(() => {
        events.push("action start")
        v.set(1)
        v.set(2)
        events.push("action end")
    })

    expect(events).toEqual([
        "action start",
        "observe 0 -> 1",
        "observe 1 -> 2",
        "action end"
    ])
})
```

**Versions**

[Latest MobX repository state](https://github.com/mobxjs/mobx/tree/9b195b17bd661b9c0c4ab3a8ef323e23c2f118e4) at the time of this writing.

sisp:
When observing observable state, `observe` fired the callback immediately (ignoring transactions) when the state changed, but when observing a computed value the callback was fired only after the outermost transaction had completed. This PR makes `observe`'s behavior consistent for both observable state and computed values.

In case this PR is considered a breaking change and not a fix, the `observe_` method of `IComputedValue` (and similarly the `observe` function) could be overloaded:

```ts
observe_(listener: (change: IComputedDidChange<T>) => void, fireImmediately?: boolean): Lambda
observe_(
  listener: (change: IComputedDidChange<T>) => void,
  options?: {
    fireImmediately?: boolean
    eager?: boolean
  }
): Lambda
```

This way, by default `autorun` would be used (as before) which respects transactions whereas `observe(someComputedValue, change => { ... }, { eager: true })` would use the new behavior that is consistent with using `observe` on observable state.

Fixes #2795.

### Code change checklist

There are 2 performance drops, I'm not sure whether they can be avoided or not.

### DIFF
--- a/packages/mobx/__tests__/v4/base/observables.js
+++ b/packages/mobx/__tests__/v4/base/observables.js
@@ -64,7 +64,76 @@ test("basic2", function () {
     expect(mobx._isComputingDerivation()).toBe(false)
 })
 
-test("computed with asStructure modifier", function () {
+test("computed with asStructure modifier - reaction", function () {
+    const x1 = observable.box(3)
+    const x2 = observable.box(5)
+    const y = m.computed(
+        function () {
+            return {
+                sum: x1.get() + x2.get()
+            }
+        },
+        { compareStructural: true }
+    )
+    const b = []
+    m.reaction(
+        () => y.get(),
+        v => b.push(v),
+        { fireImmediately: true }
+    )
+
+    expect(8).toBe(y.get().sum)
+
+    x1.set(4)
+    expect(9).toBe(y.get().sum)
+
+    m.transaction(function () {
+        // swap values, computation result is structurally unchanged
+        x1.set(5)
+        x2.set(4)
+    })
+
+    expect(b).toEqual([{ sum: 8 }, { sum: 9 }])
+    expect(mobx._isComputingDerivation()).toBe(false)
+})
+
+test("computed with asStructure modifier - reaction + get computed value", function () {
+    const x1 = observable.box(3)
+    const x2 = observable.box(5)
+    const y = m.computed(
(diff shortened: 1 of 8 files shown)


FredyC:
Please, include the changeset according to the previous comment.

mweststrate:
The problem with this PR is that it hacks the reaction mechanism and semantic guarantees around that. We really don't want to maintain an alternative reaction mechanism. A solution in this direction should rather rely on not using reactions as all. Beyond that it is unclear why this must be solved in computed, rather than building an own abstractions on top of atoms. Closing this PR as as far as I can see the additional complexity in maintenance and mental model isn't justified by a single known use case, and as the test points out, this approach makes observed computed values behave differently from their non-observed counterparts.

PR #2806, opened 2021-02-16
Annotations refactor

urugator:
**DO NOT MERGE BEFORE #2779**

This is a follow-up to #2779 - reducing code duplication, hopefully saving few bytes and making the code a bit easier to follow.
I made it a seperate PR, because I wanted to keep #2779 focused on functionality changes.


mweststrate:
Ping me after merging #2779, which probably makes the PR a lot more readable (sigh at Github for not having stacked PRs)

urugator:
Ping @mweststrate
To summarize: the prototype traversal and common checks were moved from `Annotation._make` to `ObservableObjectAdministration.make_`.
In order to do that `Annotation.make_` now returns `0|1|2` indicating whether the traversal should stop/continue or the whole operation should be cancelled (due to interceptor).
Didn't use enum, the way it's transpiled seemed like overkill to me.

packages/mobx/src/api/annotation.ts
```diff
@@ -1,8 +1,17 @@
...
+
 export type Annotation = {
     annotationType_: string
-    make_(adm: ObservableObjectAdministration, key: PropertyKey): void
+    make_(
+        adm: ObservableObjectAdministration,
+        key: PropertyKey,
+        descriptor: PropertyDescriptor,
+        source: object
+    ): 0 | 1 | 2
```
mweststrate, line 14: Nit: you could introduce a type for this: `export type MakeResult = 0 | 1 | 2`

mweststrate (approved):
Love it! I think this is a great clean up of repeated reflection / wiring logic.

packages/mobx/src/api/annotation.ts
urugator: I would like to:
`export type MakeResult = MAKE_CANCEL | MAKE_BREAK | MAKE_CONTINUE`
but obviously not possible as these are runtime...

Perhaps I can do it like:
```javascript
type MakeResultCancel = 0
type MakeResultBreak = 1
type MakeResultContinue = 2

export const MAKE_CANCEL: MakeResultCancel = 0
export const MAKE_BREAK: MakeResultBreak = 1
export const MAKE_CONTINUE: MakeResultContinue = 3

export type MakeResult = MakeResultCancel | MakeResultBreak | MakeResultContinue
```
Or am I overengineering it?

mweststrate: @urugator  pretty sure this will work:

```
export const MAKE_CANCEL = 0
export const MAKE_BREAK = 1
export const MAKE_CONTINUE = 2

export type MakeResult = typeof MakeResultCancel | typeof MakeResultBreak | typeof MakeResultContinue
```

But yeah we're kinda reinventing enums :) https://www.typescriptlang.org/tsconfig#references-to-const-enum-members

urugator: Oh cool. I will use `const enum` then. I just wanted to avoid the generated object.

PR #2817, opened 2021-02-18
spawn "jscodeshift" directly instead of rel path

jeremy-coleman:
The documentation on node's child_process spawn resolution lookups is a bit lacking. I think it generally checks for systme path -> then starts the node require.resolve algorithm found here but when using string specifiers directly, the burden is on the package manager to correctly setup the bin symlinks. See https://docs.npmjs.com/cli/v7/configuring-npm/package-json#bin for a very underwhelming explanation. The take-away is that you should be able to reliably call a package by name if it is using a bin field in its package.json (as all node cli applications do).

I also added an additional help message to the cli if users pass a --help arg. jscodeshift usage will be appended to the mobx-undecorate message

further only partially relevant info:
CJS require.resolve algorithm psuedo code: https://nodejs.org/api/modules.html#modules_all_together
ESM resolve algorithm https://nodejs.org/api/esm.html#esm_resolver_algorithm (likely 0 importance)
https://docs.npmjs.com/cli/v7/commands/npx
https://docs.npmjs.com/cli/v7/commands/npm-exec


jeremy-coleman:
This is an image of the additional --help message instructions
image

FredyC:
Please include the changeset as per the comment above.

jeremy-coleman:
@FredyC I walked through the cli which generated the changelog md and I bumped it as a minor to v1.0.8. First time using the changelog cli, is it correct?

FredyC:
Well, it gives you choice for patch/minor. I think in this case it's probably "patch", but it doesn't matter that much since there is no public API to consider.

However, most importantly, it seems you have messed it up a bit. For once you have deleted `yarn.lock` and your changes are gone as well. Not sure how you managed that :)

jeremy-coleman:
deleting this, and just submitted a different branch to merge, this history got messed up somehow from editing the cli file directly on github then trying to merge in the changeset from my machine. idk, fixed it in a fresh pr

PR #2828, opened 2021-02-24
Update api.md

davidcim:
It seems that these two descriptions are interchanged.

### Code change checklist


urugator (approved):
lgtm

PR #2857, opened 2021-03-22
Fix pathing issue in mobx-undecorate

benjamingr:
When I ran mobx-undecorate I got:

```js
node_modules/.bin/jscodeshift: No such file or directory
```

Because it was already in node_modules (since the instructions say one should run `npx mobx-undercorate`).

This fixed it (locally). Not sure if this is a general fix, another approach would be to `stat` both paths first or run `npm bin jscodeshift` first.

### Code change checklist

### DIFF
--- a/packages/mobx-undecorate/cli.js
+++ b/packages/mobx-undecorate/cli.js
@@ -3,7 +3,7 @@ const { spawn } = require("child_process")
 const path = require("path")
 
 spawn(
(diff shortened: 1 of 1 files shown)


benjamingr:
https://github.com/mobxjs/mobx/pull/2818/files looks like a better fix probably

mweststrate:
Yes, just merged that one!

PR #2858, opened 2021-03-23
Fix the example code error

feng9017:
If `makeAutoObservable(this)` doesn't occur to the last line of the `Constructor` scope, then the properties assigned initially later will lose reaction. It's better to add "make sure `makeAutoObservable(this)` occur to the last line" to the best practice。

### Code change checklist

### DIFF
--- a/docs/understanding-reactivity.md
+++ b/docs/understanding-reactivity.md
@@ -33,10 +33,10 @@ class Message {
     author
     likes
     constructor(title, author, likes) {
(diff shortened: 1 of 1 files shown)


mweststrate:
The `makeObservable` calls should be first. If it isn't working out in your case, you have to correct your babel / typescript config to compile unassigned fields in a spec compliant way: https://mobx.js.org/installation.html#use-spec-compliant-transpilation-for-class-properties

PR #2872, opened 2021-04-01
Fix #2859: Trace log only if derivation is actually about to re-run 

### LINKED ISSUE #2859: Trace logs when derivation is not run and with wrong stale dependency.
Mobx trace is triggered on POSSIBLE staleness of dependencies. If a possibly stale dependency turns out to resolve to the old value, trace in observers was already triggered but the observer is not executed. This can be confusing as a whole dependency tree can be trace-logged even if nothing changes/gets rerun.
Moreover, if the observer becomes actually stale because of some other dependency, this is not traced anymore. This can be confusing as the actual reason for staleness/rerunning stays unknown.

**Intended outcome:**

Trace only logging when the derivation (computed/autorun/reaction) is _actually_ stale.
Trace logging which (if possible even multiple) dependencies _actually_ caused the derivation to be stale.

**Actual outcome:**

Trace logs the first time a derivation becomes _possibly_ stale.
Trace logs which dependency was first to make it _possibly_ stale.

See console and comments in this jsfiddle:
https://jsfiddle.net/jasperh/apgwjsm8/38/

**Versions**

Mobx 4, 5 & 6

urugator:
Moved trace logging from `propagateMaybeChanged` to `propagateChangeConfirmed`. As a result, user should see trace output only if derivation is forced to actually re-evalutate the function provided by user.

Refactored trace tests.

Not implemented: 
It could be helpful to log any state change that would lead to invalidation, not just the first one, see:
https://github.com/mobxjs/mobx/issues/2859#issuecomment-812055816

Fixes: #2859


mweststrate (approved):
Looking good! Thanks for getting to the bottom of this

PR #2873, opened 2021-04-01
Fix #2871 toJS throws with configure({ useProxies: "ifavailable" })

### LINKED ISSUE #2871: Using toJS inside observer component crashes in ES5 mode
**Intended outcome:**
Should be able to use toJS inside observer component

**Actual outcome:**
Mobx crashes with the message
```
[MobX] MobX is currently configured to be able to run in ES5 mode, but in ES5 MobX won't be able to iterate keys to detect added / removed properties. Use `keys` from 'mobx' instead.
```

call
```
configure({
  useProxies: "ifavailable"
});
```
then use toJS inside an observer component 

Sandbox:
https://codesandbox.io/s/immutable-leaf-lcoi8?file=/src/App.js

**Versions**
mobx: 6.1.8
mobx-react-lite: 3.2.0

If this is not fixable it should probably be documented. 
There is an example that uses toJS here https://mobx.js.org/react-integration.html#dont-pass-observables-into-components-that-arent-observer and I assume that would crash in ES5 mode too

urugator:
Fixes: #2871 in BC manner.

We may want to reconsider which kind fo keys `toJS` returns.

### DIFF
--- a/packages/mobx/src/api/tojs.ts
+++ b/packages/mobx/src/api/tojs.ts
@@ -1,12 +1,13 @@
 import {
-    keys,
     isObservable,
     isObservableArray,
     isObservableValue,
     isObservableMap,
     isObservableSet,
-    getPlainObjectKeys,
-    die
+    die,
+    IIsObservableObject,
+    $mobx,
+    objectPrototype
 } from "../internal"
 
 function cache<K, V>(map: Map<any, any>, key: K, value: V): V {
@@ -49,10 +50,11 @@ function toJSHelper(source, __alreadySeen: Map<any, any>) {
         return res
     } else {
         // must be observable object
-        keys(source) // make sure keys are observed
         const res = cache(__alreadySeen, source, {})
-        getPlainObjectKeys(source).forEach((key: any) => {
-            res[key] = toJSHelper(source[key], __alreadySeen)
+        ;((source as any) as IIsObservableObject)[$mobx].ownKeys_().forEach((key: any) => {
+            if (objectPrototype.propertyIsEnumerable.call(source, key)) {
+                res[key] = toJSHelper(source[key], __alreadySeen)
+            }
         })
         return res
     }
--- a/.changeset/fluffy-coats-fry.md
+++ b/.changeset/fluffy-coats-fry.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/mobx/src/api/tojs.ts
```diff
@@ -49,10 +50,11 @@ function toJSHelper(source, __alreadySeen: Map<any, any>) {
         return res
     } else {
         // must be observable object
-        keys(source) // make sure keys are observed
         const res = cache(__alreadySeen, source, {})
-        getPlainObjectKeys(source).forEach((key: any) => {
-            res[key] = toJSHelper(source[key], __alreadySeen)
+        ;((source as any) as IIsObservableObject)[$mobx].ownKeys_().forEach((key: any) => {
```
mweststrate, line 54: I think the fix here is incorrect; we should use `keys(source)` as basis for the loop, and _not_ use `getPlainObjectKeys`, to make sure that `toJS` will track key additions, even when it is used in ES5 mode. There are probably subtle differences between `keys` and `getPlainObjectKeys` that might need to be addressed, but we should use `keys` as basis, or keep it as implemented here, but make sure a read to the keys atom is registered.

mweststrate (changes requested):
See comment, I think the proposed fix is incorrect; even in ES5 mode, toJS should still be reacting to key addition / removals, and I don't think that is the case with the current solution? (We might be missing an explicit test for that)

packages/mobx/src/api/tojs.ts
urugator: `adm.ownKeys_` handles the subscription to keys atom the same way as `keys(object)`. We simply don't have `ownKeys_` exposed via `object-api` so I call it directly on administration.

urugator: Just to clarify further, there is no longer any interesting logic (like keyAtom subscriptions) in object-api or in proxy handler. They simply delegates to `adm` and everything is done there.

mweststrate: ok great, that should work!

mweststrate (approved):

packages/mobx/src/api/tojs.ts
mweststrate: Leaving the merging up to you, in case you want to merge the open PR's in a specific order :)

PR #2878, opened 2021-04-08
Fix #2876: Add `defineProperty` and `ownKeys` to object-api

### LINKED ISSUE #2876: Object.defineProperty and useProxies: "ifavailable"
**Intended outcome:**
A way to define a property on an observable object with `useProxies: "ifavailable"` and without error.

**Actual outcome:**
Mobx with `useProxies: "ifavailable"` gives an error in dev mode when trying to set an observable object property via `Object.defineProperty`.

Obviously in this case the error is relevant, the problem is with its description.

`Use 'defineProperty' from 'mobx' instead.`

But there is no `defineProperty` export in mobx. I think at least the description should be corrected or such method should be provided.
And I haven't found an easy way to suppress this error if I'm sure of what I'm doing other than disabling `useProxies: 'ifavailable'`.

Specifically, `mobx-state-tree` uses `Object.defineProperty` to store some internal data. And working with `mobx-state-tree` in "ifavailable" mode is currently impossible.
https://codesandbox.io/s/mobx-state-tree-todolist-ifavailable-t9w2z?file=/index.js
Maybe I should post an issue to MST too, but it doesn't seem to make much sense without offering a solution to replace  `Object.defineProperty` with something else.

https://codesandbox.io/s/mobx-defineproperty-16wyc?file=/src/index.js

**Versions**
"mobx": "^6.1.8"

urugator:
Fixes #2876

Since we don't have a clear consensus yet, this PR demonstrates my current solution and awaits further instructions/discussion.
Original intention was to also include `getOwnPropertyNames` and `getOwnPropertySymbols` (as these would start throwing on "ifavailable", but works for "never|always")


packages/mobx/src/errors.ts
```diff
@@ -3,6 +3,8 @@ const niceErrors = {
     1(annotationType, key: PropertyKey) {
         return `Cannot apply '${annotationType}' to '${key.toString()}': Field not found.`
     },
+    3: "'ownKeys()' can only be used on observable objects",
```
mweststrate, line 6: NIT: if these numbers wheren't here yet, this were error codes used in the past, so it is better not to recycle them in case someone is looking for them (in which case they need to go back in history of this file, I guess we should have left a comment when removing, but alas)

mweststrate (changes requested):
Ok, looks consistent and a minimal change, feel free to go ahead :) I guess the proxy trap should be directed here in the `proxies: always` case?

packages/mobx/src/errors.ts
urugator: I will take a look if I can revive these removed in previous PRs (not sure if I hadn't already repurposed some).
What's the recommended practice here? Keeping the error as is and providing comment that it's no longer used?

mweststrate: Just leaving them in the source code, but commented out so that they don't increase bundles unnecessarily. Hopefully the remain at least searchable in GH history and issues if the numbers are not recycled. I have no idea if that actually ever happens, so don't worry too much about it :)

urugator:
... forgot the changeset ...

PR #2904, opened 2021-04-19
Fix #2903: toJS works with computed value props

### LINKED ISSUE #2903: toJS fails for observable object with computed value prop
**Intended outcome:**
toJS converts observable objects with computed value props just fine.

**Actual outcome:**
toJS fails with following exception:

```
TypeError: Cannot read property 'ownKeys_' of undefined
at toJSHelper (node_modules/mobx/src/api/tojs.ts:55:17)
      at propertyIsEnumerable (node_modules/mobx/src/api/tojs.ts:57:14)
          at Array.forEach (<anonymous>)
      at toJSHelper (node_modules/mobx/src/api/tojs.ts:55:33)
      at propertyIsEnumerable (node_modules/mobx/src/api/tojs.ts:57:14)
          at Array.forEach (<anonymous>)
      at toJSHelper (node_modules/mobx/src/api/tojs.ts:55:33)
      at Object.toJS (node_modules/mobx/src/api/tojs.ts:69:2)
```

Run the following code (-> unit test):
```
const a = observable({ key: "value" });
const c = observable({ computedValue: mobx.computed(() => a.key) });

const j = mobx.toJS(c);
expect(j).toMatchObject({ computedValue: "value" });
```

**Versions**
mobx 6.2.0 (but I guess all previous as well)

ahoisl:
### Code change checklist

### DIFF
--- a/.changeset/calm-wolves-cover.md
+++ b/.changeset/calm-wolves-cover.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Make toJS work with computed value props
--- a/packages/mobx/__tests__/v4/base/tojs.js
+++ b/packages/mobx/__tests__/v4/base/tojs.js
@@ -356,3 +356,11 @@ test("map to JS", () => {
     }
     new MyClass()
 })
+
+test("Correctly converts observable objects with computed values", () => {
+    const a = observable({ key: "value" })
+    const c = observable({ computedValue: mobx.computed(() => a.key) })
+
+    const j = mobx.toJS(c)
+    expect(j).toMatchObject({ computedValue: "value" })
+})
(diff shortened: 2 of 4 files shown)


urugator:
Thanks for the PR! Currently we ignore (non-boxed) `computed` props, so there is possibly some inconsistency here, however I am not sure how it would make sense for boxed computed eg when part of array or when called on directly `toJS(boxedComputed)`- should it replace the value with `undefined`?? I think this fix is fine, but let's double check with @mweststrate

ahoisl:
@urugator uh, ok, I just had this use case that's why I thought I might bring it to your attention and provide a possible fix. Just take your time and check consistency.

I'm also willing to invest some more time to fix inconsistencies if there are any, but you would need to direct me to them 😉

mweststrate (approved):
Thank for the PR. LGTM!

PR #2905, opened 2021-04-20
Fix broken links

ygnoh:
### Code change checklist

Some links were broken. so fixed it


urugator:
Thank you!

PR #2906, opened 2021-04-21
Provide flow.bound annotation/decorator

urugator:


mweststrate (approved):
Looks brilliant! Thanks for the detailed tests! Super cool that the actual change is so small in the new decoration setup.

PR #2923, opened 2021-04-28
Document flow.bound + few doc improvements

urugator:


mweststrate (approved):
Thanks for updating the docs!

PR #2927, opened 2021-05-05
Give proper typing to opts.equals of reaction

upsuper:
This PR gives the `opts.equals` proper typing so that TypeScript knows that the `equals` function passed to it is comparing values of the return type of `expression` rather than `any`.

### DIFF
--- a/packages/mobx/__tests__/v5/base/typescript-tests.ts
+++ b/packages/mobx/__tests__/v5/base/typescript-tests.ts
@@ -2055,6 +2055,23 @@ test("type inference of the action callback", () => {
     }
 })
 
+test("TS - type inference of reaction opts.equals", () => {
+    const data = observable({ a: 23 })
+    mobx.reaction(
+        () => data.a,
+        a => {},
+        {
+            equals: (a, b) => {
+                assert<IsExact<typeof a, string>>(false)
+                assert<IsExact<typeof b, string>>(false)
+                assert<IsExact<typeof a, number>>(true)
+                assert<IsExact<typeof b, number>>(true)
+                return a === b
+            }
+        }
+    )
+})
+
 test("TS - it should support flow as function wrapper", done => {
     const values: number[] = []
 
--- a/packages/mobx/src/api/autorun.ts
+++ b/packages/mobx/src/api/autorun.ts
@@ -86,14 +86,14 @@ export function autorun(
     return reaction.getDisposer_()
 }
 
-export type IReactionOptions = IAutorunOptions & {
+export type IReactionOptions<T> = IAutorunOptions & {
     fireImmediately?: boolean
-    equals?: IEqualsComparer<any>
+    equals?: IEqualsComparer<T>
 }
 
 const run = (f: Lambda) => f()
 
-function createSchedulerFromOptions(opts: IReactionOptions) {
+function createSchedulerFromOptions(opts: IAutorunOptions) {
     return opts.scheduler
         ? opts.scheduler
         : opts.delay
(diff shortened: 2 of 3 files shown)


iChenLei:
Hi, Thanks for your PR. Can you add some description for your PR and add required `changeset` file.

upsuper:
@iChenLei Thanks for your reply. I've updated the description and included a changeset file.

iChenLei:
Can you add a type unit test like follow ?
```typescript
// just a code example, don't copy directly
test("TS - it should infer reaction's opts.equals type", () => {
    const data = observable({ a: 23 });
    const dispose = mobx.reaction(() => {
        return data.a;
    }, (data) => {
        console.log(data);
    }, {
        fireImmediately: true,
        equals: (pre, next) => {
            assert<IsExact<typeof pre, number>>(true)
            assert<IsExact<typeof next, number>>(true)
            return pre === next
        }
    })
});
``` 
Mobx Type UT is located at **`packages/mobx/__tests__/v5/base/typescript-tests.ts`**
Write a typing unit test is an effective means of avoiding regression, similar PR -> https://github.com/mobxjs/mobx/pull/2740

截屏2021-06-16 14 13 02

packages/mobx/src/api/autorun.ts
```diff
@@ -86,14 +86,14 @@ export function autorun(
...
+export type IReactionOptions<T> = IAutorunOptions & {
     fireImmediately?: boolean
-    equals?: IEqualsComparer<any>
+    equals?: IEqualsComparer<T>
 }
 
 const run = (f: Lambda) => f()
 
-function createSchedulerFromOptions(opts: IReactionOptions) {
+function createSchedulerFromOptions(opts: IAutorunOptions) {
```
iChenLei, line 96: Why we need replace `IReactionOptions` with `IAutorunOptions`?

upsuper: Because it's also used in `autorun` above, and I had some trouble getting the generic type work properly here. This function doesn't care about anything outside `IAutorunOptions` anyway.

packages/mobx/__tests__/v5/base/typescript-tests.ts
```diff
@@ -2055,6 +2055,23 @@ test("type inference of the action callback", () => {
...
 
+test("TS - type inference of reaction opts.equals", () => {
+    const data = observable({ a: 23 })
+    mobx.reaction(
+        () => data.a,
+        a => {},
+        {
+            equals: (a, b) => {
+                assert<IsExact<typeof a, string>>(false)
+                assert<IsExact<typeof b, string>>(false)
```
upsuper, line 2066: If you don't add these two lines, it seems like TS would just converge the anys into `number` and make the test pass even without the changes below.

packages/mobx/src/api/autorun.ts
iChenLei: @mweststrate Sir, Please code review here if you have time.

iChenLei: /cc @urugator

urugator: Sorry, I am not much into TS stuff, so I do not want to comment on something I do not understand well enough.

mweststrate (approved):
LGTM

packages/mobx/src/api/autorun.ts
mweststrate: Since this is an internal function, whatever goes is fine here :)

Issue #2829, opened 2021-02-25
Component with ref forwarding has wrong displayName when it was wrapped via inject multiple times


lozinsky:
**Intended outcome:**

If target component was wrapped via `React.forwardRef`, then every execution of `inject` shouldn't affect the target `displayName`.

```js
import { forwardRef } from "react";
import { inject } from "mobx-react";

const Foo = forwardRef(function Foo() {});

Foo.displayName = "Foo";

console.log(inject()(Foo).displayName); // inject(Foo)
console.log(inject()(Foo).displayName); // inject(Foo)
console.log(Foo.displayName); // Foo
```

**Actual outcome:**

If target component was wrapped via `React.forwardRef`, then every execution of `inject` adds `inject(*)` prefix to the target `displayName`, because `displayName` property descriptor of `Injector` component refers to the target `displayName` due to `copyStaticProperties` implementation.

```js
import { forwardRef } from "react";
import { inject } from "mobx-react";

const Foo = forwardRef(function Foo() {});

Foo.displayName = "Foo";

console.log(inject()(Foo).displayName); // inject(Foo)
console.log(inject()(Foo).displayName); // inject(inject(Foo))
console.log(Foo.displayName); // inject(inject(Foo))
```

**How to reproduce the issue:**

See console output: https://codesandbox.io/s/mobx-react-inject-forward-ref-display-name-qo244?file=/src/index.js:0-254

FredyC:
Why do you wrap in inject multiple times? Either way, we don't recommend using `inject` anymore, it's an obsolete pattern and has been replaced by React Context. See docs: https://mobx.js.org/react-integration.html#using-external-state-in-observer-components

lozinsky:
Thank you for your suggestion! In our codebase we have some reusable components like `Button` and it can be wrapped in `inject` multiple times on a page, but in different places of application, e.g:
```js
const MyAFeatureButton = inject()(Button);

// ...

const MyBFeatureButton = inject()(Button);
```

FredyC:
I think we won't be changing the behavior in this case as it could be a breaking change for some people who expect multiple wrapping (for whatever reason). If it truly pains you, implements your own `inject` variant with `React.useContext(MobXProviderContext)` (exported from package).

lozinsky:
[...]

PR #2930, opened 2021-05-07
inject shouldn't change original displayName of component that uses forwardRef

vkrol:
I think it's best to use `hoist-non-react-statics` https://github.com/mridgway/hoist-non-react-statics, but it was intentionally removed some time ago.

Fixes: #2829.


iChenLei:
Thanks for your PR , I add mweststrate as reviewer.

mweststrate (approved):
LGTM. Thanks!

PR #2972, opened 2021-06-08
Polyfill `Object.is` for old/unsupported browsers

kk-gjyang:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#polyfill

### DIFF
--- a/.changeset/spotty-cheetahs-invite.md
+++ b/.changeset/spotty-cheetahs-invite.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Polyfill `Object.is` for old/unsupported browsers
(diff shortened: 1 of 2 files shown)


mweststrate:
Polyfilling is the responsibility of the hosting setup, not of this library, but since this is a very small one, I'm ok merging it. Would you mind adding a comment to the code to signal this is an IE specific polyfill? That will help future cleanups.

kk-gjyang:
Thanks @mweststrate. I'm currently working on a project that needs to work with Chromium browser(embedded in device) which version is less than 30 and the app throws errors because `mobx` is a part of the app.

PR #2976, opened 2021-06-10
Add support for flow

masylum:
No intention to merge since there is clear intention to not support flow.


mweststrate:
Marking as draft since there is no merge intention. If the changes are limited I think it is fine to release the code mod with flow support. But that is impossible to tell now since the code is formatted differently :).

masylum:
Sorry about that! I've been keeping my branch here where I'm indeed modifying way more things than just flow support. No merge intention, sorry.

PR #2980, opened 2021-06-10
Allow child classes of Map as initializers for ObservableMap

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>

### Code change checklist

### DIFF
--- a/packages/mobx/__tests__/v5/base/map.js
+++ b/packages/mobx/__tests__/v5/base/map.js
@@ -934,13 +934,22 @@ test("#1583 map.size not reactive", () => {
     expect(sizes).toEqual([0, 1, 2])
 })
 
-test("#1858 Map should not be inherited", () => {
+test("#1858 Map should not be inherited when decorated", () => {
+    class MyMap extends Map {}
+
+    const map = new MyMap()
+    expect(() => {
+        mobx.observable(map)
+    }).toThrow("Cannot initialize from classes that inherit from Map: MyMap")
+})
+
+test("#2980 Map can be inherited when used in plain function", () => {
     class MyMap extends Map {}
 
     const map = new MyMap()
     expect(() => {
         mobx.observable.map(map)
-    }).toThrow("Cannot initialize from classes that inherit from Map: MyMap")
+    }).not.toThrow()
 })
 
 test("#2274", () => {
--- a/packages/mobx/src/api/observable.ts
+++ b/packages/mobx/src/api/observable.ts
@@ -29,7 +29,8 @@ import {
     assign,
     isStringish,
     createObservableAnnotation,
-    createAutoAnnotation
+    createAutoAnnotation,
+    die
 } from "../internal"
 
 export const OBSERVABLE = "observable"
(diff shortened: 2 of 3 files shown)


Nokel81:
I don't know where to add tests, there only seem to be folders for `v4` and `v5.

mweststrate:
What are you trying with this change? `observable.map` will create a _new_ map, which won't be a `MyMap`. The check is here to avoid surprises that maps are suddenly no `MyMaps` more when made observable. If you are fine with getting only a plain map back, `observable.map(myMap.entries())` should do the trick.

Nokel81:
So I kept the same behaviour for decorators because the types are erased. But if I call `observable.map` (or for that matter `observable.box`) on `new MyMap()`. The typings are such that it wouldn't be a problem.

My use-case does not work for calling `myMap.entries()` because it is a generic function.

From my POV, this doesn't break the original ask from #1858 because that specifically targeted `@observable`.

mweststrate:
My point is more that the following won't work:

```typescript
class MyMap extends Map {

  hi() {
  
  }
}

const origMap = new MyMap()
origMap.hi()

const observableMap = observable.map(origMap)
observableMap.hi() // will throw, doesn't exist
```

Nokel81:
Yes, but from a typescript perspective the compiler will warn you about that. So it is a non-issue.

Nokel81:
I disagree that the check would be simple, because if it is an object then I would have to recursively check every single field and child object.

There is a good reason, namely electron's IPC uses the V8 structured clone serialization library, which throws an error on observable types. So we need to sanitize such types before sending them.

Also, how is this fix changing your API semantics? This value is being used for initial values.

To repeat, the line `@obserbable m: MyMap = new MyMap();` will still throw an error. Which was the original ask in #1858.

Nokel81:
I know that `toJS` is not a serialization utility, I didn't say that, I said that `mobx` doesn't work with V8's structured clone algorithm.

I have read through the documentation and `observable.map` does not say that it "converts map to its observable counterpart". Specifically it says the opposite.

https://mobx.js.org/api.html#observablemap says "Creates a new observable ES6 Map based on the provided `initialMap`."

mweststrate:
Sorry, I'm still totally missing why you convert your MyMap to an observable map, merely to turn it into a plain object. Your probably can fixed your issue with:

```
const value = observable.box(args).get()
toJS(value instanceof MyMap ? observable.map(value.entries()) : value)
```

But to me that makes still little sense since you can use `Array.from(myMap.entries())` or `Object.fromEntries(myMap.entries())` to get the same thing and not involve MobX at all.

mweststrate:
Structured clone only supports plain objects, without things that have custom prototypes / classes, since those cannot be marshalled on the other side. Coincidentally largely the same designed limitation than observable.map has ;-). 

Again toJS is not a generic api to serialize arbitrary classes. It is your `MyMap`, so it is your responsibility to make it serializable. Not MobX's. For MobX Map, which has it own proto as well, we created toJS. But not for _any_ map. Anyway, you got two work arounds offered above, that should do.

Nokel81:
You are incorrect about structured clone, see https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm. `Map` is specifically supported.

mweststrate:
yes, because it is a built in prototype. That is why it is supported by MobX as well. Anyway, to conclude. I don't see any convincing arguments why this is a problem with MobX and not something that should be solved in user land, so this won't be merged.

Nokel81:
Also I would like to note that this also prevents maps from having values that are `T extends Map`

PR #2982, opened 2021-06-10
Allows the force update to be infinitely called

zeachco:
This use case is far-fetched but in some contexts (gaming for example) the number of iterations can go beyond the maximum value supported by javascript and end up being `Infinity` which would cause the force update to stop working
because `Infinity + 1 === Infinity` is true in javascript

I didn't added tests as I suppose we don't want to run a tests with a loop of 9007199254740991+ iterations and had some trouble mocking the tick value of `useState`
### Code change checklist

-   [ ] ~~Updated `/docs`. For new functionality, at least `API.md` should be updated~~

### DIFF
--- a/packages/mobx-react-lite/src/utils/utils.ts
+++ b/packages/mobx-react-lite/src/utils/utils.ts
@@ -3,11 +3,9 @@ import { useCallback, useState } from "react"
 const EMPTY_ARRAY: any[] = []
 
 export function useForceUpdate() {
-    const [, setTick] = useState(0)
+    const [, setRef] = useState({})
 
-    const update = useCallback(() => {
-        setTick(tick => tick + 1)
-    }, EMPTY_ARRAY)
+    const update = useCallback(() => setRef(() => ({})), EMPTY_ARRAY)
 
     return update
 }


mweststrate:
Fair points :)

zeachco:
Also to mention, it's the first time I contribute to this project, I couldn't find the `npm run perf` as it seems to be only for mobx itself but I used fastest way known to me to "reset" the loop without using bit-shifting 😅

urugator:
I think `useState` uses strict comparison, can't we just `setState(NaN)`?

urugator:
Btw should we worry about
https://github.com/mobxjs/mobx/blob/eb4fb5d61403aecbbb2546dd154a2e5431128ce5/packages/mobx/src/core/globalstate.ts#L54 ?

`mobxGuid` seems to be used only for debug names, therefore harmless I suppose?

zeachco:
> I think `useState` uses strict comparison, can't we just `setState(NaN)`?

I think that would even be better, I'll give it a try when I have a chance

mweststrate:
Hope that they will never use Object.is in that case, as that would make
everything fall apart :')

zeachco:
So I tried using `NaN` but it doesn't work
for experiments, I played around with https://codesandbox.io/s/useforceupdate-cu3nh?file=/src/index.js:943-960

I think using something like this would be the best way since `{} !== {}`
```javascript
function useForceUpdate() {
  const [, setState] = useState({});
  return () => setState(() => ({}));
}
```

zeachco:
> Hope that they will never use Object.is in that case, as that would make everything fall apart :')

Spotted on! They do use a copy of the pollyfil 🤷  
https://github.com/facebook/fbjs/blob/c69904a511b900266935168223063dd8772dfc40/packages/fbjs/src/core/shallowEqual.js#L29

urugator:
Ah it's [noted here](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update), could't find it before (it's not mentioned in [useState docs](https://reactjs.org/docs/hooks-state.html)...).
[Here is explained how comparison/bailout works](https://github.com/facebook/react/issues/20817#issuecomment-778672150) (it states `===`, but whatever).  

> I think using an object reference would work better since {} !== {} 

👍 (without the extra closure)

zeachco:
> [Here is explained how comparison/bailout works](https://github.com/facebook/react/issues/20817#issuecomment-778672150) (it states `===`, but whatever).
> 

Maybe I missed something, I understand what's being said but it doesn't seems to be the actual behaviour as of react 16.8.0
```javascript
function useForceUpdateNoValue() {
  const [, setState] = useState();
  return () => setState();
}
```
it doesn't update, I'd like to blame the [shallowEqual](https://github.com/facebook/fbjs/blob/c69904a511b900266935168223063dd8772dfc40/packages/fbjs/src/core/shallowEqual.js#L29) mentioned above

> 
> 👍 (without the extra closure)

no sure what you mean by extra closures, talking about the `() => ({})` ? It's an object construction and not a closure since it's wrapped in parentheses

Or was it the initial value of `useState` ?
It's true we don't need `useState({})` and we could just have `useState()` but if react changes their compare algo, it would be easier to debug something that happens 100% rather than only on the second re-render 🤔 
but that's a personal opinion and I'm very open changing it to with or without 🤷

urugator:
> it doesn't seems to be the actual behaviour

Hm weird indeed

>no sure what you mean by extra closures

`setRef(() => ({}))` => `setRef({})`

zeachco:
Ok, my example was weak, we can see [this mobx test](https://app.circleci.com/pipelines/github/mobxjs/mobx/1564/workflows/7138050f-a96e-46cd-a3d3-bbe6b8e3f8b0/jobs/7672) failing about it while not using the react dispatcher instead of a direct value set

urugator:
> I though there was a reason at first mobx did this

I think the only reason was the need for previous value to perform increment :)

Toggle would rely on comparison happening immediately (rather than sometimes later), dunno if it's a safe choice...

urugator:
Pushed a PR #2983 as it was faster than explaning.

zeachco:
thanks, I was wondering about the `useCallback` as well 👍 
you solution looks better so I'll close this PR

PR #2983, opened 2021-06-11
forceUpdate fix and refactor, alternative to #2982

urugator:
1. **removed `useForceUpdate` abstraction**
It was used on single place and impl is trivial.
2. **removed `useCallback`**
`forceUpdate` isn't passed anywhere where it could break memoization, therefore no need for `useCallback`.
The closure is recreated every render one way or another.
3. **Replace increment with unique reference, see #2982.** 
The choice to use `[]` is arbitrary. Feel free to change to whatever performs better.

### DIFF
--- a/.changeset/plenty-rice-love.md
+++ b/.changeset/plenty-rice-love.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+Allow force update to be called infinitely times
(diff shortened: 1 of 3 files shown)


urugator:
Do I have to do anything in regards to `mobx-react` (since it depends on lite)? Include in changeset or something?

PR #2985, opened 2021-06-13
Remove duplicate global property check

fecqs:
I find that "Symbol" is duplicated and should be removed

### Code change checklist


iChenLei (approved):
LGTM

PR #2995, opened 2021-06-20
[Perf] Reduce useObserver gc pressure

Bnaya:
Tiny performance improvement,  create ObjectToBeRetainedByReact once per component instance and not every component render

### Code change checklist

### DIFF
--- a/.changeset/reduce-useObserver-gc-pressure.md
+++ b/.changeset/reduce-useObserver-gc-pressure.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+Reduce useObserver gc pressure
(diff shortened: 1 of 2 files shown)


gioragutt (approved):

urugator (approved):
add changeset

Bnaya:
@urugator can please have a look on my changeset file?
looks valid?

urugator:
Just in case you're not aware, you can run `yarn changeset` from root dir...

Bnaya:
Donno why, it insists to create major bump, so i had
To edit it manually

urugator:
If you were confused by the controls the same way as I was ... you have to pick choices with spacebar and confirm with enter...

Bnaya:
> If you were confused by the controls the same way as I was ... you have to pick choices with spacebar and confirm with enter...

The only option i had is to make major bump

urugator:
When you hit enter, it should ask for minor, enter again and it should fallback to patch...

PR #2998, opened 2021-06-21
trace(): log when computed becomes suspended

urugator:
`trace` logs when computed becomes suspended as discussed in:
https://github.com/mobxjs/mobx/issues/2859#issuecomment-831184535

`requiresObserver` no longer throws, but warns, so it's consistent with `computedRequiresReaction`, `observableRequiresReaction`, etc

Improved console messages.

### DIFF
--- a/.changeset/loud-rules-count.md
+++ b/.changeset/loud-rules-count.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+trace(): log when computed becomes suspended
--- a/.changeset/loud-rules-count.md
+++ b/.changeset/loud-rules-count.md
@@ -2,4 +2,5 @@
 "mobx": patch
 ---
 
-trace(): log when computed becomes suspended
+`trace()`: log when computed becomes suspended
+`requiresObserver` warns instead of throwing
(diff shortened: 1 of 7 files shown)


.changeset/loud-rules-count.md
```diff
@@ -0,0 +1,6 @@
+---
+"mobx": patch
+---
+
+`trace()`: log when computed becomes suspended
+`requiresObserver` warns instead of throwing
```
mweststrate, line 6: I didn't see this change in the PR?

mweststrate (approved):

mweststrate:
Looking good, thanks!

.changeset/loud-rules-count.md
urugator: typo ... `requiresReaction` ... the condition with die is deleted and merged to condition with warn:
```
 warnAboutUntrackedRead_() {
        if (!__DEV__) return
---     if (this.requiresReaction_ === true) {
---             die(`[mobx] Computed value ${this.name_} is read outside a reactive context`)
---     }
        if (this.isTracing_ !== TraceMode.NONE) {
            console.log(
                `[mobx.trace] '${this.name_}' is being read outside a reactive context. Doing a full recompute`
                `[mobx.trace] Computed value '${this.name_}' is being read outside a reactive context. Doing a full recompute.`
            )
        }
---     if (globalState.computedRequiresReaction) {
+++     if (globalState.computedRequiresReaction || this.requiresReaction_) {
            console.warn(
---            `[mobx] Computed value ${this.name_} is being read outside a reactive context. Doing a full recompute`
+++            `[mobx] Computed value '${this.name_}' is being read outside a reactive context. Doing a full recompute.`
            )
        }
    }
```

PR #3001, opened 2021-06-22
doc: babel assumptions

urugator:
https://babeljs.io/docs/en/assumptions

> Since Babel 7.13.0, you can specify an assumptions option in your configuration to tell Babel which assumptions it can make about your code, to better optimize the compilation result. Note: this replaces the various loose options in plugins in favor of top-level options that can apply to multiple plugins (RFC link).


mweststrate:
Looking good, thanks for adding this!

Issue #2526, opened 2019-02-11
Support concurrent features React / React 18


mweststrate:
Support concurrent mode React. Primary open issue, don't leak reactions for component instances that are never committed.

A solution for that is described in: https://github.com/mobxjs/mobx-react-lite/issues/53

But let's first wait until concurrent mode is a bit further; as there are no docs yet etc, introducing a workaround in the code base seems to be premature at this point (11-feb-2019)

FredyC:
Might be not relevant anymore too, so let's close it till there is someone who wants to investigate this.

mweststrate:
Reopening this as we're getting closer to a React 18 release. Progress can be followed in #3005.

At this moment, as far as initial test projects point out, mobx-react(-lite) plays fine with React 18 in Strict mode. No warnings, errors and features like `startTransaction` seem to work correctly.

At this moment we don't offer MobX specific concurrent features yet. Some tearing could (theoretically) still happen. And MobX might put React in de-opt mode (theoretically) in cases as well. 

That being said, there are no known problems at this moment that would block mobx-react(-lite) from upgrading or using specific React features. 

That conclusion is based on some pretty flimsy research, but at least it is some status update :). No known limitations or blockers for a React 18 upgrade at this moment.

franckchen:
> Relevant thread: https://twitter.com/acemarke/status/1230687035045896192

I had read the code and tried the demo. 

but i think its a little tricky. Not all scenarios are applicable.

any good idea for an common use utility functions or pattern?

ericmasiello:
> Reopening this as we're getting closer to a React 18 release. Progress can be followed in #3005.
> 
> 
> 
> At this moment, as far as initial test projects point out, mobx-react(-lite) plays fine with React 18 in Strict mode. No warnings, errors and features like `startTransaction` seem to work correctly.
> 
> 
> 
> At this moment we don't offer MobX specific concurrent features yet. Some tearing could (theoretically) still happen. And MobX might put React in de-opt mode (theoretically) in cases as well. 
> 
> 
> 
> That being said, there are no known problems at this moment that would block mobx-react(-lite) from upgrading or using specific React features. 
> 
> 
> 
> That conclusion is based on some pretty flimsy research, but at least it is some status update :). No known limitations or blockers for a React 18 upgrade at this moment. 

[...]

PR #3005, opened 2021-06-24
[WIP] Support React 18

mweststrate:
First attempt at full React 18 support.

Didn't get very far since react-testing-library isn't updated yet :) 

* [x] bump deps
* [x] check basic functionality. Seems to work fine and without warnings / errors: https://github.com/mweststrate/currencies-demo/tree/react18
* [x] startTransition seems to work fine as well: https://github.com/mweststrate/platform-app/commit/bdd995773ddc6551235a4d2b0a4c9bd57d30510e
* [ ] test against https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering
* [ ] updated react testing library https://github.com/testing-library/react-testing-library/pull/937/ / https://github.com/testing-library/react-testing-library/issues/509
* [x] replace unstable_batched_updates with flushSync
* [x] make sure all tests are run in strict mode (as well?)
* [ ] verify effect usage, especially against double firing
* [ ] ~differentiate low and high prio updates. always use startTransition by default / optionally, hook for the exceptions? `startTransition` looks to work as expected, see above~
* [ ] think about tearing scenarios
* [ ] keep an eye on https://github.com/reduxjs/react-redux/issues/1732
* [ ] rewrite "eager subscriptions + ugly GC" to a more proper "store atoms+versions read in initial render, and verify after subscribe in useEffect"?
* [ ] explaiin impact of startTransiction
* [ ] check https://twitter.com/acemarke/status/1431319024613220353


mweststrate:
Highlight impact of startTransaction updates vs local state: https://codesandbox.io/s/usedeferredvalue-issue-iteration-forked-ovvsl / https://codesandbox.io/s/usedeferredvalue-issue-iteration-forked-li6uj?file=/src/index.js

zhangchao828:
tearing scenarios

Is this problem difficult to solve in mobx ?

urugator:
> Is this problem difficult to solve in mobx ?

My current understanding (and could be wrong) is that if you want to avoid tearing, the component can not depend on mutable state. I don't think that's reasonably doable with mobx.

airhorns:
https://github.com/testing-library/react-testing-library/issues/509 has been closed with their new alpha now 🥳

zhangchao828:
React 18 is now in beta 
is there any progress

franckchen:
> React 18 is now in beta is there any progress

The author has actually made it clearer .mobx works fine with React 18. And I don’t think we need to be impatient, I dont think online project will use React concurrent mode quickly, i think there will be a lot of works to do. even if project upgrade React to ^18.0.0. i think we still use Legacy Mode or Block Mode. 

and i made some test. yes, there are some tearing problem in cooperating with mobx and react 18 concurrent mode. but the final render result is ok.

urugator:
> replace unstable_batched_updates with flushSync

1. Why is `flushSync` necessary?
2. `flusSync` warns when called from lifecycle methods, so how do we run `action` from eg `useEffect` without warning?

samcooke98:
re; >  test against https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering


I implemented it here; https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering/pull/63/files against the current released version of mobX. 

---

macbem:
@mweststrate what's the status of this? Can we expect a release soon? If there are any "good first issues" to tackle that would help prep this, can you share a list?

mweststrate:
Sorry haven't really been following along React 18 anymore in the past year, so lost most of the context I had on it😅. The primarily thing that needs to happen is to bump the peer deps and unit tests to use React 18, so PR for that is welcome!

noah79:
Is this really true?  We had to switch back to ReactDOM.render() today after hitting cases where mobx observer components wouldn't rerender randomly even though their underlying observables were changing.  Mind you we're on Mobx 5.15.7 + mobx-react-lite 2.2.2.  However this thread doesn't lead me to believe that going through the gargantuan task of updating to latest and greatest mobx will somehow resolve the issue:  https://github.com/mobxjs/mobx/issues/2526

Would appreciate guidance here as we are hugely dependent on mobx (250 occurrences of useLocalStore in the codebase) and switching to useState as is suggested in the above thread sounds worse than a root canal (or 4).

mweststrate:
We didn't have breakages reported / found, when testing with the alpha. If
something breaks for you, please reproduce, isolute and report it so we can
look at it. But at this moment we didn't release anything officially
supporting React 18 yet, since no one was able to yet carve out the time to
update everything, so personally I can't recommend jumping directly on
React 18 yet.

airhorns:
@mweststrate do you or you know any community contributors who do freelancing / consulting? My company would love to do a contract with them to flesh out support! I think there's probably some challenges around all the `useSyncExternalStore` that need to be worked out to avoid tearing problems, I've seen some murky stuff inside `valtio` and `use-query` and some other libraries that do some intense state management outside of react that makes me think its pretty involved, so I'd love to work with someone who knows mobx really well and can learn the ins and outs of that stuff.

noah79:
@kubk I'm aware it's deprecated.  We're still happily on Mobx 5.15.7 and have seen no reason to upgrade and lose decorators, be forced to use makeAutoObservable, etc.  I considered upgrading this week due to React 18 breakages, but nothing I've read here leads me to believe it's going to actually work for all of our complex cases.  Your codepen is super simplistic compared to our running app (A customizable MDI that sells for 7 figures and is used by companies like Swarovski, ASICS, Restoration Hardware, Under Armor and others).  If I _knew_ that it was bulletproof against concurrency mode then I might take the plunge and rewrite the literally hundreds of stores that use mobx, but until I see true concurrency support, it's a ton of work for likely zero tangible benefit.  Hell the benchmarks say Mobx 6 is slower than what we're on!  Decorators are even moving forward in javascript so I have a feeling someday a mobx 7 will come out that will embrace decorators again and the API will (hopefully!) return to the much kinder one that preddated version 6.

I like my decorators.  I dislike needing to call magic methods in my constructors, parent classes, etc.  I like _knowing_ if a class instance variable is actually observable or not (decorated or not), etc.  

It ain't deprecated in v5.15.7 :)

image

mweststrate:
> lose decorators

Decorators are supported in 6 as well

> it's a ton of work

Tried the codemod?

> It ain't deprecated in v5.15.7 :)

You know that that isn't the meaning of the word deprecated ;)

`useAsObservableSource` has some flaws in it that are even more problematic in concurrent mode, so you'd have to get rid of those one way or another in any case I'm afraid.

kubk:
@mweststrate Hi, there is a new Create-React-App release that ships with React 18, so it's probably time  allow installing Mobx alongside React 18. I'd like to simply bump allowed react versions in mobx-react-lite and mobx-react. Should I create a separate PR or continue working here?

packages/mobx-react-lite/package.json
```diff
@@ -39,7 +39,7 @@
     "dependencies": {},
     "peerDependencies": {
         "mobx": "^6.1.0",
-        "react": "^16.8.0 || ^17"
```
kubk, line 42: Isn't it a BC break? Can I change it to `^16.8.0 || ^17 || ^18`?

mweststrate:
Hey! I started working on a local branch, updating tests etc, but forgot to
push and I'm travelling coming week, will try to continue after that again!

mweststrate:
Being superseded by https://github.com/mobxjs/mobx/pull/3387

mweststrate:
Should be fixed in mobx-react-lite@3.4.0 / mobx-react@7.4.0

Issue #3017, opened 2021-06-28
Action is not transactional for computeds, but docs says it should be


jheeffer:
**Intended outcome:**

Computeds will not be recomputed during an action, as per the docs on [`action`](https://mobx.js.org/actions.html) 

> No observers will be updated until the outer-most action has finished, guaranteeing that intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has completed.

and [`transaction`](https://mobx.js.org/api.html#transaction).

> Used to batch a bunch of updates without notifying any observers until the end of the transaction. 

and [old action docs](https://www.mobxjs.com/refguide/action.html)

> actions will batch mutations and only notify computed values and reactions after the (outer most) action has finished.

Now, this is assuming that the term `observer` is used as in the rest of the code base: `computed`s & `reaction`s. So not the `mobx-react` terminology, where it is just a `reaction`.

**Actual outcome:**

Computeds get recomputed during an action.

**How to reproduce the issue:**

https://jsfiddle.net/jasperh/w6bkthu9/8/

```js
const { observable, autorun, action } = mobx;

const obs = observable({
  first: 'John',
  last: 'Doe',
  get full() {
  	return this.first + ' ' + this.last;
  },
  update: action(function(e) {
    this.first = 'Pete';
    console.log('during action', this.full);
    this.last = 'Moe';
    console.log('during action', this.full);
  }),
})

autorun(() => console.log('reaction', obs.full));

obs.update();
```

Gives console:
```
reaction John Doe
during action Pete Doe
during action Pete Moe
reaction Pete Moe
```

**Versions**

MobX 5 & 6

**extra info**

Found this while researching a problem I described in [this discussion](https://github.com/mobxjs/mobx/discussions/3007).

I guess docs should be updated that only `reaction` and not `computed` is postponed. 

Though I'd still like an action that is fully transactional, including for computeds.

mweststrate:
computed is postponed until the end of the transaction, *unless* it is read
by something before the end of the transaction, to make sure no
inconsistent world is shown ever, and computed values can be read safely as
input for further decision logic. If you remove the `this.full` in your
console.logs, and instead put a console.log in full itself, you will see it
won't be called eagerly.

jheeffer:
Sure, but... computeds are never called eagerly, they're always lazy, regardless of transaction context?
It's just reactions that are postponed, which call the lazy computeds, right?
I.e. `(trans)action` don't directly influence `computed` behavior.

[...]

PR #3018, opened 2021-06-29
doc: Improve description of `action` and `transaction` (#3017)

jheeffer:
Closer description of actual implementation.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -18,7 +18,7 @@ All applications have actions. An action is any piece of code that modifies the 
 
 MobX requires that you declare your actions, although [`makeAutoObservable`](observable-state.md#makeautoobservable) can automate much of this job. Actions help you structure your code better and offer the following performance benefits:
 
-1. They are run inside [transactions](api.md#transaction). No observers will be updated until the outer-most action has finished, guaranteeing that intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has completed.
+1. They are run inside [transactions](api.md#transaction). No reactions will be run until the outer-most action has finished, guaranteeing that intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has completed. However, computeds will update when they have become stale and are read during an action.
 
 2. By default, it is not allowed to change the state outside of actions. This helps to clearly identify in your code base where the state updates happen.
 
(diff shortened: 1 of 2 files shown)


docs/actions.md
```diff
@@ -18,7 +18,7 @@ All applications have actions. An action is any piece of code that modifies the
 
 MobX requires that you declare your actions, although [`makeAutoObservable`](observable-state.md#makeautoobservable) can automate much of this job. Actions help you structure your code better and offer the following performance benefits:
 
-1. They are run inside [transactions](api.md#transaction). No observers will be updated until the outer-most action has finished, guaranteeing that intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has completed.
+1. They are run inside [transactions](api.md#transaction). No reactions will be run until the outer-most action has finished, guaranteeing that intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has completed. However, computeds will update when they have become stale and are read during an action.
```
urugator, line 21: > However, computeds will update when they have become stale and are read during an action.

I don't think it makes sense to mention `computed` here as some special case. It behaves the same as any observable value in regards to transactions. It always yields current value, (trans)action or not.  It is a computed **value** in the end. 

Perhaps if necessary, something like:
*"Observable values (including computeds) are unaffected by action and always yield current value."*
??

jheeffer: Fair enough, it was more to contrast `computed` from `reaction` as an observer, but that was due to the previous text mentioning observers. Since that is gone, indeed no need to have computed here at all.

urugator (approved):

docs/actions.md
urugator: Just to elaborate a bit and perhaps justify the original wording a little... 
From user's perspective the library conceptually consists of two parts: consumers (observers) and consumables (observables).
Computed is part of the consumable part (observable). While computed is internally "observer", it can't be used alone to consume observables.
Action is only relevant for the consuming part (observers) - reaction/autorun/observer.
Unfortunatelly neither the implementation or API maps to this natural division very well.
The most precise word to use here, would be `Reaction` (with capital R), which is internally used to implement all consumers (observers) - however since it's just an internal concept, nobody would understand that either.

jheeffer: Right, yeah, building a consistent & coherent DSL is hard, especially if it happens incrementally, I feel ya : )

PR #3028, opened 2021-07-05
chore: keep homepage identically

cncolder:
Keep all subpackage homepage is same as mobx: `https://mobx.js.org/`


iChenLei (approved):
LGTM

PR #3034, opened 2021-07-13
Update error message to include those cases where there is an existin…

HairyMaclary:
### Just a minor change to an error message string.

This error can appear when the function is decorated with @action, and so the message can be confusing. As an example, this decorated function was breaking:
```js
   @action
   async closeProject(projectId: string) {
      const project = this.projects.find((p) => p.id === projectId);
      if (project) {
         await project.close();
      }
      this.projects = this.projects.filter((p) => p.id !== projectId);
   }
```
The fix was to run anything following the await in a separate runInAction

```js
   @action
   async closeProject(projectId: string) {
      const project = this.projects.find((p) => p.id === projectId);
      if (project) {
         await project.close();
      }
      runInAction(() => {
          this.projects = this.projects.filter((p) => p.id !== projectId);
      });
   }
```

### Code change checklist

-   [ ] Added/updated unit tests - nope. npm run test does not work after npm install. However a global search reveals no other places use this minor string.
-   [ ] Verified that there is no significant performance drop (`npm run perf`) command fails but this is jsut a string change.

### DIFF
--- a/packages/mobx/src/core/derivation.ts
+++ b/packages/mobx/src/core/derivation.ts
@@ -141,7 +141,7 @@ export function checkIfStateModificationsAreAllowed(atom: IAtom) {
         console.warn(
             "[MobX] " +
                 (globalState.enforceActions
-                    ? "Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: "
+                    ? "Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Also action decorated async functions will not react to code following the 'await' keyword. Tried to modify: "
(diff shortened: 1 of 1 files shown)


iChenLei:
@HairyMaclary Can you update your PR decription to be more clear ? Current decription format is too bad.

HairyMaclary:
@iChenLei Hi, do you have an example somehwere? I'm a first time contributor to this project. However, if you want the tests run I'll just withdraw this. For some reason running npm install locally fails with this project and it's not worth the time for such a small change.

HairyMaclary:
No worries @mweststrate. Thanks for the input. I'll close this off and perhaps have a go at the docs instead.

PR #3040, opened 2021-07-23
Include reaction in imports

TruePath:
The code in the example uses the mobx reaction function yet the reaction function isn't included in the imports.  I've added it.


urugator:
Thank you!

PR #3056, opened 2021-08-09
Create WHEN_TIMEOUT error earlier to capture useful stack

upsuper:
Currently, the `WHEN_TIMEOUT` error is thrown from a `setTimeout`, which usually contains no stack itself, this makes it impossible to distinguish between different places throwing it, and hard to track where it is from, and thus make it a high-frequency error in our logging which is basically not actionable.

This PR creates the error earlier when the `when` starts, so that it preserves the stack on where the `when` is called.

Unconditionally creating an error might slightly regress the performance of `when`, but it only happens when `timeout` is specified. And `when` itself is probably not all that performance-critical anyway I suppose.

### Code change checklist

### DIFF
--- a/.changeset/many-moons-turn.md
+++ b/.changeset/many-moons-turn.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Create WHEN_TIMEOUT error earlier to capture useful stack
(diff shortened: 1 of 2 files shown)


mweststrate (approved):
Looking good! Thanks for the improvement

PR #3117, opened 2021-09-21
fix: Add the typescript type

garen-men:
找不到名称“console”，是否需要更改目标摩？请尝试更改 1ib编译名逢项以


iChenLei:
`mobx` is a pure js lib, we don't need add `DOM` in tsconfig compiler options, thanks.

PR #3123, opened 2021-09-27
fix(mobx): assert that array length in obs array is not NaN

rluvaton:
### Code change checklist


elektronik2k5 (approved):

urugator:
Just out of curiosity, have you encountered any error caused by the missing check?

rluvaton:
> Just out of curiosity, have you encountered any error caused by the missing check?

No, just read the code

PR #3153, opened 2021-10-20
Idea: jsx runtime that warns about missing observer

urugator:
With [the new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), we can basically provide own `createElement` implementation:
```javascript
// .babelrc
module.exports = {
  "presets": [
    [
      "@babel/preset-react",
      {        
         "importSource": "mobx-react"
      }
    ]
  ]
}
```
So the idea is to use it to make sure that observable props can be passed only to `observer`.

### DIFF
--- a/packages/mobx-react/src/index.ts
+++ b/packages/mobx-react/src/index.ts
@@ -1,5 +1,6 @@
 import { observable } from "mobx"
-import { Component } from "react"
+import { Component, createElement as _createElement } from "react"
+import { checkMissingObserver } from './jsx-runtime';
 
 if (!Component) throw new Error("mobx-react requires React to be available")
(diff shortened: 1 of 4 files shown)


kubk:
@urugator It'd be nice to add it to `mobx-react-lite` too. The documentation page "React integration" uses mostly `mobx-react-lite`: https://mobx.js.org/react-integration.html

That's why newcomers probably would use this package in favor of `mobx-react`

kubk:
@urugator Is it possible to warn about missing observer using some eslint rule? I don't know how difficult it is to implement but checking in compile time is more useful than checking in runtime.

urugator:
@kubk https://github.com/mobxjs/mobx/pull/3219

PR #3154, opened 2021-10-20
Throw when mixing decorators with annotations

### LINKED ISSUE #3152: False positive when using decorators: Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed.
**Intended outcome:**

No warnings about modifications outside of action.

**Actual outcome:**

`[MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: ObservableMap@9.keys()`

I made a codesandbox: https://codesandbox.io/s/blue-sound-h4y8s?file=/src/index.tsx
Press the buttons and see the console, without runInAction inside of the action, the observable map will generate a warning.

**Versions**

Using codesandbox.

**Scenario**
I had issues upgrading from MobX 5 to 6. Our project is quite complex and large, relying heavily on subclassing, decorators and other stuff. I'll share some code to see what it looks like. It's unclear to me if I'm doing something wrong.

```ts
import { action, computed, configure, makeObservable, observable, runInAction } from 'mobx';

// For this test it is crucial to enforce these mobx settings for testing.
configure({
  enforceActions: 'always',
});

interface TestItem {
  id: string;
  value: string;
}

class ObservableTest {
  /**
   * Observable value
   */
  @observable
  private _value = 'default';
  @action
  /**
   * Action to set value
   */
  public setValue(value: string) {
    this._value = value;
  }
  /**
   * Computed value getter
   */
  @action
  public get value(): string {
    return this._value;
  }
  /**
   * ObservableMap of items
   */
  private _items = observable.map<string, TestItem>();
  /**
   * Action to set to map
   * @param item TestItem to set
   */
  @action
  public setItem(item: TestItem) {
    runInAction(() => {
      this._items.set(item.id, item);
    });
  }
  /**
   * Action to remove from map
   * @param item TestItem to remove
   */
  @action
  public removeItem(item: TestItem) {
    runInAction(() => {
      this._items.delete(item.id);
    });
  }
  /**
   * Returns items in an array.
   */
  @computed
  public get items() {
    return Array.from(this._items.values());
  }
  /**
   * Default Constructor
   * Boostraps observables.
   */
  public constructor() {
    // Initialize observables.
    makeObservable(this, {});
  }
}

describe('MobX', () => {
  it('should value default', () => {
    const test = new ObservableTest();
    expect(test.value).toStrictEqual('default');
  });

  it('should set value', () => {
    const test = new ObservableTest();
    test.setValue('a'); 
    expect(test.value).toStrictEqual('a');
  });

  it('should item default', () => {
    const test = new ObservableTest();
    expect(test.items).toStrictEqual([]);
  });

  it('should set item', () => {
    const test = new ObservableTest();
    const item = { id: '1', value: 'a' };
    test.setItem(item); // triggers unexpected warning
    expect(test.items).toStrictEqual([item]);
  });

  it('should remove item', () => {
    const test = new ObservableTest();
    const item = { id: '1', value: 'a' };
    test.setItem(item); // triggers unexpected warning
    test.removeItem(item); // triggers unexpected warning
    expect(test.items).toStrictEqual([]);
  });
});

```

urugator:
fixes #3152

### DIFF
--- a/packages/mobx/src/api/makeObservable.ts
+++ b/packages/mobx/src/api/makeObservable.ts
@@ -12,9 +12,9 @@ import {
     die,
     ownKeys,
     extendObservable,
-    addHiddenProp
+    addHiddenProp,
+    storedAnnotationsSymbol
 } from "../internal"
-import { storedAnnotationsSymbol } from "./decorators"
 
 // Hack based on https://github.com/Microsoft/TypeScript/issues/14829#issuecomment-322267089
 // We need this, because otherwise, AdditionalKeys is going to be inferred to be any


mweststrate:
Neat!

PR #3173, opened 2021-11-01
docs update: subclassing missing override import

JustinHallquist:
Problem Statement: 

When reading the docs for `subclassing` they reference, both in text as well as code implementation, a method `override`. 
However, from the readers point of view it's never shown where this method comes from.

Expected Behavior:

Reading the code snippet should show the usage pattern for `override` in it's entirety, including importing it.

Actual Behavior:

The code snippet uses a non implemented method `override`.

Updated Behavior:

The code snippet now imports the `override` method which resolves the missing implementation detail.


urugator:
Thank you! Can't merge the PR, checks seem to be stuck. I will try to close and reopen the PR.

JustinHallquist:
Np, looks like that worked as well -- thanks for getting those pipelines moving!

PR #3174, opened 2021-11-05
add comment to help react-native users avoid react bug with flatlist

1mike12:
There's a pretty insidious bug (https://github.com/facebook/react-native/issues/21154) where after following the original instructions, the app works seemingly fine. But it's not until the cache of the webpack server is reset that you find out that everything breaks. The error is not that specfiic and trackign it down to this babel change takes a while. Hopefully this will help others out there with the same problem. 

The location / wording could use work but just wanted to get this process started

### DIFF
--- a/docs/enabling-decorators.md
+++ b/docs/enabling-decorators.md
@@ -79,6 +79,7 @@ Install support for decorators: `npm i --save-dev @babel/plugin-proposal-class-p
 {
     "plugins": [
         ["@babel/plugin-proposal-decorators", { "legacy": true }],
+        //["@babel/plugin-transform-flow-strip-types"], //for those using react, this plugin is required to avoid a certain error in flatlists. see https://github.com/facebook/react-native/issues/21154
         ["@babel/plugin-proposal-class-properties", { "loose": false }]
         // In contrast to MobX 4/5, "loose" must be false!    ^
     ]


urugator:
The 
https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types
is part of:
https://babeljs.io/docs/en/babel-preset-flow
isn't this the plugin one should be using? I mean, sort of automatically, if one uses `flow`...?

If I follow correctly the relevant option is:
https://babeljs.io/docs/en/babel-preset-flow#allowdeclarefields
but I am actually not sure, how it should be set: `true` seems like the way forward (it wil be default in Babel 8), but possibly breaking on existing code that did not migrated to using `declare` (?).

mweststrate:
It's a bit unclear why a bug in an (old version?) of RN should be part of the MobX documentation? I can imagine that would have been convenient in your case, but it is also a bit hard this way to avoid exploding these docs with random problems of other tools, as there is not really a way in which we can track when the comment isn't relevant anymore?

1mike12:
@mweststrate yeah I get that mobx is not tied to RN and at this point I am about 50:50 myself on this being included. But my counterarguemnt is that given that this particular page is very specific to enabling classes, that the bug is very subtle and is very hard to figure out, and that about 25% of public repos of mobx users are using it with RN, I think it's important enough to have some mention somewhere to help some other poor SOB. 

Even if it's not mentioned hopefully this thread can get indexed on google and other people can find it that way too

 [23.3k all mobx](https://github.com/search?q=%22mobx-react%22+filename%3Apackage.json&type=code), [5.5k RN w/ mobx](https://github.com/search?q=%22mobx-react%22+%22react-native%22+filename%3Apackage.json&type=code)

mweststrate:
Fair enough, will merge :). Thanks!

docs/enabling-decorators.md
```diff
@@ -79,6 +79,7 @@ Install support for decorators: `npm i --save-dev @babel/plugin-proposal-class-p
 {
     "plugins": [
         ["@babel/plugin-proposal-decorators", { "legacy": true }],
+        //["@babel/plugin-transform-flow-strip-types"], //for those using react, this plugin is required to avoid a certain error in flatlists. see https://github.com/facebook/react-native/issues/21154
```
kubk, line 82: Should it be "for those using react _native_" ?

urugator:
@1mike12
Do you exclude `node_modules` (or wherever `react-native` sources resides) from transpilation?
Based on other issues, excluding `node_modules` does **not** seems to be the stardard procedure when using native. I would like to understand if that's really the case and why.
I think it would be nice to provide some guidance to react-native users, but the proposed solutions around seems to be quite ad-hoc. I would like to understand if these issues originate from the same problem (like applying the plugin to `node_modules`) or if there is something more to it (like incorrect use of `declare` in user's sources or bug in specific react-native version).

PR #3180, opened 2021-11-10
Fix type inference of observe function (Close #3179)

### LINKED ISSUE #3179: IObjectDidChange `type: "remove"` type annotation mismatch with `type="delete"`
**Intended outcome:**

The `observer` listener's event object's `type` should be inferred union of "add" | "update" | "delete", and I should not get a type error when trying perform a comparison on `type` against the `"delete"` value.

Screenshot_20211110_142742

**Actual outcome:**

Getting a type error because the actual value being provided by MobX's runtime can be `"delete"` -- whereas it is typed as `"remove"`. [`mobx/src/types/observableobject.ts#L70`](https://github.com/mobxjs/mobx/blob/6519021a3e2071189111be33aa9e05ab4188d76a/packages/mobx/src/types/observableobject.ts#L70) needs to be changed to `"delete"`

Screenshot_20211110_142251

```typescript
observe(this.anythingObservable, (event) => {
  if (event.type === "delete") { ... }
})
```

To match the screenshot.

**Versions**

`mobx@6.3.6`

kubk:
### Code change checklist

Fixes: https://github.com/mobxjs/mobx/issues/3179

Both Map & Set resulted in wrong type inference. Now the correct type inference is covered with tests.

### DIFF
--- a/.changeset/moody-kids-punch.md
+++ b/.changeset/moody-kids-punch.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
+Fix type inference of observe function
--- a/.changeset/moody-kids-punch.md
+++ b/.changeset/moody-kids-punch.md
@@ -1,5 +1,5 @@
 ---
-"mobx": minor
+"mobx": patch
 ---
 
 Fix type inference of observe function
--- a/packages/mobx/__tests__/v5/base/typescript-tests.ts
+++ b/packages/mobx/__tests__/v5/base/typescript-tests.ts
@@ -2115,6 +2115,11 @@ test("TS - type inference of observe function", () => {
     observe(store.set, argument => {
         assert<IsExact<typeof argument, ISetDidChange<number>>>(true)
     })
+
+    // set created with observable.set
+    observe(store.setMobxFactory, argument => {
+        assert<IsExact<typeof argument, ISetDidChange<number>>>(true)
+    })
 })
 
 test("TS - type inference of reaction opts.equals", () => {
--- a/packages/mobx/__tests__/v5/base/typescript-tests.ts
+++ b/packages/mobx/__tests__/v5/base/typescript-tests.ts
@@ -25,10 +25,12 @@ import {
     createAtom,
     runInAction,
     flow,
+    IMapDidChange,
+    IValueDidChange,
+    ISetDidChange,
     flowResult
 } from "../../../src/mobx"
 import * as mobx from "../../../src/mobx"
-// @ts-ignore
 import { assert, IsExact } from "conditional-type-checks"
 
 const v = observable.box(3)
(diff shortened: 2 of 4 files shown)


urugator:
I assume  [`intercept`](https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/api/intercept.ts) has the same problem, could you take a look at it as well?

packages/mobx/src/api/observe.ts
```diff
@@ -25,17 +25,17 @@ export function observe<T>(
     fireImmediately?: boolean
 ): Lambda
 export function observe<V>(
-    observableMap: ObservableSet<V>,
+    observableMap: ObservableSet<V> | Set<V>,
```
urugator, line 28: Is the observable variant `ObservableSet/Map` even good for anything here? It implements `Set`...

mweststrate: that seems problematic indeed, as `observableMap` would effectively become the least common denominator of both. Why was this needed?

urugator: > Why was this needed?

In case you're asking about the need for `Set`/`Map`, there is an overload for object, which doesn't require `ObservableObject` but any `Object`:
```javascript
export function observe(
    object: Object,
    listener: (change: IObjectDidChange) => void,
    fireImmediately?: boolean
): Lambda
```
So Map/Set requires explicit cast to `ObservableMap/Set` otherwise it matches object overload. That's inconsistent and error prone.

mweststrate: Check!

kubk: @urugator Thanks, I've removed ObservableMap and ObservableSet from the typings. For some reason tests with ObservableMap pass but tests with ObservableSet don't:

image

This issue on CI: https://app.circleci.com/pipelines/github/mobxjs/mobx/1788/workflows/4a10412e-e197-4eea-bde8-3221d9b39c31/jobs/8295

I've double checked that both ObservableMap/ObservableSet implement ES6 versions. Both of them have same overload:

image

So the reason of issue is not clear. Workarounds:
- Return ObservableSet/ObservableMap back
- change unit test to match `ISetDidChange<any>`

mweststrate: There are TS setups where Set/Map is not defined (based on the selected standard libs), so that union might not be that bad after all

urugator: If that's the case I think a explanatory (or link) comment would be nice.

packages/mobx/__tests__/v5/base/typescript-tests.ts
```diff
@@ -204,32 +208,6 @@ test("scope", () => {
...
-        return null
-    })
-    mobx.observe(ar2, (d: IArrayDidChange<number>) => {
-        // console.log(d.type)
-    })
-
-    const x: IObservableValue<number> = observable.box(3)
-})
-
-const state: any = observable({
```
kubk, line 229: @urugator This variable isn't used anywhere. It also doesn't do anything useful because `observable` gets casted to `any`

packages/mobx/__tests__/v5/base/typescript-tests.ts
```diff
@@ -204,32 +208,6 @@ test("scope", () => {
     t.equal(x3.z, 8)
 })
 
-test("typing", () => {
-    const ar: IObservableArray<number> = observable([1, 2])
-    mobx.intercept(ar, (c: IArrayWillChange<number> | IArrayWillSplice<number>) => {
```
kubk, line 209: @urugator I've merged tests for intercept & observe into one test

packages/mobx/__tests__/v5/base/typescript-tests.ts
```diff
@@ -278,6 +256,8 @@ test("box", () => {
     t.deepEqual(ar.slice(), [40, 20, 60, 210, 420])
     box.addSize()
     expect(ar.slice()).toEqual([40, 20, 60, 210, 420, 700])
+
+    const x: IObservableValue<number> = observable.box(3)
```
kubk, line 260: This line was moved from the removed test above

kubk:
@urugator I've fixed `intercept` typings and added the link to this discussion.

.changeset/moody-kids-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"mobx": minor
```
urugator, line 2: Why not patch?

kubk: Agree, fixed

urugator:
Thank you!

PR #3182, opened 2021-11-10
Fix outdated PR template (Close #3181)

### LINKED ISSUE #3181: Pull Request template is outdated
Here is the current PR template:

image

I seems like there is no `npm run perf` script anymore. Also I had to use `yarn` instead of `npm` in order to set up the project locally. Should the third step of this template be updated to `yarn mobx test:performance` ?

kubk:
Fixes https://github.com/mobxjs/mobx/issues/3181


mweststrate:
Thanks!

Issue #3187, opened 2021-11-17
makeObservable create an action setter?


RoXuS:
Hey,

We want to migrate from redux to mobx and we don't understand why `makeObservable` on a class property create a setter and then this setter is define as @action.

The consequence is that we can set the property without use an action.

For example:

```ts
configure({
  enforceActions: 'always',
  userProxies: 'never'
})

export default class UserMobx {
  public lastName?: string = undefined;

  constructor() {
    makeObservable(this, {
      lastName: observable,
    });
  }

  public setLastName(lastName: UserMobx['lastName']) {
    this.lastName = lastName;
  }
}
```

In this code we can do `this.userMobx.lastName = ...` without an error even if we set `enforceActions` to `always` in configure.

So if it's the expected case, we don't understand the purpose of `enforceActions`?

urugator:
Can you setup a codesandbox repro?

RoXuS:
it works as expected on https://codesandbox.io/embed/frosty-violet-ebpxc?fontsize=14&hidenavigation=1&theme=dark.

Sorry for disturb, I don't understand what is going on with my setter, we have a simple Mobx class...

RoXuS:
For info, it works as expected we had `window.process = { env: { NODE_ENV: 'PRODUCTION' } };` so the warning was hidden.

With `window.process = { env: { NODE_ENV: 'DEVELOPMENT' } };`, it's ok.

PR #3188, opened 2021-11-18
doc: env info

urugator:
in reaction to:
https://github.com/mobxjs/mobx/issues/3187#issuecomment-971659488


mweststrate:
Looking great, thanks for elaborating!

PR #3189, opened 2021-11-20
Fix: RangeError in ObservableArray.replace for large arrays

RvanderLaan:
Hello! I noticed an error in my application when working with long lists, traced it back down to here.
Here's a minimal repro (included in the new test case) https://codesandbox.io/s/inspiring-matsumoto-kbjzc?file=/src/index.ts
It only seems to happen when both arrays are larger than `MAX_SPLICE_SIZE` and the original array is larger than its replacement

Cheers!

### Code change checklist

### DIFF
--- a/.changeset/ninety-rings-move.md
+++ b/.changeset/ninety-rings-move.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
+Fix for RangeError in ObservableArray.replace for large arrays
--- a/.changeset/ninety-rings-move.md
+++ b/.changeset/ninety-rings-move.md
@@ -1,5 +1,5 @@
 ---
-"mobx": minor
+"mobx": patch
 ---
 
 Fix for RangeError in ObservableArray.replace for large arrays
--- a/packages/mobx/src/types/observablearray.ts
+++ b/packages/mobx/src/types/observablearray.ts
@@ -236,9 +236,12 @@ export class ObservableArrayAdministration
         if (newItems.length < MAX_SPLICE_SIZE) {
             return this.values_.splice(index, deleteCount, ...newItems)
         } else {
+            // The items removed by the splice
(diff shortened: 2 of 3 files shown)


urugator:
Good call, thank you! Can you add a changeset?
Run `yarn changeset` in repo root dir  (spacebar to check, enter to confirm)

.changeset/ninety-rings-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"mobx": minor
```
urugator, line 2: `"mobx": patch` please

packages/mobx/src/types/observablearray.ts
```diff
@@ -237,7 +238,7 @@ export class ObservableArrayAdministration
         } else {
             const res = this.values_.slice(index, index + deleteCount)
             let oldItems = this.values_.slice(index + deleteCount)
-            this.values_.length = index + newItems.length - deleteCount
```
mweststrate, line 240: Could you elaborate a bit on the rationale of this change and add a unit test with a splice larger then MAX_SPLICE_SIZE that also deletes items? Maybe this assignment should happen after the for loop, but not correcting for deleteCount sounds incorrect to me? Maybe this original assignment is faulty as well though. Looking at it, I think it should read:

`this.values_.length = this.values_.length + newItems.length - deleteCount;`. That `index` there looks very off :)

mweststrate (changes requested):
See comments

packages/mobx/src/types/observablearray.ts
RvanderLaan: You are correct! I didn't completely understand what was happening, now I do after another read-through

mweststrate:
Looking great! Thanks for the tests and comments, and digging into this problem in the first place!

mweststrate (approved):

PR #3192, opened 2021-11-24
Support `displayName` on anonymous components

urugator:
partial fix for #2721 

### Background: 
For whatever reason, in react 17, `getComponentName` returns the name of the memoized component, rather than the name of the wrapper itself:
https://github.com/facebook/react/blob/12adaffef7105e2714f82651ea51936c563fe15c/packages/shared/getComponentName.js#L88
So how does `displayName` work on memoized component? Memoized component defines `displayName` as a getter/setter pair, where setter in addition to setting "own" name for the wrapper, [also modifies the name of the original component](https://github.com/facebook/react/blob/12adaffef7105e2714f82651ea51936c563fe15c/packages/react/src/ReactMemo.js#L41):
```javascript
const Original = function() {};
const Memoized = memo(Original);
Memoized.displayName = 'memoized';
console.log(Original.displayName === Memoized.displayName); // true
```
However it does it [only if the original component does not have a name yet](https://github.com/facebook/react/blob/12adaffef7105e2714f82651ea51936c563fe15c/packages/react/src/ReactMemo.js#L40):
```javascript
const Original = function() {};
Original.displayName = 'original'
const Memoized = memo(Original);
Memoized.displayName = 'memoized';
console.log(Original.displayName === Memoized.displayName); // false
```
In this case the `Memoized.displayName` is practically useless as it's ignored by devtools (you will see `original` in component tree).

### Problem:
Setting `displayName` (by us) on either original or memoized component prevents customizing `displayName` by user.

### Solution:
Since setting `displayName` is mostly relevant for anonymous functions, don't set `displayName` on these (we would set them to empty string anyway).
Additionally, don't hoist `displayName` as it would redefine memo's getter/setter.

### Consequences
Component returned from `observer` does not have it's own `displayName`, but the `wrapperComponent` (available as `.type`) does, which should be sufficient.

### DIFF
--- a/.changeset/warm-foxes-hang.md
+++ b/.changeset/warm-foxes-hang.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+Support customizing `displayName` on anonymous components [#2721](https://github.com/mobxjs/mobx/issues/2721).
--- a/packages/mobx-react-lite/__tests__/observer.test.tsx
+++ b/packages/mobx-react-lite/__tests__/observer.test.tsx
@@ -562,8 +562,7 @@ it("should hoist known statics only", () => {
     MyHipsterComponent.compare = "Nope!"
     MyHipsterComponent.render = "Nope!"
 
-    const wrapped = observer(MyHipsterComponent)
-    expect(wrapped.displayName).toBe("MyHipsterComponent")
+    const wrapped = observer(MyHipsterComponent)    
     expect(wrapped.randomStaticThing).toEqual(3)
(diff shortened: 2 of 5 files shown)


urugator:
Attempt to improve the situation a bit with #3129.
I did not test it, so I would appreciate if someone would give it a try.

urugator:
@LandonSchropp added `console.log("PR3192")` at the beginning of `observer`, in case you wanna give it another try.
Dunno if necessary, but don't forget to remove the original deps, before linking your locals versions.

PR #3196, opened 2021-11-26
Doc: Correct the instance name

vinzid:
In the example above, the instance name is `message` without `s`


kubk:
@vinzid Thank you.

PR #3210, opened 2021-12-02
Make all builtin comparers generic

Nokel81:
Without this change, if someone were to use the `comparer.structural` (or for that matter any other one) function. Then typescript (versions tested were 4.4 and 4.5) would assume `T = any` for reactions. Which is not ideal.

### Code change checklist

This change I believe is already covered by unit tests. Though I am a bit unsure if/where this change should be made part of docs.

### DIFF
--- a/.changeset/hungry-baboons-shout.md
+++ b/.changeset/hungry-baboons-shout.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix builtin comparers to be generic
--- a/packages/mobx/src/utils/comparer.ts
+++ b/packages/mobx/src/utils/comparer.ts
@@ -19,7 +19,9 @@ function shallowComparer<T>(a: T, b: T): boolean {
 function defaultComparer<T>(a: T, b: T): boolean {
     if (Object.is) return Object.is(a, b)
 
-    return a === b ? (a as any) !== 0 || 1 / (a as any) === 1 / (b as any) : a !== a && b !== b
+    return a === b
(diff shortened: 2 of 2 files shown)


Nokel81:
@kubk I am currently using mobx 6.3.7 and typescript 4.5.

However, I could not find a test that would work in the test files. I made this PR because I ran into this while working on https://github.com/lensapp/lens. Specially, the reaction (https://github.com/lensapp/lens/blob/master/src/main/cluster.ts#L300-L304) types `prefs` as `any`. But if I add a cast `comparer.structural as <T>(a: T, b: T) => boolean` then `prefs` is typed as `ClusterPrometheusPreferences` (specifically not `any`).

kubk:
@Nokel81 I've created a simple [demo](https://github.com/kubk/mobx-computed-struct-reproduction) and I am unable to reproduce the issue. The demo uses latest Mobx + TS 4.5. You can run the project with `npx tsc`. This command gives the following TS error: `"Property 'z' does not exist on type '{ x: number; y: number; }'."`. It means that the type of reaction argument is inferred correctly, it is not `any`

Nokel81:
@kubk Yes I see, I don't know why this happens for the lens repo. I guess we can close this PR then since I cannot make it fail outside of my repo.

kubk:
@Nokel81 Thank you. I am closing the PR because the issue isn't reproducible. If you find a Mobx bug - feel free to rise an issue or make a PR 👍

PR #3219, opened 2021-12-11
eslint: missing-observer, no-anonymous-observer

urugator:
See [README.md](https://github.com/mobxjs/mobx/pull/3219/files#diff-f7dad23cf5aef1d36e8bc064e49615120f537edb709fc91312b5a68440856f6c)

### DIFF
--- a/.changeset/curly-vans-enjoy.md
+++ b/.changeset/curly-vans-enjoy.md
@@ -2,4 +2,5 @@
 "eslint-plugin-mobx": patch
 ---
 
-Add [`mobx/missing-observer`](https://github.com/mobxjs/mobx/tree/main/packages/eslint-plugin-mobx#mobxmissing-observer) rule
+Add [`mobx/missing-observer`](https://github.com/mobxjs/mobx/tree/main/packages/eslint-plugin-mobx#mobxmissing-observer),
+[`mobx/no-anonymous-observer`](https://github.com/mobxjs/mobx/tree/main/packages/eslint-plugin-mobx#mobxno-anonymous-observer) rules,
--- a/.changeset/curly-vans-enjoy.md
+++ b/.changeset/curly-vans-enjoy.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-mobx": patch
+---
+
+Add [`mobx/missing-observer`](https://github.com/mobxjs/mobx/tree/main/packages/eslint-plugin-mobx#mobxmissing-observer) rule
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -68,9 +68,14 @@ configure({
 At some point you will discover that this level of strictness can be pretty annoying.
 It is fine to disable these rules to gain productivity once you are sure you (and your colleagues) grokked the mental model of MobX.
 
-Also, occassionally you will have a case where you have to supress the warnings triggered by these rules (for example by wrapping in `runInAction`).
+Also, occasionally you will have a case where you have to supress the warnings triggered by these rules (for example by wrapping in `runInAction`).
 That is fine, there are good exceptions to these recommendations.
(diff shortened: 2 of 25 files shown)


kubk:
@urugator Thank you for the implementation. I've gathered some feedback from people who use Mobx. The main question is can I wrap all react components with `observer` without getting performance overhead? 
I've only found this in the docs:

image

urugator:
We also suggest to [dereference the values as late as possible](https://mobx.js.org/react-optimizations.html#dereference-values-late), which practically means that all components *must be* `observer`. So `observer` should better be cheap otherwise it doesn't make much sense.
I would ask the other way around, how many components in your code base don't have to be `observer`? If it's a significant portion, then I would argue you don't use mobx idiomatically. If it's insignificant, then there is no issue, because the potential overhead is most likely insignificant as well. You will always have exceptions, but imo it's better to threat these, well, "exceptionally", rather than making everything else more complicated.
I would definitely start with just slap `observer` everywhere and if it wouldn't work (like fundamentally), then probably I wouldn't use mobx, because the selling point is that you don't have to think about and manage the subscriptions.

urugator:
Added `no-anonymous-observer` and changed both to `warn` by default. Basically I want to force people to try it out and provide feedback. Any opinions?

iChenLei:
Hi, Mr.urugator. I think you can add eslint-plugin-mobx to `mobx.js.org`, pin a `beta` flag. You should do more promotion (twitter, blog or whatever), no users,  no feedback. @urugator

kubk:
@urugator Once the plugin is released, I'll promote it among our Mobx group in Telegram (around 300 people). Since it's not merged yet, I can't try it on our projects.

iChenLei:
I developed a eslint plugin last two monthes, this is the online documentation for my plugin -> https://eslint-plugin-wxml.js.org

urugator:
Not compatible:
```
{ observableKind: "map"; debugObjectName: string; } & { object: ObservableMap<K, V>;             name: K;       type: "delete"; oldValue: V; }
{ observableKind: "map"; debugObjectName: string; } & { object: ObservableMap<unknown, unknown>; name: unknown; type: "delete"; oldValue: unknown; }
```
Any ideas?

kubk:
@urugator Have you accidentally updated TS version? Also as far I remember Mobx never used TS strict mode internally

urugator:
`main` has the same TS version in `package.json` (4) as I do locally, but I think the `yest.lock` on `main` is outdated, so it actually uses older version: https://github.com/mobxjs/mobx/blob/main/yarn.lock#L13419

urugator:
I am starting to loose my temper. Yesterady I went through all packages to solve/workaround compat with TS4. Run tests in each, all good, push to server, nope error on CI. Today I've rebuild `eslint-plugin-mobx`, run tests, same error, that's good, because at least it's replicable. Installed `jest:27` in `eslint-mobx-plugin`, error is gone and guess what, it breaks test in all other packages...

mweststrate:
@urugator let me know if you want me to take a look into it. I haven't followed the discussions, but if you can tell me what steps fail I can check it out

urugator:
Seems like there is a compatibility issue between `eslint:8` and `jest:25/26` (dunno which, because we have `26` in deps, but `tsdx` uses `25`). For the moment I just downgraded `eslint` to `7` and I will call it a day.
However I would love to understand why installing deps in one package affects other packages. I feel like I am missing something about this monorepo setup.
I would also like to know if there is a difference when I run a command like `yarn test` from individual package and when I run it from `root` - does it use the same version of `jest` for all packages or can a package use it's own `jest` (or typescritp etc) version?
Also we should get rid of TSDX as it's not maintained, we are stuck with the deps and imo it just obfuscates the whole setup.

PR #3233, opened 2021-12-21
fix: `isFunction`shouldn't be used

anderlaw:
### Code change checklist

### DIFF
--- a/.changeset/late-elephants-remain.md
+++ b/.changeset/late-elephants-remain.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+`eq.ts` fix comparer logic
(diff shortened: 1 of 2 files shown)


anderlaw:
I happened to see this logic:
```js
// Exhaust primitive checks
const type = typeof a
if (!isFunction(type) && type !== "object" && typeof b != "object") return false
```
But `isFunction`function uses `type of` inside, there is no way to pass the string of `type of` result into that function.

The author said it was copied from `underscore`, but copy wrong, here is `underscore` sourcecode:
```js
// Exhaust primitive checks
var type = typeof a;
if (type !== 'function' && type !== 'object' && typeof b != 'object') return false;
return deepEq(a, b, aStack, bStack);
```

urugator:
Good catch, can you please `yarn changeset`?

anderlaw:
> Good catch, can you please `yarn changeset`?

of course

urugator:
CI got stuck again, I will close and reopen the PR to force re-run

Issue #2912, opened 2021-04-23
toJS() doens't work well with partial observables (mobx-6)


ixrock:
**Example:**

```ts
class X {
  @observable num = 123;
  @observable list: any[] = [];
  @observable map = new Map<any, any>();
  @observable object: Record<string, any> = {};

  constructor() {
    makeObservable(this);
  }

 // works as expected, but could be easily broken with toJS({...this})
  toJSON() {
    return toJS(this);
  }

  // broken: still contains Proxy()-objects / non-serializible in some envs (e.g. electron.js ipc messaging)
  getState() {
    return toJS({
      num: this.num,
      list: this.list,
      map: this.map,
      object: this.object,
    });
  }
}

const test = new X();
console.log("toJSON():", test.toJSON())
console.log("getState():", test.getState())
```

**Result:**

![image](https://user-images.githubusercontent.com/6377066/115855504-fbd10900-a433-11eb-8ae8-b9bf0cce7b7a.png)

**Versions**

```
"mobx": "^6.2.0",
"webpack": "^4.44.2",
"typescript": "4.0.2",
```

urugator:
This is expected, `toJS` does not recurse into non-observables. You would have to either 
`return toJS(observable({ list: this.list }))` 
or 
`return { list: toJS(this.list) }`

ixrock:
This is not expected regarding to docs:

> To convert a data tree **recursively** to plain objects, the toJS utility can be used. For classes, it is recommend to implement a toJSON() method, as it will be picked up by JSON.stringify.

https://mobx.js.org/observable-state.html#converting-observables-back-to-vanilla-javascript-collections

urugator:
Feel free to improve it.

ixrock:
> This is expected, `toJS` does not recurse into non-observables. You would have to either
> `return toJS(observable({ list: this.list }))` (not very efficient)
> or
> `return { list: toJS(this.list) }`

Would be `JSON.parse(JSON.stringify({list: this.list}))` more performant btw?

urugator:
Dunno, to minimize overhead of creating observable, you may try:
`toJS(extendObservable({ list: this.list }, {}))`

I think it shouldn't be too difficult to write own `toJS` that suits your needs.

@mweststrate any chance of reviving `recurseEverything` option?

mweststrate:
[...]

PR #3239, opened 2021-12-31
Warn in `toJS` docs about deep cloning partial observables

bernardobelchior:
Warn in `toJS` docs about deep cloning partial observables, to avoid issues with some properties being deep clones while others are not. 

I have faced this issue while upgrading to MobX 6 and I believe it will save other people some headaches if it is clear that `toJS` has this limitation.

Relates to https://github.com/mobxjs/mobx/issues/2912

### DIFF
--- a/packages/mobx/src/api/tojs.ts
+++ b/packages/mobx/src/api/tojs.ts
@@ -62,11 +62,10 @@ function toJSHelper(source, __alreadySeen: Map<any, any>) {
 }
 
 /**
- * Basically, a deep clone, so that no reactive property will exist anymore.
- * Note that this function will only deep clone observables. If it finds a 
- * non-observable (e.g., plain object), it will not clone that. 
- * If you use classes that mixes observables with non-observables, you will
- * need to clone the non-observables by hand.
+ * Recursively converts an observable to it's non-observable native counterpart.
+ * It does NOT recurse into non-observables, these are left as they are, even if they contain observables.
+ * Computed and other non-enumerable properties are completely ignored.
+ * Complex scenarios require custom solution, eg implementing `toJSON` or using `serializr` lib.
  */
 export function toJS<T>(source: T, options?: any): T {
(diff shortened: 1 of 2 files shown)


packages/mobx/src/api/tojs.ts
```diff
@@ -63,6 +63,10 @@ function toJSHelper(source, __alreadySeen: Map<any, any>) {
 
 /**
```
urugator, line 64: Thank you for the improvement, could you please replace whole comment with:
```
/**
 * Recursively converts an observable to it's non-observable native counterpart.
 * It does NOT recurse into non-observables, these are left as they are, even if they contain observables.
 * Computed and other non-enumerable properties are completely ignored.
 * Complex scenarios require custom solution, eg implementing `toJSON` or using `serializr` lib.
 */
```
"deep copy but not really" bit was just confusing
Observables and non-observables can be mixed even without classes, no need to be specific about these.

bernardobelchior: Agreed. It makes it much clearer, thanks for the suggestion 👍

kubk:
@bernardobelchior Thank you for the clarification. Would you like to add this information to Mobx docs too? https://mobx.js.org/api.html#tojs

urugator (approved):

bernardobelchior:
@kubk I wouldn't mind. Although we would have to merge the docs and this comment, since they overlap a bit.
If you wouldn't mind, I'd send a PR shortly after this one with that

kubk:
@bernardobelchior OK, feel free to create a separate PR to Mobx docs. Thank you.

@urugator Is there a way to restart pipelines?

urugator:
@kubk Yes by closing and reopening the PR :D

bernardobelchior:
@kubk As promised, here it is: https://github.com/mobxjs/mobx/pull/3249

PR #3263, opened 2022-01-11
Documentation: Add installation.md note about Next.js

phoenixeliot:
This adds a note for Next.js users, who (unintuitively) must configure babel (which is normally not needed at all) even if their project uses typescript. I think this is worth including because Next.js is relatively popular, and following the instructions as stated here simply doesn't work for unclear reasons. I spent the better part of an evening debugging this, only to eventually figure out that Next.js uses babel for compilation even if the project is set up with/written in typescript and doesn't have a babel config file yet.

Had I not had an evening free that I felt like spending on this, I probably would have given up entirely on MobX, as this was my first project to try it out and it seemed to just not work when used as directed (as all typescript-declared class fields without initial values were just mysteriously not tracked by makeAutoObservable)

See https://github.com/mobxjs/mobx/discussions/2831#discussioncomment-1937782 for related discussion / my journey in debugging this

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -38,6 +38,8 @@ For verification insert this piece of code at the beginning of your sources (eg.
 if (!new class { x }().hasOwnProperty('x')) throw new Error('Transpiler is not configured correctly');
 ```
 
+Note that for Next.js you must [customize Babel](https://nextjs.org/docs/advanced-features/customizing-babel-config) instead of TypeScript, even if your project is set up to use TypeScript. 
+
 ## MobX on older JavaScript environments
 
 By default, MobX uses proxies for optimal performance and compatibility. However, on older JavaScript engines `Proxy` is not available (check out [Proxy support](https://kangax.github.io/compat-table/es6/#test-Proxy)). Examples of such are Internet Explorer (before Edge), Node.js < 6, iOS < 10, Android before RN 0.59, or Android on iOS.


urugator:
The most staggering thing is that nowadays you shouldn't need to set this up at all, because these are defaults. But it depends on framework, it's version, it's compilation setup, babel version and presets version ... it's a mess.

Btw here: https://babeljs.io/docs/en/babel-plugin-transform-typescript#typescript-compiler-options they state:
*`--useDefineForClassFields` You can use the `onlyRemoveTypeImports` option to replicate this behavior.*
Dunno what other implications this has or if/how you can modify babel typescript plugin options in nextjs.

jamieathans:
> The most staggering thing is that nowadays you shouldn't need to set this up at all, because these are defaults. But it depends on framework, it's version, it's compilation setup, babel version and presets version ... it's a mess.
> 
> Btw here: https://babeljs.io/docs/en/babel-plugin-transform-typescript#typescript-compiler-options they state: _`--useDefineForClassFields` You can use the `onlyRemoveTypeImports` option to replicate this behavior._ Dunno what other implications this has or if/how you can modify babel typescript plugin options in nextjs.
> 
> EDIT based on this: https://github.com/vercel/next.js/blob/canary/packages/next/build/babel/preset.ts I think it should be doable like:
> 
> ```json
> "presets": [
>   ["next/babel", { "preset-typescript": { "onlyRemoveTypeImports": true } }]
> ],
> ```

Just verifying that the above suggested `.babelrc` does indeed fix the issue for Next.js.

Maybe it should be added to the updated documentation comment?

kubk:
@jamieathans Does the issue exist in a fresh Next.js setup?

jamieathans:
> @jamieathans Does the issue exist in a fresh up-to-date Next.js setup?

Just double checked this again using the following `npm` versions (fresh Next.js setup):

```
{
  "name": "nextjs-mobx",
  "version": "0.1.0",
  "private": true,
  "scripts": {
    "dev": "next dev",
    "build": "next build",
    "start": "next start",
    "lint": "next lint"
  },
  "dependencies": {
    "mobx": "6.6.0",
    "mobx-react": "7.5.0",
    "next": "12.1.6",
    "react": "18.2.0",
    "react-dom": "18.2.0"
  },
  "devDependencies": {
    "@types/node": "18.0.0",
    "@types/react": "18.0.14",
    "@types/react-dom": "18.0.5",
    "eslint": "8.18.0",
    "eslint-config-next": "12.1.6",
    "typescript": "4.7.4"
  }
}
```

What I have discovered is just the presence of a minimal `.babelrc` as described [here](https://nextjs.org/docs/advanced-features/customizing-babel-config) will fix the mobx issue:

```
{
  "presets": ["next/babel"],
  "plugins": []
}
```

Maybe this implies the issue is with [SWC](https://swc.rs/) as I also see the following messages in the node dev terminal:

```
info  - Disabled SWC as replacement for Babel because of custom Babel configuration ".babelrc" https://nextjs.org/docs/messages/swc-disabled
.
.
.
info  - Using external babel configuration from /Users/jamie/projects/nextjs-mobx/.babelrc
```

mweststrate:
Ugh this PR complete fell off, let me just merge it and if it turns out to be incorrect anyone feel free to remove it again :)

PR #3284, opened 2022-02-09
eslint missing-make-observable-rule support for namespaces and fix imports

WearyMonkey:
Improves the `missing-make-observable-rule` to:

* Support namespace imports e.g. `@mobx.observable.ref`
* Insert imports if required

For example:

```
import * as mobx from 'mobx';

class C {
   @mobx.observable.ref c = 1;
}
```

Will be fixed to:
```
import * as mobx from 'mobx';

class C {
  constructor() {
    mobx.makeObservable(this)
  }

   @mobx.observable.ref c = 1;
}
```

And

```
import { observable } from 'mobx';

class C {
   @observable.ref c = 1;
}
```

Will be fixed to:
```
import { observable, makeObservable  } from 'mobx';

class C {
  constructor() {
    makeObservable(this)
  }

   @observable.ref c = 1;
}
```

### Code change checklist


packages/eslint-plugin-mobx/src/missing-make-observable.js
```diff
@@ -28,31 +53,67 @@ function create(context) {
         }
       } else {
         const fix = fixer => {
+          // The class extends a another unknown class so we can not safely create a super call.
```
urugator, line 56: Doesn't matter, create `constructor` without `super` and let another eslint rule handle missing super.
I think that handling React component specifically as part of this rule is unnecessary as well, but I am ok with it.

PR #3287, opened 2022-02-10
Remove any from ObservableMap public API (Close #3286)

### LINKED ISSUE #3286: Removing `any` types from `replace` and `merge` methods
### Discussed in https://github.com/mobxjs/mobx/discussions/3285

<div type='discussions-op-text'>

<sup>Originally posted by **Nokel81** February  9, 2022</sup>
The current typings for `ObservableMap<K, V>` include the following method:

```
replace(values: ObservableMap<K, V> | IKeyValueMap<V> | any): ObservableMap<K, V>;
```

I would love to see that `| any` removed, what could I do to help with that? Including for `ObservableSet` and `ObservableArray`.

I know that, for instance, also `[K, V][]` is an acceptable type for `values` since I use it a lot.</div>

kubk:
Related issue: #3286

Not needed
These changes are TS-only, they don't affect runtime performance 

### Summary
- Removed any from ObservableMap public API
- Added explicit `any` to assertion functions
- Removed obsolete `ts-ignore`
- All the codestyle changes were made by the pre-commit hook

### DIFF
--- a/packages/mobx/src/types/observablemap.ts
+++ b/packages/mobx/src/types/observablemap.ts
@@ -313,7 +313,7 @@ export class ObservableMap<K = any, V = any>
         transaction(() => {
             if (isPlainObject(other))
                 getPlainObjectKeys(other).forEach((key: any) =>
-                    this.set(key as any as K, (other as any)[key])
+                    this.set(key as K, (other as IKeyValueMap)[key])
                 )
             else if (Array.isArray(other)) other.forEach(([key, value]) => this.set(key, value))
             else if (isES6Map(other)) {
--- a/packages/mobx/src/types/observablemap.ts
+++ b/packages/mobx/src/types/observablemap.ts
@@ -306,14 +306,14 @@ export class ObservableMap<K = any, V = any>
     }
 
     /** Merge another object into this object, returns this. */
-    merge(other: ObservableMap<K, V> | IKeyValueMap<V> | any): ObservableMap<K, V> {
+    merge(other?: IObservableMapInitialValues<K, V>): ObservableMap<K, V> {
         if (isObservableMap(other)) {
             other = new Map(other)
         }
         transaction(() => {
             if (isPlainObject(other))
                 getPlainObjectKeys(other).forEach((key: any) =>
-                    this.set(key as any as K, other[key])
+                    this.set(key as any as K, (other as any)[key])
                 )
             else if (Array.isArray(other)) other.forEach(([key, value]) => this.set(key, value))
             else if (isES6Map(other)) {
@@ -332,7 +332,7 @@ export class ObservableMap<K = any, V = any>
         })
     }
 
-    replace(values: ObservableMap<K, V> | IKeyValueMap<V> | any): ObservableMap<K, V> {
+    replace(values: IObservableMapInitialValues<K, V>): ObservableMap<K, V> {
         // Implementation requirements:
         // - respect ordering of replacement map
         // - allow interceptors to run and potentially prevent individual operations
--- a/.changeset/smooth-penguins-pretend.md
+++ b/.changeset/smooth-penguins-pretend.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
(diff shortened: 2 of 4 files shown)


packages/mobx/src/types/observablemap.ts
```diff
@@ -306,14 +306,14 @@ export class ObservableMap<K = any, V = any>
     }
 
     /** Merge another object into this object, returns this. */
-    merge(other: ObservableMap<K, V> | IKeyValueMap<V> | any): ObservableMap<K, V> {
+    merge(other?: IObservableMapInitialValues<K, V>): ObservableMap<K, V> {
```
kubk, line 309: The `undefined` type was added because method `merge` may accept `undefined` according to ObservabelMap's constructor: https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/types/observablemap.ts#L101

Merging Map with undefined doesn't make much sense, but adding `if` type guard would be a BC break. That's why I decided not to affect runtime

packages/mobx/src/types/observablemap.ts
```diff
@@ -306,14 +306,14 @@ export class ObservableMap<K = any, V = any>
...
         if (isObservableMap(other)) {
             other = new Map(other)
         }
         transaction(() => {
             if (isPlainObject(other))
                 getPlainObjectKeys(other).forEach((key: any) =>
-                    this.set(key as any as K, other[key])
+                    this.set(key as any as K, (other as any)[key])
```
kubk, line 316: This new `as any` was introduced because of the `undefined` above.

urugator: `other as IKeyValueMap`

I wonder if it would be correct for `isPlainObject` to also return something like `value is Object`. The compiler then should know `other[key]` is fine, right?

kubk: I've tried with no success. Consider this example:

```js
const a: Object = [];
const b: {[k:string]: any} = [1,2,3]; 
const c: Record<string, any> = [1,2,3]; 
const d: object = [1,2,3]; 
```

None of these lines give a compilation error: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhgLhgeQEYCsCmxYF4YDaAugNwBQoksqSA3gQNYLQBOAlmAOZFJxgCeAXxj4CARgA0AJgkBmUjArhoMYEgBK2ECwAmAHlYdOE+AIB8IwpJnySiyip1IQGbHivS5CskA

It's a bit surprising but it works like this due to JS nature. Running `typeof []` in JS gives `object`. And TS `object` "represents any non-primitive type": https://mariusschulz.com/blog/the-object-type-in-typescript

Why I mentioned this? Because if I make `isPlainObject` return something like `value is object` other TS issues appear:

Screen Shot 2022-02-10 at 15 52 09

The type of `other` became `never` because TS narrowed its type to a primitive type. And primitive type can't be an array.

To summarize. We can either leave it as is, or add `if (other === undefined) return` to avoid any, or change `as any` to `as IKeyValueMap`

urugator: Ah, I absolutely did not realize it also works for the `else` branch, assuming the opposite of `Object` :)
I was not familiar with this `thing as Type` construct, I am asking largely for my own educational pruposes, so thank you for the elaborate answer.

> merging with undefined doesn't make much sense anyway

Not sure if it's comparable, but I find `Object.assign` allowing for `undefined`s quite handy sometimes.

> as any to as IKeyValueMap

Let's go with that, even the `any` is fine with me. `IKeyValueMap` just seemed more correct since that's what this `isPlainObject` is about...

urugator: I do prefer braces here (perhaps in general), but I am unopinionated about eslint rule. However, if you decide to, it will probably involve more changes, so you may want to do it as part of seperate PR, to keep diff focused.

kubk:
@Nokel81 ObservableSet's `replace` method doesn't have any in signature: https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/types/observableset.ts#L225

Which methods you'd like to improve?

PR #3289, opened 2022-02-11
Postfix `dist/mobx.esm*` with .mjs; Add package exports

shrinktofit:
### Code change checklist

### DIFF
--- a/packages/mobx/package.json
+++ b/packages/mobx/package.json
@@ -7,9 +7,23 @@
     "umd:main": "dist/mobx.umd.production.min.js",
     "unpkg": "dist/mobx.umd.production.min.js",
     "jsdelivr": "dist/mobx.umd.production.min.js",
-    "jsnext:main": "dist/mobx.esm.production.min.js",
-    "module": "dist/mobx.esm.js",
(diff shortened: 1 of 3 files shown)


shrinktofit:
This PR makes ESM formats of `mobx` usable in Node native ESM system:

```js
import { makeAutoObservable } from 'mobx';
console.log(makeAutoObservable);
```

urugator:
Why is the `mjs` extension necessary?

I think it's a BC:
1. paths are different
2. on ESM supported envs, it will by dafult load non-minified version without resolved `process.env.NODE_ENV` checks, which has performance implications.

shrinktofit:
> Why is the `mjs` extension necessary?
> 
> I think it's a BC:
> 
> 1. paths are different
> 2. on ESM supported envs, it will by default load non-minified version without resolved `process.env.NODE_ENV` checks, which has performance implications.

See here: https://nodejs.org/api/esm.html#enabling

urugator:
I did see that, but I would have assumed that the `import/require` mapping in `package.json` is sufficient heuristic for determining which loader to use for which file. Turns out it isn't, the `cjs/mjs` extension is stil necessary :(

PR #3291, opened 2022-02-11
Clear up documentation

ninjapiratica:
We ran into a problem while installing mobx. We copied and used the original snippet and it broke our app (couldn't log or get hot-reload on react native). We got a warning about the 'loose': false but following this documentation didn't get us to where we needed to be.

I believe this is a clearer and more complete explanation of what is actually needed for configuration.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -26,8 +26,11 @@ There are two types of React bindings, `mobx-react-lite` supports only functiona
 -   **Babel**: Make sure to use at least version 7.12, with the following configuration:
     ```json
     {
+        // Babel < 7.13.0
         "plugins": [["@babel/plugin-proposal-class-properties", { "loose": false }]],
+        
         // Babel >= 7.13.0 (https://babeljs.io/docs/en/assumptions)
+        "plugins": [["@babel/plugin-proposal-class-properties"]],
         "assumptions": {
             "setPublicClassFields": false
         }


urugator:
Well, you were supposed to just copy-paste that (whole) config. I don't think documentation suggests otherwise, but whatever. The comment is only there, because it would throw `Unknown option` with older `babel`.

urugator:
Thank you for the PR.

ninjapiratica:
We did copy and paste the whole config. It fails on our app. It no longer logs to the react-native Metro pages and the hot-reload fails. We had to remove the { "loose": false } entirely for it to actually work. It might work fine in a browser, but the android emulator and Metro fails. I just wanted to clear up the documentation so others don't have that problem.

urugator:
I see. Could you double check it's really caused by that extra option? It sounds more like a coincidence - the output produced by babel with or without the option should be exactly the same.

ninjapiratica:
We've been hitting our heads on this for 2 days. The only thing that got it to work was removing the { 'loose' : false } and adding in the extra assumption.  We have no idea why that would cause it stop the connection with Metro, it makes no sense to us either. As soon as we switch it over, it started working. When we switched back, it stopped again. We had several apps and every one we tried, as soon as we made that one change, they started/stopped working. No other changes were made. 

It's one of the strangest things we've ever run into. We were testing all the way down to socket connections and making sure everything was/wasn't communicating properly. Huge headache that was resolved with the tiniest of changes.

PR #3294, opened 2022-02-12
Apply 'curly' ESLint rule + run Prettier on changed files

kubk:
Related discussion: https://github.com/mobxjs/mobx/pull/3287#discussion_r803723827

This PR applies ESLint rule "curly" on the whole code base: https://eslint.org/docs/rules/curly
What ESLint documentation says about this rule: "it is considered by many to be best practice to never omit curly braces around blocks, even when they are optional, because it can lead to bugs and reduces code clarity" 

The commit was generated automatically using `yarn lint --fix && yarn prettier`. Prettier changes were applied only for changed files. Just in case I checked that this change doesn't affect bundle at all. Here are the results of running command `yarn lerna run test:size` before and after applying changes:

Screen Shot 2022-02-12 at 22 48 57

It's a screenshot from https://www.diffchecker.com/diff

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,6 +15,7 @@ module.exports = {
     rules: {
         "no-fallthrough": "off",
         "no-constant-condition": "off",
+        curly: "error",
         "getter-return": "off",
         "no-console": "off",
         "no-var": "error",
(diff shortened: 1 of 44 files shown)


.eslintrc.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     rules: {
         "no-fallthrough": "off",
         "no-constant-condition": "off",
+        curly: "error",
```
kubk, line 18: Prettier decided to remove quotes here

urugator (approved):
lgtm

PR #3332, opened 2022-03-10
Little grammar fix in doc

morfey13:
### Code change checklist


kubk:
I see no grammar issue here. Functions `onBecomeObserved` and `onBecomeUnobserved` are hooks into observability system.

3imed-jaberi (approved):

urugator:
Perhaps *"They hook into"* would be a good compromise?
Also 
*when the observables _start_ and _stop_ becoming observed*
=>
*when an observable _starts_ and _stops_ being observed*

kubk:
@urugator Thanks, I've updated the wording

PR #3346, opened 2022-03-28
Fix word in docs

mr-nati:
Change 'a temporarily action' to 'a temporary action'


urugator:
What do you think about *"one-off action"*?

mr-nati:
Sorry I didn't get back to your suggestion in time! I agree that "one-off" is probably better than "temporary"

PR #3352, opened 2022-04-02
Mock console.warn in tests and verify warnings are called in the right place

kubk:
Mock `console.warn` in tests and verify warnings are called in the right place. I used Jest snapshots if warning messages are stable. ~~Here is an example of an "unstable" warning message:~~

~~`"[MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: ObservableObject@1.multiplier",`. The ObservableObject@**1**.multiplier part is unstable here and may change between test invocations.~~

UPD: All the snapshots are stable now.

This PR aims to get rid of Mobx warnings in tests. Getting rid of React's errors is a separate task.

How to verify changes? Go to the `ci/circleci: test` CI job and notice there are no console warnings anymore. Compare it to a previous PR, for example: https://app.circleci.com/pipelines/github/mobxjs/mobx/1911/workflows/f6bb5b20-37b4-4c66-aa18-384430efbc59/jobs/8614

### Code change checklist

-  Updated `/docs`. For new functionality, at least `API.md` should be updated
-  Verified that there is no significant performance drop (`yarn mobx test:performance`)


packages/mobx/__tests__/v4/base/__snapshots__/array.js.snap
```diff
@@ -0,0 +1,17 @@
...
+
+exports[`very long arrays can be safely passed to nativeArray.concat #2379 1`] = `
+[MockFunction] {
+  "calls": Array [
+    Array [
+      "[mobx] Out of bounds read: 10000",
+    ],
+  ],
+  "results": Array [
+    Object {
```
kubk, line 11: The snapshot include additional information, but it looks like it's the only way to use snapshots on mocks: https://github.com/facebook/jest/issues/2814#issuecomment-347095063

packages/mobx-react-lite/__tests__/observer.test.tsx
```diff
@@ -912,7 +915,6 @@ it("dependencies should not become temporarily unobserved", async () => {
 
     // @ts-ignore
     React.useEffect.mockImplementation(effect => {
-        console.warn("delaying useEffect call")
```
kubk, line 915: It looks like it was left here for debugging purposes

urugator:
> Here is an example of an "unstable" warning message:

Did you try/consider making messages stable by providing `name` option to relevant observables/reactions?

packages/mobx-react-lite/__tests__/observer.test.tsx
```diff
@@ -270,7 +270,10 @@ function runTestSuite(mode: "observer" | "useObserver") {
 
     describe("issue 309", () => {
         test("isObserverBatched is still defined and yields true by default", () => {
+            const warn = jest.spyOn(console, "warn").mockImplementation(() => {})
             expect(isObserverBatched()).toBe(true)
+            expect(warn).toHaveBeenCalled()
+            warn.mockReset()
```
urugator, line 276: Shouldn't it be `mockRestore`?
Also what happens if the test throws before `mockRestore`? Does each test run in complete isolation or will it after other tests and therefore we should restore console in  `try {} finally {}`/`afterEach`?

kubk:
@urugator Regarding making console warnings stable. How can I specify the `name` of `multiplier` property here?

```js
const store = useLocalObservable(() => ({
    multiplier,
    count: 10,
    get multiplied() {
        return this.multiplier * this.count
    },
    inc() {
        this.count += 1
    }
}))
```

Here is the warning message: `"[MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: ObservableObject@15.multiplier",`

Is it possible here?

packages/mobx-react-lite/__tests__/observer.test.tsx
kubk: It makes sense. I suggest to use something like `afterEach`

urugator:
> Is it possible here?

I don't think so, maybe hacking the name directly through _getAdministration etc, but it's not worth it.
I didn't check the details, but in case the test does not test `useLocalObservable` specifically and uses it just as a helper to create observable, then feel free to change it to something that accepts the name like `useState(observable({}, { name }))`
If it's one of the deprecated tests, then just don't bother with it at all.

urugator:
Or if you want you can match just the beginning of the warning with regex.

Btw these legacy `strict-mode` warnings needs to be replaced via something that reflects current config options like `Since enforceAction: "always"  is set, ....`. Just mentioning it if you would get out of ideas for contributions 😁

Also I dunno if you've noticed but I've added you to mobx team gitter. If you would need something directly from Michel, it's more likely you will catch him up there.

kubk:
@urugator Now all the `console.warn` invocations are tested using snapshots. Thank you for the suggestion. This change brought a lot of improvements. Here is an example: 

https://github.com/mobxjs/mobx/blob/65fcf1d010a7c060222d2beb48febfeae4c75f85/packages/mobx-react/__tests__/__snapshots__/observer.test.tsx.snap#L3-L10

As you can see from this snapshot, now the test is actually verifying that there is a warning. There are more of such snapshots in the PR. Also thank you for adding me to gitter. Feel free to suggest other improvements if you have 👍

packages/mobx-react-lite/__tests__/observer.test.tsx
```diff
@@ -18,6 +23,7 @@ function runTestSuite(mode: "observer" | "useObserver") {
             return observer(component)
         } else {
             const c = (props: P) => {
+                consoleWarnMock = jest.spyOn(console, "warn").mockImplementation(() => {})
                 return useObserver(() => {
```
kubk, line 27: @urugator I had to just silence console warnings here because the function runTestSuite is called twice with different arguments and the snapshots didn't work for me.

https://github.com/mobxjs/mobx/blob/main/packages/mobx-react-lite/__tests__/observer.test.tsx#L481-L482

Since it is deprecated API I think it's better to leave it as is and remove `useObserver` in next major verions

packages/mobx-react-lite/__tests__/useLocalStore.test.tsx
```diff
@@ -475,25 +478,39 @@ describe("enforcing actions", () => {
...
-                count: 10,
-                get multiplied() {
-                    return this.multiplier * this.count
-                },
-                inc() {
-                    this.count += 1
-                }
-            }))
+
+            const [store] = useState(() =>
```
urugator, line 489: I am sorry as I've probably misguided you, but this must use `useLocalObservable`, otherwise the test seems useless.
However I think these `enforceAction` tests are actually wrong. 
Take a look at the expectation: `expect(result.error).not.toBeDefined()`. I guess it was supposed to test that there are no errors/warnings, when calling a method (action)?
The effect  with `store.multiplier = multiplier` seems to be incorrectly copy pasted from other tests and since it was passing (due to incorrect expectation) it was left as is?
Also `"only when 'observed' should work"` and `"'never' should work"` are not tested correctly, because the `store.multipler` isn't used anywhere in the `render` (if it would then `'observed'` would actually warn as well).
What do you think, does it seem like a plausible explanation?

packages/mobx-react-lite/__tests__/useLocalStore.test.tsx
```diff
@@ -5,12 +5,15 @@ import { act, cleanup, fireEvent, render } from "@testing-library/react"
...
 
 afterEach(cleanup)
 
-afterEach(cleanup)
+let consoleWarnMock: jest.SpyInstance | undefined
+afterEach(() => {
+    consoleWarnMock?.mockRestore()
+})
 
 test("base useLocalStore should work", () => {
```
urugator, line 18: Please change to `useLocalObservable`

packages/mobx-react-lite/__tests__/useLocalStore.test.tsx
kubk: Ok, it's clear for me that these `enforceAction` tests should be reconsidered. But why useLocalObservable is required here? If we test `enforceAction` there is no difference between `useLocalObservable` and `useState(() => observable(...))`

packages/mobx-react-lite/__tests__/useLocalStore.test.tsx
kubk: Good catch. Done

packages/mobx-react-lite/__tests__/useLocalStore.test.tsx
urugator: > If we test enforceAction 

It's not supposed to test `enforceAction`, but `useLocalObservable` compatibility with `enforceAction`. 

> no difference between  useLocalObservable and useState(() => observable(...))

It's an implementation detail. IIRC `useLocalObservable` (or `useLocalStore`) did not use `action` for functions in the past, because we did not have `autoAction`. This was causing troubles, so these test were probably supposed to prove it's no longer an issue.
You're correct that since v6, `useLocalObservable` is defined as an alias for `useState(() => observable(...))`. So you could in principle rewrite all test in this file using `useState(() => observable(...))`, move them to a more appropriate location and eventually introduce a test proving that `useLocalObservable` is an alias for `useState(() => observable(...))`. Your call. One way or another these 3 test should be probably changed into something like:
```javascript
configure({ enforceActions: "..." })
const TestComponent = observer(() => {
  const obs = useLocalObservable(() => ({
    x: 1,
    inc() {
      obs.x += 1
    }
  }))

  useEffect(() => obs.inc()) // or replace with click handler if you want
  
  return obs.x;  // must subscribe
})
render(<TestComponent />)
// TODO expect it doesn't warn/throw
```
It shouldn't warn so no issue with unstable snapshot.

urugator: One more thing, these test should also finally call `_resetGlobalState()`.

mweststrate:
Thanks for tidying this up!

packages/mobx-react-lite/__tests__/useLocalStore.test.tsx
kubk: @urugator 

I've updated the PR. The summary of my changes:
- I've extracted `enforceAction` tests into a separate file
- Added `_resetGlobaState` after each test of this file
- Added `observer` and `useLocalObservable` as suggested + added console check after each test to make sure that there is actually a warning

kubk:
@urugator Friendly bump. Feel free to share your concerns if you have. There may be too many changes for such a simple feature. I can decrease the PR size by removing all the snapshots and suppressing all the `console.warn` calls. Tests shouldn't throw console warning because it's unclear what a developer should do with them. It's also difficult to keep track of them, because in the future we may get unwanted / unexpected `console.warn` calls and it'd nice if tests could detect that.

urugator (approved):
Feel free to merge to main

urugator:
Thank you!

mweststrate:
Great stuff @kubk! Thanks!

PR #3364, opened 2022-04-09
Fix individual backers link in main readme

jteration:
Individual backers link on https://mobx.js.org/README.html leads to a Not Found message due to relative pathing.

This will point to the file on GitHub which is what a user is taken to when clicking the link on https://github.com/mobxjs/mobx.


kubk:
Thank you 👍

PR #3382, opened 2022-04-30
chore: replace the deprecated react type definition

iChenLei:
> Alternative PR for #2741

### Code change checklist

### PR Backgroud
When I try to solve this problem #2721 , I noticed that mobx-react(and lite) source code use 
deprecated react type definition, Such as `RefForwardingComponent ` `React.Props`
and `StatelessComponent `, so I created a PR to replace these with official recommended type def.

#### RefForwardingComponent
```ts
  /**
   * @deprecated Use ForwardRefRenderFunction. forwardRef doesn't accept a
   *             "real" component.
   */
  interface RefForwardingComponent <T, P = {}> extends ForwardRefRenderFunction<T, P> {}
```
#### React.Props
```ts
  /**
   * @deprecated. This was used to allow clients to pass `ref` and `key`
   * to `createElement`, which is no longer necessary due to intersection
   * types. If you need to declare a props object before passing it to
   * `createElement` or a factory, use `ClassAttributes<T>`:
   *
   * ```ts
   * var b: Button | null;
   * var props: ButtonProps & ClassAttributes<Button> = {
   *     ref: b => button = b, // ok!
   *     label: "I'm a Button"
   * };
   * ```
   */
  interface Props<T> {
      children?: ReactNode;
      key?: Key;
      ref?: LegacyRef<T>;
  }
```

### DIFF
--- a/.changeset/brown-camels-smash.md
+++ b/.changeset/brown-camels-smash.md
@@ -1,6 +1,6 @@
 ---
-"mobx-react": patch
-"mobx-react-lite": patch
+"mobx-react": minor
+"mobx-react-lite": minor
 ---
 
 replace the deprecated react type definition with recommended type definition
--- a/.changeset/brown-camels-smash.md
+++ b/.changeset/brown-camels-smash.md
@@ -1,0 +1,6 @@
+---
+"mobx-react": patch
+"mobx-react-lite": patch
+---
+
+replace the deprecated react type definition with recommended type definition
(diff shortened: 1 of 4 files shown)


mweststrate:
Thanks for this! Would you mind marking it as minor? Beyond that looking good to me :)

.changeset/brown-camels-smash.md
```diff
@@ -0,0 +1,6 @@
+---
+"mobx-react": patch
```
mweststrate, line 2: Let's do minor, as there might be some nuisance where people with old React typings need to update those as well (not fully sure).

mweststrate (approved):

.changeset/brown-camels-smash.md
iChenLei: Ok, Sir.

PR #3387, opened 2022-05-03
React18 support

### LINKED ISSUE #3363: Install mobx-react peer dependency error on react v18.0.0
**Intended outcome:**

install mobx-react without any react peer dependency

**Actual outcome:**

when installing mobx-react , "could not resolve dependency" and searching for react 16 || 17

install react app, and it install react 18, and when try to install mobx-react

**Versions**

react@18.0.0

mweststrate:
This pull adds experimental / poor man's support for React 18. Fixes #3363, #2526. Supersedes #3005

* Updated tests, test / build infra, peerDependencies to React 18
* **[breaking icmw upgrading to React 18]** Already deprecated hooks like `useMutableSource` will trigger warnings in React 18, which is correct and those shouldn't be used anymore.
* **[breaking icmw upgrading to React 18]** When using React 18, it is important that `act` is used in **unit tests** around every programmatic mutation. Without it, changes won't propagate!
* The React 18 support is poor-mans support; that is, we don't do anything yet to play nicely with suspense features. Although e.g. [startTransition](https://github.com/mweststrate/platform-app/commit/bdd995773ddc6551235a4d2b0a4c9bd57d30510e) basically works, MobX as is doesn't respect the suspense model and will always reflect the latest state that is being rendered with, so tearing might occur. I think this is in theoretically addressable by using `useSyncExternalStore` and capturing the current values with together with the dependency tree of every component instance. However that isn't included in this pull request 1) it would be a breaking change, whereas the current change is still compatible with React 16 and 17. 2) I want to collect use cases where the tearing leads to problems first to build a better problem understanding. If you run into the problem, please submit an issue describing your scenario, and a PR with a unit tests demonstrating the problem in simplified form. For further discussion see  #2526, #3005


packages/mobx-react-lite/__tests__/strictAndConcurrentMode.test.tsx
```diff
@@ -42,36 +41,32 @@ test("uncommitted observing components should not attempt state changes", () =>
     }
 })
 
-const strictModeValues = [true, false]
+const strictModeValues = [true]
```
kubk, line 44: If checking in non-strict mode is no longer necessary, can we get rid of `.forEach` loop here?

mweststrate:
Writing a good changelog what is and is not addressed. Hope to get to that tomorrow😅

kubk (approved):

.changeset/swift-ravens-dress.md
```diff
@@ -0,0 +1,11 @@
...
+-   The React 18 support is poor-mans support; that is, we don't do anything yet to play nicely with suspense features. Although e.g. [startTransition](https://github.com/mweststrate/platform-app/commit/bdd995773ddc6551235a4d2b0a4c9bd57d30510e) basically works, MobX as is doesn't respect the suspense model and will always reflect the latest state that is being rendered with, so tearing might occur. I think this is in theoretically addressable by using `useSyncExternalStore` and capturing the current values with together with the dependency tree of every component instance. However that isn't included in this pull request 1) it would be a breaking change, whereas the current change is still compatible with React 16 and 17. 2) I want to collect use cases where the tearing leads to problems first to build a better problem understanding. If you run into the problem, please submit an issue describing your scenario, and a PR with a unit tests demonstrating the problem in simplified form. For further discussion see #2526, #3005
```
vkrol, line 11:
```suggestion
-   The React 18 support is poor man's support; that is, we don't do anything yet to play nicely with Suspense features. Although e.g. [startTransition](https://github.com/mweststrate/platform-app/commit/bdd995773ddc6551235a4d2b0a4c9bd57d30510e) basically works, MobX as is doesn't respect the Suspense model and will always reflect the latest state that is being rendered with, so tearing might occur. I think this is in theoretically addressable by using `useSyncExternalStore` and capturing the current values together with the dependency tree of every component instance. However that isn't included in this pull request 1) it would be a breaking change, whereas the current change is still compatible with React 16 and 17. 2) I want to collect use cases where the tearing leads to problems first to build a better problem understanding. If you run into the problem, please submit an issue describing your scenario, and a PR with a unit tests demonstrating the problem in simplified form. For further discussion see #2526, #3005
```

akphi:
@mweststrate sorry, what does `icmw` mean?

mweststrate:
In combination with. Sorry should have spelled that out.

Issue #3392, opened 2022-05-11
mobx-react: computed properties aren't recomputed while accessing React context with static contextType


pixelkritzel:
This is a CodeSandbox displaying the bug.

https://codesandbox.io/s/zealous-tesla-wioeys?file=/src/Item.tsx

The checkboxes don't update.

This version works because the context field is made observable. But this causes warnings in the console.

https://codesandbox.io/s/sweet-wilbur-2m70bd?file=/src/Item.tsx

urugator:
Guess we would have to make `this.context` implicitly observable similary to `this.props` and `this.state`:
https://github.com/mobxjs/mobx/blob/2e14aad3e283bf75041e4bfc956ddc9ce5b6580b/packages/mobx-react/src/observerClass.ts#L46-L51
@mweststrate opinions?

Hackish workaround
```javascript
@observable
_context!: FormContextType;
set context(val: FormContextType){
  runInAction(() => this._context = val)    
};
get context() {
  return this._context;
}
```

You can also avoid relying on context propagation and pass something stable+observable that children can subscribe to (eg Store instance itself).

pixelkritzel:
@mweststrate 

My assumption was this would to the same trick. But obviously it doesn't work.

```
@observable
  selectedId: FormContextType["selectedId"];

  @action
  setSelectedId = (id: string) => {
    console.log("setting value");
    this.selectedId = id;
  };

  @computed
  get contextValue(): FormContextType {
    return { selectedId: this.selectedId, setSelectedId: this.setSelectedId };
  }
  
 ```

**Comment by mweststrate** [comment]
That is because you copy selectedId into a new non obsevable object. It's
the property you made observable,  but the value you're passing down
(primitive values dont change themselved)

**Comment by mweststrate** [comment]
https://mobx.js.org/understanding-reactivity.html for some more context. But note that this doesn't really relate to MobX, but just to how JavaScript works :) What you did is roughly the equivalent to:

```
const store = {
  selectedId: 1
}

context = {
   selectedId: store.selectedId // this copies the current value into a new prop, doesn't alias!
}

printSelectedId() {
  console.log(store.selectedId)
}

printSelectedId() // prints 1

store.selectedId = 2

printSelectedId() // still prints 1, not 2! context and store are unrelated objects

```

urugator:
@mweststrate 
>  But note that this doesn't really relate to MobX

[...]

PR #3404, opened 2022-05-19
 Bugfix #3392

pixelkritzel:
Bugfix [#3392](https://github.com/mobxjs/mobx/issues/3392)

### Code change checklist

### DIFF
--- a/packages/mobx-react/__tests__/observer.test.tsx
+++ b/packages/mobx-react/__tests__/observer.test.tsx
@@ -957,11 +957,10 @@ test("this.context is observable if ComponentName.contextType is set", () => {
     }
 
     const parentRef = React.createRef<Parent>()
-    const childRef = React.createRef<Child>()
 
     const app = (
         <Parent ref={parentRef}>
-            <Child ref={childRef} />
+            <Child />
         </Parent>
     )
 
--- a/packages/mobx-react/__tests__/observer.test.tsx
+++ b/packages/mobx-react/__tests__/observer.test.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { createContext } from "react"
 import { inject, observer, Observer, enableStaticRendering } from "../src"
 import { render, act } from "@testing-library/react"
 import {
@@ -907,4 +907,72 @@ test("Missing render should throw", () => {
         }
     }
     expect(() => observer(Component)).toThrow()
+})
+
+test("this.context is not observable if ComponentName.contextType is not set", () => {
+    const store = observable({ counter: 0 })
+
+    const Context = createContext(store)
+
+    @observer
+    class Component extends React.Component {
+        context = { counter: 0 }
+
+        @computed
+        get counterAsString() {
+            return String((this.context as any).counter)
+        }
+
+        render() {
+            return <div />
+        }
+    }
+
+    const compRef = React.createRef<Component>()
+
+    const { unmount } = render(
+        <Context.Provider value={store}>
+            <Component ref={compRef} />
+        </Context.Provider>
+    )
+
+    store.counter = 1
+
+    expect(compRef.current?.counterAsString).not.toBe("1")
+    unmount()
+})
+
+test("this.context is observable if ComponentName.contextType is set", () => {
+    const store = observable({ counter: 0 })
+
+    const Context = createContext(store)
+
+    @observer
+    class Component extends React.Component {
+        static contextType = Context
+
+        @computed
+        get counterAsString() {
+            return String((this.context as any).counter)
+        }
+
+        render() {
+            return <div />
+        }
+    }
+
+    const compRef = React.createRef<Component>()
+
+    const { unmount } = render(
+        <Context.Provider value={store}>
+            <Component ref={compRef} />
+        </Context.Provider>
+    )
+
+    store.counter = 1
+
+    console.log(compRef.current?.counterAsString)
+
+    expect(compRef.current?.counterAsString).toBe("1")
+    unmount()
 })
--- a/packages/mobx-react/__tests__/observer.test.tsx
+++ b/packages/mobx-react/__tests__/observer.test.tsx
@@ -912,15 +912,27 @@ test("Missing render should throw", () => {
 test("this.context is observable if ComponentName.contextType is set", () => {
     const Context = createContext({})
 
-    class Parent extends React.Component<any> {
-        state = { counter: 0 }
+    let renderCounter = 0
 
-        componentDidMount() {
-            this.setState({ counter: 1 })
+    @observer
+    class Parent extends React.Component<any> {
+        constructor(props: any) {
+            super(props)
+            makeObservable(this)
+        }
+
+        @observable
+        counter = 0
+
+        @computed
+        get contextValue() {
+            return { counter: this.counter }
         }
 
         render() {
-            return <Context.Provider value={this.state}>{this.props.children}</Context.Provider>
+            return (
+                <Context.Provider value={this.contextValue}>{this.props.children}</Context.Provider>
+            )
         }
     }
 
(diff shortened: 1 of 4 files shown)


urugator:
run `yarn changeset` in root dir, choose mobx-react and minor (spacebar to check, enter to confirm)

packages/mobx-react/__tests__/observer.test.tsx
```diff
@@ -908,3 +908,71 @@ test("Missing render should throw", () => {
     }
     expect(() => observer(Component)).toThrow()
 })
+
+test("this.context is not observable if ComponentName.contextType is not set", () => {
+    const store = observable({ counter: 0 })
```
urugator, line 913: You don't need any stores or observables here. All we have to test is:
Parent component has a state (eg number).
Update parent component state (eg increment the number).
Parent component passes this state (number) via context to child.
Child returns `this.context` from `@computed`.
Child returns `@computed` in render.
Prior fix the child's output won't correlate with parent's state.

pixelkritzel: It's better now?

packages/mobx-react/__tests__/observer.test.tsx
```diff
@@ -908,3 +908,46 @@ test("Missing render should throw", () => {
...
+
+    const { unmount } = render(
+        <Parent>
+            <Child ref={compRef} />
+        </Parent>
+    )
+
+    console.log(compRef.current?.counterValue)
+
+    expect(compRef.current?.counterValue).toBe(1)
```
urugator, line 951: Have you tried to run the test without the fix? I think this would pass even before.

pixelkritzel: This passes in both cases. But this why I used the store before. Like here https://github.com/mobxjs/mobx/blob/36d3d35047c3002376fca676d79399f381af7319/packages/mobx-react/__tests__/observer.test.tsx#L56

packages/mobx-react/__tests__/observer.test.tsx
```diff
@@ -908,3 +908,46 @@ test("Missing render should throw", () => {
...
+    class Child extends React.Component {
+        static contextType = Context
+
+        @computed
+        get counterValue() {
+            return (this.context as any).counter
+        }
+
+        render() {
+            return <div />
```
mweststrate, line 937: what you want to test here as well is `<div>{this.counterValue}</div>`, is getting rerendered (once) to make sure that the `@observer` as well picks up the change in the computed value, as createObservableProp sets some internal flags, for which I'm not 100% sure that is correct in the case of context (I suspect it is correct)

packages/mobx-react/__tests__/observer.test.tsx
mweststrate: A test that passes before and aft doesn't sound like a good test :) I can't see the original test clearly anymore, would you mind adding that as well for now at least? I do think atm this PR is correct, but that it isn't the "correct" solution to the original problem, from scanning the original issues 😅

pixelkritzel: You can see the original test in this commit.

https://github.com/mobxjs/mobx/pull/3404/commits/4088edd197ef79f694e9245544f4dee48281cce5

mweststrate: To clarify; in the original problem your 'store' isn't a store at at all, it is an unstable object reference for something that should be stable in the mobx idioms, and the correcter way to solve it would be simple: `<FormContext.Provider value={this}>` and skip that whole computed thing. Computed values shouldn't normally be new objects that are observable / mutable; either they are a ref to an already existing object, or they are immutable value objects, just representing the derived value at a certain moment in time. 

In your current setup you create many objects that might all live at the same moment in time, representing many different selections. 

Edit: fixed snippet

urugator: Just use the computed value inside the render and it will fail. The computed doesn't cache if it's not observed, therefore the problem doesn't occur.

mweststrate: Ah, for the first test, yeah that didn't seem to make sense indeed, that should pass even if context would be an observable.

pixelkritzel:
@urugator @mweststrate 

New test is up.

It checks

- [x] the output of `<Child>` render function 
- [x] how often `Child.render` runs.

The `@computed` on the parent is necessary to fail if `Child.contextType` is undefined. This is the same as [original example](https://codesandbox.io/s/zealous-tesla-wioeys?file=/src/Form.tsx).

pixelkritzel:
And funny thing I discovered is `this.context` is always defined (even if `contextType` is not set)  and an empty object which event overrides values set on the class.

https://codesandbox.io/s/heuristic-mayer-th9058?file=/src/App.tsx

packages/mobx-react/__tests__/observer.test.tsx
```diff
@@ -908,3 +908,72 @@ test("Missing render should throw", () => {
     }
     expect(() => observer(Component)).toThrow()
 })
+
+test("this.context is observable if ComponentName.contextType is set", () => {
+    const Context = createContext({})
+
+    let renderCounter = 0
+
+    @observer
```
mweststrate, line 917: Thanks this test looks a lot better! Still want to point out that this component doesn't have to be an observer at all if you'd put an observable object as context value, saving re-rendering the root of the tree and any unmemo-ed child thereunder, whenever you increase the counter. That is the way it is typically done, and I suspect also the reason why no one reported this earlier.

pixelkritzel: What difference does it make to the render passes whether the context value is in an observable or in a computed?

The component must react either way to pass the new value to the context.

mweststrate: In your setup, it is `Parent` that reacts to a change in `counter`, by creating a new object (at line 929) as part of the render process to compute `this.contextValue` (you basically make an unnecessary state copy here). If you'd pass a stable observable object (or even `this`) as Provider value, the counter itself isn't read at all by the Parent, but instead by the Child, so that the effect of changing `counter` would be much more local effect. So instead of re-rendering in principle the whole subtree of Parent, you'd only be re-rendering individual Child components.

pixelkritzel: So basically cutting the Reacts render logic completely out and using parent only as Context Provider? This is kind of strict.

pixelkritzel: In real life I often have something like this:

```
  @computed
  get contextValue(): DropdownContextType {
    return {
      toggleDropdown: this.toggleDropdown,
      closeDropdown: this.closeDropdown,
      open: this.open,
      type: this.props.type || this.computedType,
    };
  }
  ```
  
 and `this.open` is also accessed in render to determine if dropdown should be rendered open or not. So it doesn't matter here.

Comment (mweststrate) [reply packages/mobx-react/__tests__/observer.test.tsx:917]: I don't see the benefit of constructing that object over and over again over just passing `this` (you strongly typed out the other members anyway). If you really want to strict down the object passed, you could also allocate it once and just make it a delegator object, so that you can keep using the same `contextValue` even when you're state changes over time:

```
contextValue = {
  toggleDropdown: this.toggleDropdown,
  closeDropdown: this.closeDropdown,
  get open() { return this.open }, 
  get type() { return this.props.type || this.computedType }
}
```

> So basically cutting the Reacts render logic completely out and using parent only as Context Provider?

If there is no need for it yes, if there is need for it, fine as well. In you're cases shown so far there is no need (as in, there is no visual changes to your UI 'above' the Child components.

mweststrate (approved):

packages/mobx-react/__tests__/observer.test.tsx
pixelkritzel: But this way you get into nasty `this` bugs.

https://codesandbox.io/s/admiring-stallman-hcnrpv?file=/src/Form.tsx

pixelkritzel: Thank you for your help and support! And for Mobx obviously! Luckily it came out early enough so that I didn't dipped to much in Redux.

pixelkritzel:
@urugator can this PR merged?

urugator:
Thank you

PR #3409, opened 2022-05-20
Remove observable.box type inconsistancy

Nokel81:
Unlike `observable.map`, `observable.set`, and `observable.array` where it is possible to create a type correct empty instance without any initial values but like `observable.object` I think that `observable.box` should require an initial value.

This is because the following code type checks but will throw an error at runtime:

```ts

const myObvs = observable.box<string>();
myObvs.get().includes("hello world"); // this will throw a TypeError
```

### Code change checklist


packages/mobx/__tests__/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "../tsconfig.json",
```
kubk, line 2: @Nokel81 Could you describe why a new config file was introduced here?

packages/mobx/__tests__/v5/base/observables.ts
```diff
@@ -0,0 +1,18 @@
+import { observable } from "../../../src/mobx"
+
+test("argumentless observable adds undefined to the output type", () => {
```
kubk, line 3: Currently all the TS related tests are located in the [typescript-tests.ts](https://github.com/mobxjs/mobx/blob/main/packages/mobx/__tests__/v5/base/typescript-tests.ts) file. Could you move these tests to that file?

packages/mobx/src/api/observable.ts
```diff
@@ -175,17 +180,15 @@ export interface IObservableFactory extends Annotation, PropertyDecorator {
...
         return new ObservableValue(value, getEnhancerFromOptions(o), o.name, true, o.equals)
     },
     array<T = any>(initialValues?: T[], options?: CreateObservableOptions): IObservableArray<T> {
         const o = asCreateObservableOptions(options)
-        return (
-            globalState.useProxies === false || o.proxy === false
-                ? createLegacyArray
-                : createObservableArray
-        )(initialValues, getEnhancerFromOptions(o), o.name)
+        return (globalState.useProxies === false || o.proxy === false
```
kubk, line 189: Was it added by prettier?

kubk:
@urugator What are your thoughts on it? For me it seems like a valid change. I can't think of a usecase where `observable.box` should be empty by default.

Nokel81:
What does `TypeError: packages/mobx/__tests__/v5/base/observables.js: Emit skipped for language service` mean on the test failures?

kubk (changes requested):

kubk:
@Nokel81 

> What does `TypeError: packages/mobx/__tests__/v5/base/observables.js: Emit skipped for language service` mean on the test failures?

It means that TS is trying to process a file that shouldn't be processed. A possible cause is the new `tsconfig` introduced in this PR. Can you remove it and check if the error is gone?

.changeset/good-toes-wave.md
```diff
@@ -0,0 +1,8 @@
+---
+"mobx": minor
+"mobx-react": minor
+"mobx-react-lite": minor
+"mobx-undecorate": minor
```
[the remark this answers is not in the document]
Nokel81: For some reason the `npx changelog` command thought that all of them were affected. Will remove.

packages/mobx/__tests__/tsconfig.json
Nokel81: This is required because otherwise `strict` mode was not being applied to the typescript test files which causes `string | undefined` to be flattened to `string` thus making the tests not pass.

packages/mobx/src/api/observable.ts
Nokel81: Will revert all stylistic changes, they were made but some auto formatter.

Nokel81:
> It means that TS is trying to process a file that shouldn't be processed. A possible cause is the new tsconfig introduced in this PR. Can you remove it and check if the error is gone?

I need the new tsconfig file to make sure that `strict` mode is enabled for the test file. Will try and fix it.

urugator:
CI got stuck again or whats the status here?

kubk:
Hi @Nokel81 
I've pulled `main` branch locally and added your tests to the `typescript-tests.ts` file. The test fails as expected even without the new `tsconfig.json`. How did you run the tests? I tried with 2 commands - `yarn test` in the root folder and `cd packages/mobx && yarn test`. Both of them fail. It means that test are being executed with TS strict mode or at least `strictNullCheck` TS option.

Let me clarify. It's better to avoid adding a new `tsconfig.json` configuration file because it patches the issue rather than fixes it. Ideally we should have only 2 config files  - one for tests and another one for build. Could you verify locally that this new `tsconfig.json` is not needed?

kubk:
@urugator @Nokel81 I'd be happy to hear your opinion on that:

What if we just forbid to create `observable.box()` with no initial value? Here are the benefits:

- It solves the initial issue. A user will have to initialize the box with undefined or null and this value will be propagated to the return type. Example:

```ts
const a = observable.box<string | undefined>(undefined); // IObservableValue<string | undefined>
const b = observable.box<string | null>(null); // IObservableValue<string | null>
```

- It won't complicate typings. It should be enough to remove the undefined type (`?`) from the first argument here: https://github.com/mobxjs/mobx/blob/8a7c5b36e2cae0d0d963b74769bc2f06c3fa5e4a/packages/mobx/src/api/observable.ts#L149

The downsides is that it is kinda BC break. But in my opinion it's a bug fix rather than a BC break. This solution also forces a developer to be more explicit, but this explicity also helps TS to infer types correctly.

urugator:
I think it's not a design issue, but a typesystem problem. If something can exist without a value, it feels somehow natural to make that value optional and on the contrary a bit artifical/weird to actually require `box(undefined)`. But I am not 100% sure about that.
I would prefer if it would keep working without a value, but I am fine if it wouldn't compile with TS. If we could get both it would be the best at this point, so we don't have to deal with BC.

Nokel81:
@kubk what do you mean that they fail even without the new tsconfig? They pass with the new tsconfig.

As far as I can see CI is green. Without the tsconfig then `tsc` would complain about the unnecessary `// @ts-ignore` comments.

kubk:
@Nokel81 

> Without the tsconfig then tsc would complain about the unnecessary // @ts-ignore comments.

I've added a few changes to the PR to check that. First I added a line that won't compile in TS strict mode: https://github.com/mobxjs/mobx/pull/3409/commits/54914766786384d0d758245bf4dce28d83bb9a81 , then I removed the new config file: https://github.com/mobxjs/mobx/pull/3409/commits/6e06dc621d4b6a94322ded746242115bf61a934d

Here is related CI job: https://app.circleci.com/pipelines/github/mobxjs/mobx/2025/workflows/b569591f-17a6-4b57-82fb-6905af65506d/jobs/8916

It fails with an error: "Type 'null' is not assignable to type 'string'". It means that tests are already being executed in TS strict mode and we don't have to add a new config file. Lastly I've replaced `ts-expect-error` with exact type assertions: https://github.com/mobxjs/mobx/pull/3409/commits/0d113eec797eed28bf58cd2029b7c7658642dae3

Here we can be more specific and check exact types to avoid false positives in the future. The CI passes and I'll merge the PR. Thank you for your help and patience 👍

kubk (approved):

kubk:
@urugator Thanks for you input! Indeed there is no best choice, so let's avoid BC at least.

PR #3410, opened 2022-05-20
chore(hooks): update useObserver for no-conditional-hooks compliance

j-mendez:
* update useObserver to comply with no conditional hooks. ( May become standard upon future releases ). 

### Code change checklist


kubk:
Hi @j-mendez 
`useObserver` is deprecated: https://github.com/mobxjs/mobx/tree/main/packages/mobx-react-lite#deprecated-apis

It is either going to be removed from public API or even removed completely

PR #3421, opened 2022-06-07
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -2,8 +2,14 @@ name: Coveralls
 
 on: ["push", "pull_request"]
 
+permissions:
+  contents: read
+
 jobs:
   mobx:
+    permissions:
+      checks: write  # for coverallsapp/github-action to create new checks
+      contents: read  # for actions/checkout to fetch code
     name: Packages coverage
     runs-on: ubuntu-latest
     steps:


.github/workflows/coveralls.yml
```diff
@@ -2,8 +2,14 @@ name: Coveralls
 
 on: ["push", "pull_request"]
 
+permissions:
+  contents: read
+
 jobs:
   mobx:
+    permissions:
+      checks: write  # for coverallsapp/github-action to create new checks
```
mweststrate, line 11: @naveensrinivasan could you elaborate a bit more why a write permissions is needed for coveralls? I'd expect only read would be enough, and I didn't see anything in there docs?

naveensrinivasan: Let me get back to you on this. It's been a while. Thanks

urugator: @naveensrinivasan friendly bump

naveensrinivasan: Can someone please take over this PR? I don't have the bandwidth now.  Thanks

PR #3434, opened 2022-06-25
Fix #3395: observer not working with <React.StrictMode>

### LINKED ISSUE #3395: React 18:  `observer` not working with <React.StrictMode>
See https://github.com/mobxjs/mobx/discussions/3393#discussioncomment-2737011 for explanation and fix proposal.

### Discussed in https://github.com/mobxjs/mobx/discussions/3393

<div type='discussions-op-text'>

<sup>Originally posted by **acataluddi** May 11, 2022</sup>
Hi guys,

I was wondering whether any of you experienced any issue with class-based React components and MobX when the `<React.StrictMode>` is enabled. I've created a quick [CodeSandbox Example](https://codesandbox.io/s/7suusz) to explain the issue.

There is a very basic MobX observable store and the following function which triggers the `enableService()` *action*:

```typescript
window.setTimeout(() => {
  store.enableService().then();
}, 1500);
```

The strange thing is that while functional components get updated as expected (the first "service is enabled string) the class-based one is not (the green one):

immagine

However, by removing the `<React.StrictMode>` wrapper, everything works as expected:

```tsx
// Doesn't work
root.render(
  <StrictMode>
    <App />
  </StrictMode>
);

// Works
root.render(
    <App />
);
```

The same code was working without problems in older version of React even with the `<React.StrictMode>` enabled.

Is it possible there is any bug or it's just me missing something obvious?

Thanks,
Adriano

</div>

urugator:
Fixes #3395

### DIFF
--- a/.changeset/clean-islands-retire.md
+++ b/.changeset/clean-islands-retire.md
@@ -1,0 +1,5 @@
+---
+"mobx-react": patch
+---
+
+Support re-mounting of class components. Fixes #3395: observer not working with React@18 &lt;StrictMode&gt;.
--- a/packages/mobx-react/__tests__/observer.test.tsx
+++ b/packages/mobx-react/__tests__/observer.test.tsx
@@ -1,4 +1,4 @@
-import React, { createContext } from "react"
+import React, { createContext, StrictMode } from "react"
 import { inject, observer, Observer, enableStaticRendering } from "../src"
 import { render, act } from "@testing-library/react"
 import {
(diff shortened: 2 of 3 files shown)


urugator:
I wonder whether instead of patching `componentDidMount`, we could just call `forceUpdate` in `componentWillUnmount`, so the component is immediately dirty (and therefore forced to re-render) in case it re-mounts. The doc states:

> You should not call setState() in componentWillUnmount() because the component will never be re-rendered. Once a >component instance is unmounted, it will never be mounted again.

But that's obviously no longer true...

mweststrate:
Thanks for investigating @urugator! Looks great

PR #3458, opened 2022-07-09
Documentation: A slight revamp of the README, wording, and clearer links

egilll:
Hi. This pull request changes a handful of things which I found to be confusing when reading this otherwise excellent documentation. I apologize if this PR is too bold — do of course feel free to close this PR should you disagree with these changes.

The changes are:

* I found the lack of a link to the API in the sidebar to be extremely confusing. It turns out that it was in the header, but it still always takes me a while to remember where to look for it.
* The API docs include links to more detailed articles on some topics, but that link is "hidden" behind the "Usage:" label. I did not figure that out at first, and so I've changed the link to a "(further information)" label.
* In "Creating custom observables", it was not clear that "the concept of atoms" refers to a class used interally by MobX.
* README:
  * The main README in GitHub did not include a link to the main documentation at mobx.js.org which I occasionally found confusing.
  * <s>I believe it is better to let readers know *what* the project is about before showing a long list of sponsors, so I've moved the introduction above the list of sponsors.</s>
  * I removed the "What others are saying" section. The quotes were not convincing and fairly childish, which I associate mainly with brand new, non-stable projects.
  * Changed the link to the MobX book from an incomplete Google Books preview to PacktPub/Amazon.
  * Changed the term "Transparent Functional Reactive Programming" to simply "functional reactive programming". The README read as if TFRP is some general term one could be expected to know, but no, it is practically non-existent, not being used by anyone apart from MobX and two (low-impact) papers. "Transparent Reactive Programming" was used internally by someone at Meteor but that term did not catch on either. In my view it is therefore not helpful to readers to include it.
* Removed the sentence "Collections such as arrays, Maps and Sets are made observable automatically" from the page "Creating observable state". It appears that this claim is missing some additional context.
* Added "(£5)" to "Download the Mobx cheat sheet".
* Also some minor wording fixes.

PR #3477, opened 2022-07-23
Coveralls badges

kubk:
The PR brings coveralls.io badges back. I've also removed Typescript/Prettier badges because nowadays these tools are de facto standard. Let's avoid excess information to not overwhelm reader.

Regarding allowing coveralls job to fail on CI, there is no clean solution yet, the issue is still open: https://github.com/actions/toolkit/issues/399

For example, one of the workarounds https://github.com/actions/toolkit/issues/399#issuecomment-607465968 would allow to merge a PR with optional failed job, but GitHub will show the whole PR as failed (with ❌ symbol). 

@urugator How often do you encounter coveralls random fails?

Here coveralls job failed because the tests failed too: https://github.com/mobxjs/mobx/pull/3458/commits/be29cd01257004c4433414ea4d3101d459ef95c8

### DIFF
--- a/packages/mobx-react-lite/README.md
+++ b/packages/mobx-react-lite/README.md
@@ -1,9 +1,8 @@
 # mobx-react-lite
 
-[![CircleCI](https://circleci.com/gh/mobxjs/mobx-react-lite.svg?style=svg)](https://circleci.com/gh/mobxjs/mobx-react-lite)[![Coverage Status](https://coveralls.io/repos/github/mobxjs/mobx-react-lite/badge.svg)](https://coveralls.io/github/mobxjs/mobx-react-lite)[![NPM downloads](https://img.shields.io/npm/dm/mobx-react-lite.svg?style=flat)](https://npmjs.com/package/mobx-react-lite)[![Minzipped size](https://img.shields.io/bundlephobia/minzip/mobx-react-lite.svg)](https://bundlephobia.com/result?p=mobx-react-lite)
-
-[![TypeScript](https://badges.frapsoft.com/typescript/code/typescript.svg?v=101)](https://github.com/ellerbrock/typescript-badges/)[![code style: prettier](https://img.shields.io/badge/code_style-prettier-ff69b4.svg)](https://github.com/prettier/prettier)
-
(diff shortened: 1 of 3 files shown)


packages/mobx-react-lite/README.md
```diff
@@ -1,9 +1,8 @@
...
-
-[![TypeScript](https://badges.frapsoft.com/typescript/code/typescript.svg?v=101)](https://github.com/ellerbrock/typescript-badges/)[![code style: prettier](https://img.shields.io/badge/code_style-prettier-ff69b4.svg)](https://github.com/prettier/prettier)
-
+[![CircleCI](https://circleci.com/gh/mobxjs/mobx-react-lite.svg?style=svg)](https://circleci.com/gh/mobxjs/mobx-react-lite)
+[![Coverage Status](https://coveralls.io/repos/github/mobxjs/mobx-react-lite/badge.svg)](https://coveralls.io/github/mobxjs/mobx-react-lite)
+[![NPM downloads](https://img.shields.io/npm/dm/mobx-react-lite.svg?style=flat)](https://npmjs.com/package/mobx-react-lite)[![Minzipped size](https://img.shields.io/bundlephobia/minzip/mobx-react-lite.svg)](https://bundlephobia.com/result?p=mobx-react-lite)
 [![Discuss on Github](https://img.shields.io/badge/discuss%20on-GitHub-orange)](https://github.com/mobxjs/mobx/discussions)
```
kubk, line 6: https://github.com/mobxjs/mobx/tree/main/packages/mobx-react-lite#readme

Badges are to close to each other, I've added a bit of space

urugator:
Do we know why it fails? To be clear I dunno if badges are always the cause. It seems quite random, so I thought it's because of race conditions or someting, but if it's just some service being unresponsive, then there is not much we can do I guess (besides increasing timeout).

kubk:
@urugator Sorry for confusing you. The badges has nothing to do with CI. They just query some information from an endpoint and display it.

Regarding CI failures, I saw a few times how Coveralls Github action failed due to 504 from Coveralls API. Yes, there isn't much we can do because it seems like GitHub doesn't have a way to make job optional. In GitLab you can allow job to fail and still get the green pipeline: https://docs.gitlab.com/ee/ci/yaml/#allow_failure

Let's keep an eye on CI to understand how often we encounter such a situation. Then if it fails often, we can apply the workaround described in the PRs description.

urugator (approved):

Issue #2777, opened 2021-02-02
Mobx codesandbox is outdated


kubk:
The link used in GitHub mobx template is outdated:
https://codesandbox.io/s/minimal-mobx-react-project-ppgml

![image](https://user-images.githubusercontent.com/22447849/106613891-c2e36f80-657b-11eb-9e04-72ec062f7e4e.png)

Mobx and React versions should be updated, the code should be updated to Mobx strict actions mode. Is this issue template under Git control, can I change it?

EDIT: Here is up-to-date version: https://codesandbox.io/s/magical-dhawan-jq8t5?file=/index.js

FredyC:
We have MobX workspace on CodeSandbox, I've invited you there. It's the better way than having different versions all over the place.

kubk:
@FredyC Thank you. Codesandbox documentation says that to edit a template I should unfreeze it first, otherwise Codesandbox will create a fork: https://codesandbox.io/docs/templates 

For some reason the "unfreeze" option is not available for me:

![image](https://user-images.githubusercontent.com/22447849/106720889-0f2ebe00-6615-11eb-9dd5-b74b1abd78d5.png)

Codesandbox shows me that I am a member of Mobx docs: https://codesandbox.io/u/kubk

Am I missing some permissions? Anyway, you can copy the code from my fork and update all the dependencies to the latest: https://codesandbox.io/s/magical-dhawan-jq8t5?file=/index.js

FredyC:
That's really strange, you have the same permissions as me (Editor). It's updated now, thanks for noticing it.

![image](https://user-images.githubusercontent.com/1096340/106734696-98dd8c00-6613-11eb-93a8-620c68c86c22.png)

PR #3483, opened 2022-07-27
Add React Native codesandbox to the issue template

kubk:
The PR adds React Native codesandbox link to the issue template. This codesandbox has a web preview, so it's much more easier to check than installing React Native with a bunch of related tools like emulator locally. For example here https://github.com/mobxjs/mobx/discussions/3480 the provided codesandbox helped to narrow down the issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -27,7 +27,11 @@ A description of what happened, including a screenshot or copy-paste of any rela
 **How to reproduce the issue:**
 
 <!--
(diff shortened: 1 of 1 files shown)


urugator:
Awesome, thank you!

Btw do we have any access to that https://codesandbox.io/s/minimal-mobx-react-project-ppgml or it can be modified only by author? Would be nice to update the deps there.

Seems like *Projects* could help with this a bit: https://codesandbox.io/docs/projects/learn/integrations/github-app#privacy-and-permissions

urugator (approved):

kubk:
@urugator No we don't, would you like to change something there?

Once I tried to fix the code in that codesandbox, FredyC added me as a member but it didn't work out: https://github.com/mobxjs/mobx/issues/2777#issuecomment-772340507

Eventually I asked FredyC to just apply my changes manually

PR #3487, opened 2022-07-30
refactor: using useRef in useLocalObservable

misaeldossantos:
### Code change checklist

Change useState to useRef in useLocalObservable, creating a mutable reference.


kubk:
Hey, could you describe the necessity of that change?

misaeldossantos:
i think useRef is more appropriate, because the mobx observable object is a mutable reference

kubk:
I'd say that both `useState` and `useRef` are appropriate here because they remember their data after re-renders. The approach with `useRef` is just more verbose (hence more bytes in the bundle). Thank you for the desire to improve things, but I think this change is not necessary.

PR #3491, opened 2022-08-08
Update readme

iminside:
Very modest fix. When converting to inches, we need to divide the number by 2.54. 😊

### Code change checklist


kubk:
Correct, thanks. Keep in mind that useAsObservableSource is going to be removed in the future because it's deprecated.

PR #3504, opened 2022-08-17
docs: Fix misleading direction

brunocoelho:
Hi there.

The example is below the explanation, not above.

### DIFF
--- a/docs/react-integration.md
+++ b/docs/react-integration.md
@@ -239,7 +239,7 @@ You might be wondering, when do I apply `observer`? The rule of thumb is: _apply
 `observer` works best if you pass object references around as long as possible, and only read their properties inside the `observer` based components that are going to render them into the DOM / low-level components.
 In other words, `observer` reacts to the fact that you 'dereference' a value from an object.
 
-In the above example, the `TimerView` component would **not** react to future changes if it was defined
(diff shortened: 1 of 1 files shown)


kubk:
@brunocoelho Nice catch 👍

urugator:
Perhaps it could be formulated better, but I think the original statement is correct:

***In the above example**, the TimerView component **would not** react to future changes **if it was defined as follows**,*

It says what *would* happen to the example [*above*](https://mobx.js.org/react-integration.html#using-local-observable-state-in-observer-components), if you would change `TimerView` definition, as shown in *following* (below) snippet (it's not a complete example).

brunocoelho:
@urugator I agree with you. Misleading is not the right word, maybe just a little confusing.
What do you think if we change to:

~~In the above example,~~ _The `TimerView` component would **not** react to future changes if it was defined as follows..._

At this point, we know what TimerView is about because it was the last example used.

After your explanation, I'm fine leaving the documentation unchanged though.

urugator:
> What do you think if we change to:

Agree, I think it's sufficient and less confusing, thank you.

brunocoelho:
Cool, I will close this PR and create a new one with the change discussed.

PR #3515, opened 2022-09-05
Fix #3492: throw warning when use class component and suspense

### LINKED ISSUE #3492: An exception is thrown when using Mobx6 and react18 in development mode
**Intended outcome:**

When I use React18's new rendering API `ReactDOM.createRoot().render`, I get a lot of warnings. It should be guaranteed to be consistent with the previous behavior, run normally, and not throw warnings

warning message: react_devtools_backend.js:3973 Warning: Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously later calls tries to update the component. Move this work to useEffect instead.

**Actual outcome:**

I get a lot of warnings, but can't find the information I have in it, like below

**Versions**

mobx@^6.6.1 mobx-react@^7.5.1

Dzdw92s8dG

guochengcheng04:
Fix [#3492](https://github.com/mobxjs/mobx/issues/3492)

-   [x] unit test all passed.
-   [x] performance all passed.


guochengcheng04:
@mweststrate hi bro, do you have any ideas about this PR?

urugator (commented):
See the comment: https://github.com/mobxjs/mobx/issues/3492#issuecomment-1237326790

packages/mobx-react/src/utils/FinalizationRegistryWrapper.ts
```diff
@@ -0,0 +1,12 @@
+declare class FinalizationRegistryType<T> {
```
urugator, line 1: Any reason to not import these utils from `mobx-react-lite`?

packages/mobx-react/src/observerClass.ts
```diff
@@ -165,6 +191,11 @@ function createReactiveRender(originalRender: any) {
         isRenderingPending = false
         // Create reaction lazily to support re-mounting #3395
         const reaction = (reactiveRender[mobxAdminProperty] ??= createReaction())
+
+        if (!this[reactionTrack]) {
+            addReactionToTrack(this, reaction, {})
```
urugator, line 196: Move at the end of `createReaction`, otherwise you would have to clear `this[reactionTrack]` whenever `reactiveRender[mobxAdminProperty]` is cleared.

packages/mobx-react/src/observerClass.ts
```diff
@@ -73,9 +78,26 @@ export function makeClassComponentObserver(
...
+        recordReactionAsCommitted(this)
+        if (this[reactionTrack]) {
+            this[reactionTrack].mounted = true
+            if (!this.render[mobxAdminProperty] || this[reactionTrack].changedBeforeMount) {
+                // Reaction is re-created automatically during render, but a component can re-mount and skip render #3395.
+                // To re-create the reaction and re-subscribe to relevant observables we have to force an update.
+                Component.prototype.forceUpdate.call(this)
+                this[reactionTrack].changedBeforeMount = false
+            }
+        } else {
```
urugator, line 90: In order for this to happen, you have to unset `this[reactionTrack]` in `componentWillUnmount`, like here
https://github.com/mobxjs/mobx/blob/f0e066f427d573cbef4b92f3310eb069c3aca205/packages/mobx-react-lite/src/useObserver.ts#L103-L104
And also `this[reactionTrack]` must be unset when reaction is disposed in tracking utils, is it the case?

urugator:
Btw, I didn't want to get into it initially, but for completeness, it can also be done without tracking utils in the following fashion:
```
// timeout could be used as well
const queueMicrotask = window.queueMicrotask ?? (microtask) => Promise.resolve().then(microtask)

// in render
Promise.resolve().then(() => { 
  if (!mounted) {
     disposeReaction()
  }
})

// in componentDidMount
mounted = true
if (reactionDisposed) {
  recreateReaction() // can be done lazily in render
  forceUpdate()
} else if (reactionNotified) { // state changed in beetwen
  forceUpdate()
}
``` 
This is because `componenDidMount` is quaranteed to run sychronously (before the microtask).
The same can be done with functional components, but we would have to use `useLayoutEffect` instead of `useEffect` - could have negative impact on concurrent features, dunno.

packages/mobx-react/src/utils/FinalizationRegistryWrapper.ts
urugator: Note, it's a slightly artifical, but you can `this[reactionTrackingRef] = React.createRef()`

guochengcheng04: > Note, it's a slightly artifical, but you can `this[reactionTrackingRefSymbol] = React.createRef()`

Yes, but I don't think it's necessary. When a function component uses ref, because it can't get the component instance, the class component can directly cache this to achieve the same effect

guochengcheng04: > Any reason not to import these utils from `mobx-react-lite`?

If you think this is more reasonable, I can modify it here. Initially, a separate file was created for decoupling between different packages

urugator: Yes, `mobx-react` already depends on `mobx-react-lite`, it more or less only adds class support.

guochengcheng04: i will do it

guochengcheng04:
@urugator all change have done, check please, thank you

guochengcheng04:
@urugator Can you help me approve and publish? If need me to do anything please tell me

guochengcheng04:
> @urugator Can you help me approve and publish? If need me to do anything please tell me

@mweststrate @kubk  hi, can we merge and publish to finish this pr?

PR #3516, opened 2022-09-05
Fix 3514: legacy array compat with Safari 9

### LINKED ISSUE #3514: Mobx6 array index assignment does not work in Safari
Hey mate,
In Safari 9.3.1, assigning an object to an array by index is not working for me. My code:
```javascript
constructor() {
    makeAutoObservable(this);
    this.list.forEach((i, idx) => this.updateList(i, idx));
 }

updateList = (item, idx) => {
    const newItem = new Item(this, item); // another store
    this.list[idx] = newItem;
    return newItem;
  }
```
After updateList, this.list[0] is undefined, but the others are good.

The mobx version is 6.6.1.

I find the similar issue https://github.com/mobxjs/mobx/issues/364#issuecomment-231151569, and it was fixed in the previous version. Perhaps it has been missed in mobx6.

urugator:
Reintroduced workaround from mobx < 6.

Did not test as I don't have a means to do so. 

Fixes #3514

### DIFF
--- a/.changeset/fresh-suns-listen.md
+++ b/.changeset/fresh-suns-listen.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+fix regression #3514: LegacyObservableArray compat with Safari 9.\*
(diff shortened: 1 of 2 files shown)


mweststrate (approved):
I think it was part of just a clean up, assuming we wouldn't run anymore on that old Safaris

PR #3542, opened 2022-10-16
docs(collection-utilities): Add missing import

tony:
Missing import for `autorun` in the example on https://mobx.js.org/collection-utilities.html


urugator:
Thank you!

PR #3551, opened 2022-10-24
feature: abort when() using AbortSignal

deadbeef84:
### Code change checklist

Initially I tried adding it to the `_when` function to support the callback-based `when()`, but I wasn't sure how to properly clean up the event listener in the disposer so I went for this approach instead.

Also not sure why it requires @types/node 18, according to node docs AbortSignal was added in node 15 / 14.18.

### DIFF
--- a/.changeset/popular-comics-tap.md
+++ b/.changeset/popular-comics-tap.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
+Added new option `signal` to `when()`, to support abortion using an AbortSignal / AbortController.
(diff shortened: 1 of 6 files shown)


kubk (approved):

urugator:
Can you add a changeset (`yarn changeset`) and update docs?

deadbeef84:
> Can you add a changeset (`yarn changeset`) and update docs?

Done!

urugator:
Thank you!

rluvaton:
Hey, why did this remove? I don't see it in the code anymore

PR #3553, opened 2022-10-27
Add missing webfont weight and style for getting started page

chadluo:
Documentation change only.

When including the webfont from Google Fonts, only regular roman was added, and italics/bolds are fallbacking to unexpected styles. This change updates the webfont url to include roman, italics, roman bold and italics bold.

Before:
mobx js org_getting-started

After:
mobx js org_getting-started (1)


urugator:
LGTM. Thank you!

PR #3562, opened 2022-11-03
doc: fix flowResult example

jimmyhchan:
I was confused by the return type of the unboxed Promise returned by `flowResult`. It seems the documentation is incomplete since the generator is not returning anything.

https://codesandbox.io/s/vigorous-paper-zp79z5?file=/src/reduced.ts
https://codesandbox.io/s/vigorous-paper-zp79z5?file=/src/store.ts

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -360,8 +360,6 @@ class Store {
 ```javascript
 import { flow, makeAutoObservable, flowResult } from "mobx"
 
-type ProjectList = string[]
-
 class Store {
     githubProjects = []
     state = "pending"
@@ -378,7 +376,7 @@ class Store {
         this.state = "pending"
         try {
             // Yield instead of await.
-            const projects: ProjectList = yield fetchGithubProjectsSomehow()
+            const projects = yield fetchGithubProjectsSomehow()
             const filteredProjects = somePreprocessing(projects)
             this.state = "done"
             this.githubProjects = filteredProjects
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -360,6 +360,8 @@ class Store {
 ```javascript
 import { flow, makeAutoObservable, flowResult } from "mobx"
 
+type ProjectList = string[]
+
 class Store {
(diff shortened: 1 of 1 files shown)


kubk:
Hi @jimmyhchan , thank you for your PR.

Looking at documentation and for me it doesn't seem like this code snippet uses typescript at all: https://mobx.js.org/actions.html#asynchronous-actions

All of the examples use plain JS.

Regarding the return statement - it's also not needed since method `*fetchProjects` fetches data from API and stores the result in the class property `githubProjects`. But I agree with you it's not clear how to use `flow` with TS considering that TS coerces every generator result to `any`. There is a workaround to get the type inference working: 

It's not merged yet, but feel free to try it and share your results: https://github.com/mobxjs/mobx/discussions/3195

jimmyhchan:
Thanks for the feedback. I agree it's odd to introduce typescript when all the other examples are JS. The need to coerce to get the right type is annoying and i see how the workaround can be useful but i'm less concerned about that since that's just Typescript~isms and the warning `'yield' expression implicitly results in an 'any' type because its containing generator lacks a return-type annotation.` is actionable enough.  Happy to remove that entirely.

IMO, the return statement is the main issue since it seems necessary given the line 
```
const projects = await flowResult(store.fetchProjects())
```
projects  is typed `void` without the return. 


It seems the docs should be updated to `return project`  or  use `store. githubProjects` after the await.
```
const store = new Store()
await flowResult(store.fetchProjects())
// store.githubProjects is available at this point
store. githubProjects.forEach(doRender)
```

kubk:
@jimmyhchan You are right, the return statement is missing, thanks. I haven't noticed that line in the end of the snippet. Let's leave return but remove TS type addition since all the examples are in plain JS.

jimmyhchan:
Updated!  Thanks @kubk

kubk (approved):

kubk:
@jimmyhchan Thank you 👍

PR #3564, opened 2022-11-04
chore: remove the unused variables

Maidang1:
### Code change checklist


kubk:
I confirm that these variables aren't used. Thanks!

Issue #3532, opened 2022-09-23
Error "Type 'S' does not satisfy the constraint 'IValueMap'" with TypeScript 4.8 in mobx-react


DanielSWolf:
**Intended outcome:**

`mobx-react` 6.3.1 compiles with TypeScript 4.8

**Actual outcome:**

When using `mobx-react` 6.3.1 in a project that uses TypeScript 4.8, I get the following compiler error:

```text
node_modules/mobx-react/dist/inject.d.ts:5:63 - error TS2344: Type 'S' does not satisfy the constraint 'IValueMap'.

5 export declare function inject<S, P, I, C>(fn: IStoresToProps<S, P, I, C>): <T extends IReactComponent>(target: T) => T & IWrappedComponent<P>;
```

This error occurs because TypeScript 4.8 is stricter than previous versions regarding unconstrained type parameters (see the [blog entry announcing TypeScript 4.8](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#unconstrained-generics-no-longer-assignable-to)).

The fix is to change `function inject<S, P, I, C>` to `function inject<S extends IValueMap, P extends IValueMap, I extends IValueMap, C extends IValueMap>`.

**How to reproduce the issue:**

Use `mobx-react` 6.3.1 in any project that uses TypeScript 4.8.

**Versions**

DanielSWolf:
I'm not using Inject in any way. It just happens to be part of `mobx-react/dist/inject.d.ts`. So whenever I use `mobx-react` in a TypeScript 4.8 project, TypeScript tries to compile the types in this file and fails.

Codex-:
Also running into this issue, I pulled the mobx repository and was able to reproduce it by bumping typescript and just running a build:
```ascii
src/inject.ts:84:24 - error TS2344: Type 'S' does not satisfy the constraint 'IValueMap'.

84     fn: IStoresToProps<S, P, I, C>
                          ~

  src/inject.ts:83:24
    83 export function inject<S, P, I, C>(
                              ~
    This type parameter might need an `extends IValueMap` constraint.
```

ClaysonIO:
I'm finding the same issue when using typescript `4.6.4`, `4.7.2`, and `4.8.4`. My short term solution is using `skipLibCheck: true` in `tsconfig.json`, to ignore the error.

DanielSWolf:
[...]

PR #3565, opened 2022-11-04
Make mobx-react compatible with TS 4.8+

kubk:
Resolves: https://github.com/mobxjs/mobx/issues/3532

As @DanielSWolf wrote (thanks for that!): "This error occurs because TypeScript 4.8 is stricter than previous versions regarding unconstrained type parameters (see the [blog entry announcing TypeScript 4.8](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#unconstrained-generics-no-longer-assignable-to))." 

I confirm that adding constraint to the generic fixes the issue. Now the generic constraints are identical to `IStoresToProps`: https://github.com/mobxjs/mobx/blob/a39ce7fbe19f92348470ef71ad60555c37a5d5e9/packages/mobx-react/src/types/IStoresToProps.ts#L2+L6

I've chosen `minor` instead of `patch` because it's kinda BC break but at the same time it's a bugfix.

### DIFF
--- a/packages/mobx-react/src/inject.ts
+++ b/packages/mobx-react/src/inject.ts
@@ -6,7 +6,6 @@ import { IReactComponent } from "./types/IReactComponent"
 import { IValueMap } from "./types/IValueMap"
 import { IWrappedComponent } from "./types/IWrappedComponent"
 import { IStoresToProps } from "./types/IStoresToProps"
-import { IKeyValueMap } from "mobx"
 
 /**
  * Store Injection
@@ -81,7 +80,7 @@ export function inject(
 ): <T extends IReactComponent<any>>(
     target: T
 ) => T & (T extends IReactComponent<infer P> ? IWrappedComponent<P> : never)
-export function inject<S extends IKeyValueMap = {}, P extends IKeyValueMap = {}, I extends IKeyValueMap = {}, C extends IKeyValueMap = {}>(
+export function inject<S extends IValueMap = {}, P extends IValueMap = {}, I extends IValueMap = {}, C extends IValueMap = {}>(
     fn: IStoresToProps<S, P, I, C>
 ): <T extends IReactComponent>(target: T) => T & IWrappedComponent<P>
 
--- a/packages/mobx-react/src/inject.ts
+++ b/packages/mobx-react/src/inject.ts
@@ -6,6 +6,7 @@ import { IReactComponent } from "./types/IReactComponent"
 import { IValueMap } from "./types/IValueMap"
 import { IWrappedComponent } from "./types/IWrappedComponent"
 import { IStoresToProps } from "./types/IStoresToProps"
+import { IKeyValueMap } from "mobx"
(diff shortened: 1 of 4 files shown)


packages/mobx-react/src/inject.ts
```diff
@@ -80,7 +81,7 @@ export function inject(
 ): <T extends IReactComponent<any>>(
     target: T
 ) => T & (T extends IReactComponent<infer P> ? IWrappedComponent<P> : never)
-export function inject<S, P, I, C>(
+export function inject<S extends IKeyValueMap = {}, P extends IKeyValueMap = {}, I extends IKeyValueMap = {}, C extends IKeyValueMap = {}>(
```
urugator, line 84: Why not 
https://github.com/mobxjs/mobx/blob/51a1338fcb5afc5a9bae5fff4f906e5d6f27592e/packages/mobx-react/src/types/IValueMap.ts#L1
as used here
https://github.com/mobxjs/mobx/blob/51a1338fcb5afc5a9bae5fff4f906e5d6f27592e/packages/mobx-react/src/types/IStoresToProps.ts#L1-L7
?

DanielSWolf: I don't know much about the internals of MobX, but using `IValueMap` from `mobx-react` rather than `IKeyValueMap` from `mobx` does seem more straight-forward to me, too.

Technically, they're identical, so both solutions should work. But the reason for the error is that the generic type arguments need to be compatible with those in `IStoresToProps`. And that interface happens to use `IValueMap`, not `IKeyValueMap`.

kubk: Thanks @urugator @DanielSWolf 

Yes, it makes sense to use `IVavalueMap` and I've changed it. The issue is that I got failing tests on CI but I can't reproduce it locally. That's why I marked the PR as Draft. I'll get back to the issue once find time 👍

DanielSWolf:
Regarding the failing tests: It seems to me that these tests have always been broken. TypeScript just didn't realize it until you fixed the types.

I've only skimmed the problems, but here are my best guesses:

* [This line](https://github.com/mobxjs/mobx/blob/main/packages/mobx-react/__tests__/observer.test.tsx#L329) should probably read `inject(() => ({}))(` (note the extra parentheses).
* [This line](https://github.com/mobxjs/mobx/blob/main/packages/mobx-react/__tests__/context.test.tsx#L74) needs to explicitly specify the arrow function signature. This might work: `const MainInjected = inject(({ store }: { store: { thing: number } }) => ({ thing: store.thing }))(Main)`

kubk:
@DanielSWolf Thank you for the suggestions 👍  The issue with explicit signature is that we create a BC break. It looks like previously TS could infer the type correctly. But since `inject` is deprecated we can ask people to avoid using it completely in case of issues.

kubk:
@urugator The PR is ready to review 🙏

PR #3566, opened 2022-11-06
Make return value of reportObserved match invoke of onBecomeObserved

### LINKED ISSUE #3563: Return value of Atom.reportObserved is inconsistent with whether onBecomeObserved is called
**Intended outcome:**

`onBecomeObserved` should always be invoked when `reportObserved` returns `true`, or the other way around, `reportObserved` should only return `true` when `onBecomeObserved` has been run.

**Actual outcome:**

In some cases, `reportObserved` returns `true` but `onBecomeObserved` is never run, which breaks the way it is relied on, and makes atom a footgun to use.

This actually breaks the example shown for atom in [the document](https://mobx.js.org/custom-observables.html) that if it can be returning wrong time.

See this example: https://codesandbox.io/s/magical-jang-zcplf2?file=/src/index.ts

```ts
import * as mobx from "mobx";

const atom = mobx.createAtom(
  "test",
  () => console.log("become observed"),
  () => console.log("become unobserved")
);

const computed1 = mobx.computed(() => {
  console.log("computed observed", atom.reportObserved());
  return 0;
});
computed1.get();
const computed2 = mobx.computed(() => computed1.get());
computed2.get();
```

The console log is:
```
computed observed false
computed observed true
```
and the `onBecomeObserved` is never called.

On MobX 5, however, the console log would be
```
computed observed false
become observed
computed observed true
become unobserved
```
which is not ideal, but fine.

**Versions**

It is a change in MobX 6, more specifically I believe it is a regression from 5536533ed26d6bd8c7989b49990994daa3831009.

upsuper:
This fixes #3563.

### Code change checklist

### DIFF
--- a/.changeset/lemon-moles-clap.md
+++ b/.changeset/lemon-moles-clap.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Make return value of reportObserved match invoke of onBecomeObserved
--- a/packages/mobx/__tests__/v5/base/observables.js
+++ b/packages/mobx/__tests__/v5/base/observables.js
@@ -1400,6 +1400,49 @@ test("atom events #427", () => {
     expect(stop).toBe(2)
 })
 
+test("#3563 reportObserved in batch", () => {
+    let start = 0
+    let stop = 0
+    let computed = 0
+    let observed = 0
+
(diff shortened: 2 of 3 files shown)


urugator:
LGTM, thank you!

PR #3577, opened 2022-11-16
Remove outdated information from mobx-react README

kubk:
Since we only support mobx-react 7+ I've removed confusing information from the README. I've removed the comparison table completely because there is nothing to choose from. Only the latest version is maintained.


urugator:
Thank you for the PR. I think the version table can still prove useful/relevant, there seems to be quite a few users still using mobx 5 or even 4. Perhaps just change the title to "Version overview" and add something like:
"Only the latest version is actively maintained. If you're missing a fix or a feature in older versions, consider upgrading or using [patch-package](https://www.npmjs.com/package/patch-package)"

kubk:
@urugator Agreed, done 👍

kubk:
Hi @urugator 

Since it's just a README change I think we can merge it without waiting for CI. Thanks

urugator (approved):

urugator:
> Since it's just a README change I think we can merge it without waiting for CI. Thanks

I don't have permissions to merge unless it's all green.

mweststrate:
merge overriden, I see if I can uninstall netlify, I suspended it, not sure why taht isn't enough

PR #3583, opened 2022-11-22
Fix #3582, AbortSignal leaks @types/node

### LINKED ISSUE #3582: mobx 6.7.0 leaks @types/node global typings
**Intended outcome:** When using mobx in a TypeScript in a frontend-only project, it should be possible to use the DOM typings for globals such as `setTimeout` (which should return a `number`).

**Actual outcome:** The globals get their typings from `@types/node`, because `mobx/dist/api/when.d.ts` contains a spurious

```typescript
/// <reference types="node" />
```

since https://github.com/mobxjs/mobx/commit/223e3688631528a327c79d39e2f497c6e1506165. Unfortunately, global typings referenced by imported files are automatically propagated to importing packages: https://github.com/microsoft/TypeScript/issues/22331#issuecomment-613816908

In particular, `setTimeout` now return `NodeJS.Timeout` even if `"node"` is _omitted_ from `compilerOptions.types` in `tsconfig.json`. Globals not present in DOM, such as `process`, are also visible, which is particularly dangerous for type safety.

Create a TypeScript project with a `tsconfig.json` containing

```json
{
    "compilerOptions": {
        "libs": []
    }
}
```

Observe that if mobx 6.6.2 is imported in the project, globals like `setTimeout` have their types as specified in DOM (e.g., `setTimeout` returns `number`): https://codesandbox.io/s/currying-worker-yj8gt6?file=/index.tsx

If mobx 6.7.0 is imported, the globals have their types as in `@types/node`, because mobx references the node global typings: https://codesandbox.io/s/sleepy-bird-dh354u?file=/index.tsx

**Versions**

* typescript 4.9.3
* mobx 6.7.0

urugator:
Fixes #3582

Not sure if optimal, but I used the same solution as the one in `axios` package:
https://github.com/axios/axios/blob/d032edda08948f3f613ff34f32af49cad3aa74db/index.d.ts#L215-L220
https://github.com/axios/axios/pull/4229
https://github.com/axios/axios/pull/5021 (why most of the interface is optional)

Another apporach I found is forward declaration:
`
declare global {
  export interface AbortSignal {}
}
`
as eg in: https://github.com/aws/aws-sdk-js-v3/blob/826807a8e0435580cac0d209ca8199a6231a9f96/packages/types/src/serde.ts#L59-L70
Dunno if applicable.


kubk (approved):

packages/mobx/src/api/when.ts
```diff
@@ -9,11 +9,19 @@ import {
     allowStateChanges
 } from "../internal"
 
+// https://github.com/mobxjs/mobx/issues/3582
+export interface GenericAbortSignal {
```
mweststrate, line 13: NIT: I don't think this type necessarily needs exporting?

mweststrate (approved):

Issue #3438, opened 2022-06-30
DisplayName not being forwarded when wrapped in 


jtwigg:
When I receive a printed call stack from React, the components wrapped in `observer` show up as `observerComponent`

```
    at observerComponent (http://localhost:3000/static/js/bundle.js:119423:69)
    at div
    at observerComponent (http://localhost:3000/static/js/bundle.js:119423:69)
    at PrimaryWindow (http://localhost:3000/static/js/bundle.js:4346:5)
```

and in the callstack in the debugger shows up as `anonymous`
```
<anonymous> (/Users/johntwigg/development/crypto/loopNFT/ui/src/components/Cinema/CinemaMain.tsx:35)
<anonymous> (/Users/johntwigg/development/crypto/loopNFT/ui/node_modules/mobx-react-lite/src/observer.ts:104)
<anonymous> (/Users/johntwigg/development/crypto/loopNFT/ui/node_modules/mobx-react-lite/src/useObserver.ts:115)
```

**Intended outcome:**
Ideally, the underlying React Name would be maintained. `CinemaMain` in this case.

**Actual outcome:**
`anonymous` and `observerComponent`

Looking at issue #3422 doesn't seem to resolve or change the behavior for me. 
For example
```
const FocusLane : React.FC = observer(() => {...})
FocusLane.displayName = "FocusLane"
```
has no impact.

Any ideas? I'm open to workarounds.

**Versions**
```
➜  ui git:(mobx) ✗ npm list --depth 2 | grep mobx
├─┬ mobx-react-lite@3.4.0
│ ├── mobx@6.6.1
```

mweststrate:
Did you search for existing issues first?

jtwigg:
> Did you search for existing issues first?

Yes. I referenced the one here 

```
Looking at issue https://github.com/mobxjs/mobx/issues/3422 doesn't seem to resolve or change the behavior for me.
```
It claims to be fixed.

urugator:
Have you tried it without `observer`? I don't think `displayName` has any effect on callstacks, it's only relevant for dev tools or error messages.

mdeem:
I can confirm that setting displayName on the return value of observe as imported from mobx-react-lite 3.4.0 doesn't work with the lasted react memo code. This code is now checking for name as well as displayName, and no longer forwards displayName changes if name is set, which is the case for observerComponent.

[...]

PR #3590, opened 2022-12-17
Better React 18 support

### LINKED ISSUE #3492: An exception is thrown when using Mobx6 and react18 in development mode

urugator:
### mobx
It now keeps track of a global state version, which updates with each mutation.

### mobx-react-lite
It now uses `useSyncExternalStore`, which should get rid of tearing (you have to update mobx, otherwise it should behave as previously).
Replaced reaction tracking utils with `UniversalFinalizationRegistry`. It works the same way, but I found it hard to orient myself in the original impl, so I rewrote it completely, hopefully for the better. It's also easier to reuse for class components.
Improved displayName/name handling as discussed in #3438.

### mobx-react (class component)
Reactions of uncommited components are now correctly disposed. (fixes #3492)
Reactions don't notify uncommited components, avoiding the warning. (fixes #3492)
Removed symbol "polyfill" and replaced with actual Symbols.
Removed `this.render` replacement detection + warning. `this.render` is no longer configurable/writable (possibly BC *).
Reaction is no longer exposed as `component[$mobx]`  (possibly BC *)
Component instance is no longer exposed as `component[$mobx]["reactComponent"]` (possibly BC *)
Deprecated `disposeOnUnmount`, it's not compatible with remounting.
Refactored code.
Fixed tests.

(*) BC for non-idiomatic usage or when depending on low level (private?) API.

I will update changeset once the changes settles.


packages/mobx-react/__tests__/__snapshots__/observer.test.tsx.snap
```diff
@@ -35,23 +37,7 @@ exports[`Redeclaring an existing observer component as an observer should log a
 }
 `;
 
-exports[`SSR works #3448 1`] = `
```
urugator, line 38: Was not supposed to warn. Warning was caused by calling `disableStaticRendering` before `unmount` in the test. Therefore `componentWillUnmount` used logic for non-static rendering.

packages/mobx-react/__tests__/inject.test.tsx
```diff
@@ -102,7 +102,7 @@ describe("inject based context", () => {
         expect(C.displayName).toBe("inject(ComponentC)")
     })
 
-    test.only("shouldn't change original displayName of component that uses forwardRef", () => {
```
urugator, line 105: Was left in here for quite some time by mistake.

packages/mobx-react/src/observerClass.ts
```diff
@@ -50,142 +90,156 @@ export function makeClassComponentObserver(
...
-    /**
-     * If props are shallowly modified, react will render anyway,
-     * so atom.reportChanged() should not result in yet another re-render
-     */
-    setHiddenProp(this, skipRenderKey, false)
-    /**
-     * forceUpdate will re-assign this.props. We don't want that to cause a loop,
-     * so detect these changes
-     */
-    setHiddenProp(this, isForcingUpdateKey, false)
```
urugator, line 127: Merged these together into `isUpdating` as they basically handle the same thing:
One is set by reaction to prevent reportChanged on `props` setter,
second is set by `props` setter to prevent reaction.
So both says "I am already updating, do not cause another update"

packages/mobx-react-lite/src/utils/UniversalFinalizationRegistry.ts
```diff
@@ -0,0 +1,62 @@
+export declare class FinalizationRegistryType<T> {
```
mweststrate, line 1: Would it be possible to do the FinalizationRegistery refactoring (makes lot of sense) in a separate PR? Then we could first rollout that as a patch change as, as we don't expect changes, and then separately roll out the change for React 18, to reduce the size of the change we have to debug if something goes awry? But if the changes are too intertwined I can live with this as well.

packages/mobx-react-lite/src/utils/UniversalFinalizationRegistry.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private scheduleSweep() {
+        if (this.sweepTimeout === undefined) {
+            this.sweepTimeout = setTimeout(this.sweep, REGISTRY_SWEEP_INTERVAL)
+        }
+    }
+}
+
+export const UniversalFinalizationRegistry = FinalizationRegistry ?? TimerBasedFinalizationRegistry
```
mweststrate, line 62: Neat setup and implementation!

packages/mobx-react/src/disposeOnUnmount.ts
```diff
@@ -17,12 +20,22 @@ function runDisposersOnWillUnmount() {
...
+/**
+ * @deprecated
+ */
 export function disposeOnUnmount<TF extends Disposer | Array<Disposer>>(
     target: React.Component<any, any>,
     fn: TF
 ): TF
 
+/**
+ * @deprecated
```
mweststrate, line 37: NIT: Let's clarify why it is deprecated, as that shows up in VSCode suggestions

packages/mobx/src/core/atom.ts
```diff
@@ -66,6 +67,9 @@ export class Atom implements IAtom {
     public reportChanged() {
         startBatch()
         propagateChanged(this)
+        // We could update state version only at the end of batch,
+        // but we would still have to switch some global flag here to signal a change.
+        globalState.stateVersion = Symbol()
```
mweststrate, line 72: Why not increase a global number? Symbol() is probably a lot more expensive (both mem and cpu), given that this is probably hit very often

packages/mobx-react-lite/package.json
```diff
@@ -36,7 +36,9 @@
         "url": "https://github.com/mobxjs/mobx/issues"
     },
     "homepage": "https://mobx.js.org",
-    "dependencies": {},
+    "dependencies": {
+        "use-sync-external-store": "^1.2.0"
+    },
     "peerDependencies": {
         "mobx": "^6.1.0",
```
mweststrate, line 43: peer dependency probably has to be bumped to make sure globalState.stateVersion is available?

packages/mobx/__tests__/v5/base/observables.js
```diff
@@ -2360,3 +2360,30 @@ describe("`requiresObservable` takes precedence over global `reactionRequiresObs
         expect(consoleWarnSpy).not.toHaveBeenCalled()
     })
 })
+
+test("state version updates correctly", () => {
+    // This test was designed around the idea of updating version only at the end of batch,
+    // which is NOT an implementation we've settled on, but the test is still valid.
```
mweststrate, line 2366: Sounds correct to me though :)

mweststrate (commented):
Looking good so far! I didn't review everything yet (still have to do the new observer implementations), but submitting this review already so that you can see the comments, as coming days I probably won't be able to review further with all Christmas activities :).

Merry Christmas and happy new year!

packages/mobx/src/core/atom.ts
ericmasiello: What makes you say `Symbol()` is more expensive?

mweststrate:
Every Symbol is unique, so at least requires an allocation and GC.

packages/mobx/src/core/atom.ts
urugator: > Why not increase a global number?

Becomes non-unique at some point. Dunno how much of an issue it would be in practice. For reference https://github.com/mobxjs/mobx/pull/2982

packages/mobx-react-lite/package.json
urugator: The current impl should be BC, see
https://github.com/mobxjs/mobx/pull/3590/files#diff-6b93715e19eb9119588dbe2af609a132062e61484d0d3de41b9e32e2d4a4ad5cR13-R17
https://github.com/mobxjs/mobx/pull/3590/files#diff-6b93715e19eb9119588dbe2af609a132062e61484d0d3de41b9e32e2d4a4ad5cR31-R34
https://github.com/mobxjs/mobx/pull/3590/files#diff-6b93715e19eb9119588dbe2af609a132062e61484d0d3de41b9e32e2d4a4ad5cR76-R78

but bumping dep is a way too.

packages/mobx/__tests__/v5/base/observables.js
urugator: I am not sure what you mean. Just to provide some background. I've been looking for a common place where I could update state version. `endBatch` (when nesting level gets back to zero) seemed like a good candidate, because it's actually called even for individual mutations, but I was worried about some edge cases, like autorun invoking itself. So I wrote this test and gave it a go. However I quickly realized `endBatch` is also often called without any actual mutation, therefore I threw the idea away and moved the version update to `reportChanged`, but I kept the test as is. I've left the comment to clarify why the test looks this way.

packages/mobx-react/src/disposeOnUnmount.ts
urugator: Done, but message probably isn't entirely accurate. I am no sure which react version started to perform re-mounting. Maybe it's already done by 17 only in StrictMode, but I don't remember.

packages/mobx-react-lite/src/utils/UniversalFinalizationRegistry.ts
urugator: PR is here https://github.com/mobxjs/mobx/pull/3598. However I've also changed `useObserver` impl a bit.

packages/mobx-react-lite/package.json
```diff
@@ -36,7 +36,9 @@
         "url": "https://github.com/mobxjs/mobx/issues"
     },
     "homepage": "https://mobx.js.org",
-    "dependencies": {},
+    "dependencies": {
+        "use-sync-external-store": "^1.2.0"
```
ericmasiello, line 40: Isn't this built into React 18? I'm not sure if there's a way to handle this, but if the user installs React 18, ideally we wouldn't require they also install this, no?

urugator: We would have to maintaing seperate mobx-react versions (for >= 18 and < 18), which isn't worth it.
Note the shim prefers React's impl if available. Dunno if it's tree shakable, but it's not very big.

ericmasiello: I was hoping something in https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta would solve this problem but I'm not seeing anything unfortunately

packages/mobx-react/__tests__/finalizationRegistry.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+afterEach(cleanup)
+
+function sleep(time: number) {
+    return new Promise<void>(res => {
+        setTimeout(res, time)
+    })
+}
+
+// TODO dunno why it's not available
```
tonyraoul, line 17: project was missing a configuration file, please take a look https://github.com/mobxjs/mobx/pull/3620
or simply add the file manually.

packages/mobx-react/src/observerClass.ts
```diff
@@ -1,44 +1,84 @@
-import { PureComponent, Component } from "react"
+import { PureComponent, Component, ComponentClass, ClassAttributes } from "react"
 import {
     createAtom,
     _allowStateChanges,
     Reaction,
-    $mobx,
+    //$mobx, // TODO
```
mweststrate, line 6: // TODO reminder

mweststrate (commented):
Looking great @urugator! Some small comments around publishing, but this looks really solid. Thanks for the extensive comments in the code!

mweststrate (commented):
Looking great @urugator! Some small comments around publishing, but this looks really solid. Thanks for the extensive comments in the code!

packages/mobx-react-lite/package.json
urugator: If this is gonna be a major should it still be focused on React 18 support with otherwise minimal changes, or should this be an opportunity to introduce larger changes? (removing deprecated APIs, hooks, inject, options, cleanups, etc ...)

I assume the former.

mweststrate (approved):

packages/mobx-react-lite/package.json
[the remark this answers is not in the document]
urugator: Hm, I think you're correct, did not occured to me.

packages/mobx/src/core/atom.ts
urugator: Changed to number for now. The only thing I am concerned is that if this should ever become a problem, nobody would probably ever notice anyway... not during development at least.

ericmasiello:
I’m curious, how would I observe the “better React 18” support? Can you provide some reproduction steps? For example, I’d love to see something on the order of:
1. Follow these steps with the current version of MobX and React 18 and you can clearly observe *these* negative outputs
2. Repeat the above steps and now you can observe in this branch, those problems are solved.

urugator:
@ericmasiello Besides mentioned fixed issues for which we do have tests, you should see improvements if you run it against:
https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering
see this comment https://github.com/mobxjs/mobx/issues/2526#issuecomment-1248172620
Personally did not try and I don't have any setup for this, so at this point there is no easily reproducible evidence and you're welcome to change that :)

packages/mobx-react/package.json
```diff
@@ -36,10 +36,10 @@
     },
     "homepage": "https://mobx.js.org",
     "dependencies": {
-        "mobx-react-lite": "^3.4.0"
+        "mobx-react-lite": "^3.4.3"
     },
     "peerDependencies": {
-        "mobx": "^6.1.0",
+        "mobx": "^6.9.0",
```
urugator, line 42: A guess - I don't know the actual version it's gonna be released with. Dunno what's the correct way to handle it.

packages/mobx-react/package.json
```diff
@@ -51,8 +51,9 @@
         }
     },
     "devDependencies": {
-        "mobx": "^6.7.0",
-        "mobx-react-lite": "^3.4.0"
+        "mobx": "^6.8.0",
+        "mobx-react-lite": "^3.4.3",
```
urugator, line 55: Should become 6.9?.0 and 4.0.0 at the time of release. Again not sure how to handle that.

packages/mobx-react/package.json
```diff
@@ -36,10 +36,10 @@
     },
     "homepage": "https://mobx.js.org",
     "dependencies": {
-        "mobx-react-lite": "^3.4.0"
+        "mobx-react-lite": "^3.4.3"
```
urugator, line 39: Should become ^4.0.0 at the time of release. Again not sure how to handle that.

urugator:
@mweststrate Please take a look at https://github.com/mobxjs/mobx/pull/3649 and let me know if I should merge it to master first and incorporate the change to this PR.

mweststrate:
Sorry for the delay! I'll dive into it this weekend

mweststrate (approved):
Looking good! Imho it's okay to merge #3649 after this one.

packages/mobx-react-lite/package.json
mweststrate: I think we can do the former; decorators are also around the corner, an I expect after that is a better moment to do a bigger clean up (e.g. legacy decorator support etc)

mweststrate (approved):
AWESOME STUFF!

ericmasiello:
I pulled this repo locally and configured it to use 

I see *mostly* progress with 1 regression.  In the output below, "mobx" is `mobx` 6.8.0 and `mobx-react-lite` 3.4.0. "mobx-beta" uses the latest from this branch.

```
  mobx
    With useTransition
      Level 1
        ✓ No tearing finally on update (4571 ms)
        ✓ No tearing finally on mount (5472 ms)
      Level 2
        ✕ No tearing temporarily on update (8641 ms)
        ✕ No tearing temporarily on mount (5460 ms)
      Level 3
        ✓ Can interrupt render (time slicing) (3586 ms)
        ✕ Can branch state (wip state) (2952 ms)
    With useDeferredValue
      Level 1
        ✓ No tearing finally on update (9622 ms)
        ✓ No tearing finally on mount (6486 ms)
      Level 2
        ✓ No tearing temporarily on update (14596 ms)
        ✕ No tearing temporarily on mount (5468 ms)
  mobx-beta
    With useTransition
      Level 1
        ✓ No tearing finally on update (7975 ms)
        ✓ No tearing finally on mount (4555 ms)
      Level 2
        ✓ No tearing temporarily on update (12924 ms)
        ✓ No tearing temporarily on mount (4514 ms)
      Level 3
        ✕ Can interrupt render (time slicing) (7911 ms)
        ✕ Can branch state (wip state) (6685 ms)
    With useDeferredValue
      Level 1
        ✓ No tearing finally on update (9647 ms)
        ✓ No tearing finally on mount (4535 ms)
      Level 2
        ✓ No tearing temporarily on update (14596 ms)
        ✓ No tearing temporarily on mount (4494 ms)
```

ericmasiello:
> @ericmasiello Besides mentioned fixed issues for which we do have tests, you should see improvements if you run it against: https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering see this comment [#2526 (comment)](https://github.com/mobxjs/mobx/issues/2526#issuecomment-1248172620) Personally I did not try and I don't have any setup for this, so at this point there is no easily reproducible evidence and you're welcome to change that :)

FYI, I have an open PR to add the latest release of mobx/mobx-react-lite to this repo https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-rendering/pull/75

zhantx:
Hi @urugator 
can you please elaborate why disposeOnUnmount is deprecated? 
I tested it in react 18, it seems the function version is working fine if it's in WillMount (so the reaction is recreated when remount)
Is it only the decorator version problematic in react 18? (I tested it too, it seems can't dispose the reaction properly)

upsuper:
Just want to mention that `disposeOnUnmount` is very useful for us, and it's widely used in our codebase. I'd also note that this utility doesn't encourage, nor does lack of it discourage the pattern. There exists lots of old code that simply collects disposers in `componentDidMount` and invokes them in `componentWillUnmount`. `disposeOnUnmount` helps simplify the code and ensure that they are disposed properly.

upsuper:
I think it's fine that we maintain our own version of this utility in our codebase. We use it both for reaction/autorun and for other stuff following a similar pattern.

Mainly we want to understand the reason that it's deprecated, so that we know whether our use of it has any fundamental issue, in which case our own version of this utility would suffer the same problem as well.

It's good to know that it isn't that the utility really has anything incompatible with React.

urugator:
A potential source of some trouble is the need to patch `componenWillUnmount`, which we already patch by https://github.com/mobxjs/mobx/blob/main/packages/mobx-react/src/utils/utils.ts#L125
The complexity involved is beyond my current knowledge.

PR #3598, opened 2022-12-31
`useObserver` registry refactor

urugator:
As discussed here https://github.com/mobxjs/mobx/pull/3590#discussion_r1056788371


mweststrate (approved):
Looking good, thanks for putting that in a separate PR!

packages/mobx-react-lite/__tests__/strictAndConcurrentModeUsingFinalizationRegistry.test.tsx
```diff
@@ -1,21 +1,20 @@
...
-
-// @ts-ignore
 import gc from "expose-gc/function"
+import { observerFinalizationRegistry } from "../src/utils/observerFinalizationRegistry"
+
+if (typeof globalThis.FinalizationRegistry !== "function") {
+    throw new Error("This test must run with node >= 14")
+}
+
+expect(observerFinalizationRegistry).toBeInstanceOf(globalThis.FinalizationRegistry)
```
Bnaya, line 13: What does it do if it fails? it's not inside any `test`

packages/mobx-react-lite/__tests__/strictAndConcurrentModeUsingTimers.test.tsx
```diff
@@ -2,21 +2,22 @@ import "./utils/killFinalizationRegistry"
...
-    CLEANUP_LEAKED_REACTIONS_AFTER_MILLIS,
-    CLEANUP_TIMER_LOOP_MILLIS
-} from "../src/utils/reactionCleanupTrackingCommon"
+    REGISTRY_FINALIZE_AFTER,
+    REGISTRY_SWEEP_INTERVAL
+} from "../src/utils/UniversalFinalizationRegistry"
+import { observerFinalizationRegistry } from "../src/utils/observerFinalizationRegistry"
+import { TimerBasedFinalizationRegistry } from "../src/utils/UniversalFinalizationRegistry"
+
+expect(observerFinalizationRegistry).toBeInstanceOf(TimerBasedFinalizationRegistry)
```
Bnaya, line 13: Same as above

Bnaya (approved):
less code = It's better :D

packages/mobx-react-lite/__tests__/strictAndConcurrentModeUsingTimers.test.tsx
urugator: > What does it do if it fails? 

Throws?

> it's not inside any test

Does it matter?

It's mainly intented to assert environment (bug in the test), rather than testing that `observerFinalizationRegistry` resolves to correct type (bug in the actual tested code).
Arguably I could split these into two, assertion to check `globalThis` for (not) being `FinalizationRegistry` and then an actual `test` for `observerFinalizationRegistry`. However I don't see much value in doing so.

packages/mobx-react-lite/__tests__/strictAndConcurrentModeUsingFinalizationRegistry.test.tsx
```diff
@@ -1,21 +1,20 @@
...
-
 import { useObserver } from "../src/useObserver"
 import { sleep } from "./utils"
-import { FinalizationRegistry } from "../src/utils/FinalizationRegistryWrapper"
-
-// @ts-ignore
 import gc from "expose-gc/function"
+import { observerFinalizationRegistry } from "../src/utils/observerFinalizationRegistry"
+
+if (typeof globalThis.FinalizationRegistry !== "function") {
```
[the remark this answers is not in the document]
Bnaya: ... But then the error message won't be the same *think think*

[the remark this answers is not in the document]
urugator: Yes, I did want to preserve the message.

PR #3599, opened 2022-12-31
Add name to WHEN_TIMEOUT error if available

evelant:
It's useful to have a name attached to the WHEN_TIMEOUT error for debugging/tracing. Since `when` already takes a name as an option this PR just adds the name to the WHEN_TIMEOUT error if it is available.

### Code change checklist

-   [X ] Added/updated unit tests
-   [X ] Updated `/docs`. For new functionality, at least `API.md` should be updated
-   [X ] Verified that there is no significant performance drop (`yarn mobx test:performance`)


mweststrate (changes requested):
Sorry for the late follow up. I imagine `WHEN_TIMEOUT` is an error message that is programmatically checked against in catch handlers, so I think it'd be better to store the debug name in a separate property on the same error object. E.g `whenName` or something like that.

dartess:
This may break the logic associated with checking for a timeout error.

PR #3603, opened 2023-01-10
Remove old documentation mentions

kubk:
Related discussion: https://github.com/mobxjs/mobx/discussions/3602

I've searched for all the mentions of `https://mobx-react.js.org` in the codebase.

PR #3604, opened 2023-01-10
Replace Circle CI with GitHub CI

kubk:
There is a new `Build and check` check that includes all the checks from the Circle CI config file


.github/workflows/coveralls.yml
```diff
@@ -1,6 +1,6 @@
 name: Coveralls
 
-on: ["push", "pull_request"]
+on: ["push"]
```
kubk, line 3: This line caused coveralls to run twice on each push. Example: https://github.com/mobxjs/mobx/pull/3603

Screenshot 2023-01-10 at 11 50 34

It should be enough to run it once

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: yarn test -i
+
+      - name: Test size
+        run: yarn lerna run test:size
+
+      - name: Test flow
+        run: yarn mobx test:flow
+
+      - name: Test performance
+        run: yarn mobx test:performance
```
kubk, line 40: Initially I wanted to have multiple jobs (like build, test, test size) the same way it was implemented in CircleCI config. On CircleCI it was done using [persist_to_workspace](https://circleci.com/docs/configuration-reference/) option:

https://github.com/mobxjs/mobx/blob/42f8ac057ec70c508232339016cc7249123f0fd0/.circleci/config.yml#L25

There is an alternative way doing so using GitLab actions: https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow

But each time I added `actions/upload-artifact@v3` to the config the pipeline mysteriously disappeared from the GitHub pipeline list and I saw only outdated CircleCI: https://github.com/mobxjs/mobx/pull/3604/commits/b9b473df637212b5553cd540086bce66fd202e90

Screenshot 2023-01-10 at 11 55 32

So I decided to keep it flat and it worked out.

kubk:
The only thing that I didn't cover here is creating artifacts out of `test:performance` command:

https://github.com/mobxjs/mobx/blob/19de5da5c4e98e61b67e6176cc3a95c94d013bbd/.circleci/config.yml#L49+L52

Can someone describe why was it needed?

.github/workflows/coveralls.yml
iChenLei: If remove `pull_request` options, PR will not trigger CI.

urugator:
> Can someone describe why was it needed?

I would assume so you can investigate/compare the results. The test itself doesn't pose any limits, you have to check the results manually.

kubk:
@urugator Yes, it doesn't pose any limits. GitHub CI shows the whole output of the command so it looks like we don't need to create artifacts out of the performance result anymore:

Screenshot 2023-01-11 at 20 23 24

kubk:
@urugator It's not related to `actions/upload-artifact@v3`, I just don't understand how this artifact uploading in CircleCi supposed to work. Sure, lets wait for Michel's input 👍

mweststrate:
Oh wow, I never knew perf artifacts were saved, so I never looked at them :-). So I think the setup as is is fine with one job. The primary thing about perf tests is to make sure they don't break by a PR. I'm fine to just run those tests locally whenever we have a PR we fear might affect them. I'll disable the circle ci tests above as well to make sure they don't block anything anymore.

kubk:
@mweststrate Thanks, I see that Circle CI is no longer blocking the CI. May I merge this?

mweststrate (approved):

mweststrate:
Yes, LGTM! Thanks for picking up!

iChenLei (approved):

PR #3605, opened 2023-01-10
update makeObservable doc

T1mL3arn:
Though this should be clearly understood from the rest of the doc (it wasn't clear for me :sweat: ), I believe it should be stated explicitly.

### DIFF
--- a/docs/observable-state.md
+++ b/docs/observable-state.md
@@ -24,7 +24,7 @@ Usage:
 
 This function can be used to make _existing_ object properties observable. Any JavaScript object (including class instances) can be passed into `target`.
 Typically `makeObservable` is used in the constructor of a class, and its first argument is `this`.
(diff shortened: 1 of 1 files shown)


docs/observable-state.md
```diff
@@ -24,7 +24,7 @@ Usage:
 
 This function can be used to make _existing_ object properties observable. Any JavaScript object (including class instances) can be passed into `target`.
 Typically `makeObservable` is used in the constructor of a class, and its first argument is `this`.
-The `annotations` argument maps [annotations](#available-annotations) to each member. Note that when using [decorators](enabling-decorators.md), the `annotations` argument can be omitted.
+The `annotations` argument maps [annotations](#available-annotations) to each member. Only annotated members become observable. Note that when using [decorators](enabling-decorators.md), the `annotations` argument can be omitted.
```
urugator, line 27: Thank you for the PR. Since annotations are also used to introduce actions etc, could you just please change it to:
`Only annotated members are affected.`

PR #3608, opened 2023-01-12
Fix #3606, Do not expose deprecated observe_ and intercept_ methods.

emereum:
Fixes #3606

**Context**
The `observe` and `intercept` methods are no longer exposed on arrays, maps, and boxes as per [6.0.0 breaking changes](https://github.com/mobxjs/mobx/blob/main/packages/mobx/CHANGELOG.md#changes-that-might-affect-you) but they are still exposed as `observe_` and `intercept_` on `IObservableValue` and `IComputedValue`.

**Problem**
* It may seem inconsistent to expose these methods on some parts of the API but not others.
* An `IObservableValue` can not be assigned to an `IComputedValue` from MobX 6 onwards because of a type discriminator in the `observe_` event handler.

**Solution**
Do not expose `observe_` or `intercept_` on computed values and observable values.


kubk (approved):

kubk:
Thank you!

PR #3609, opened 2023-01-12
Fix #3517, Restore types for newValue and oldValue on IValueDidChange and IBoxDidChange.

emereum:
Fixes #3517

This PR restores the generic types for `newValue` and `oldValue` on `IValueDidChange` and `IBoxDidChange`, removing the `unknown` type.

I think it would be worthwhile to update the generic default from `T = any` to `T = unknown` in the next major version bump as discussed in #3517


kubk:
Nice addition, thank you 👍

PR #3620, opened 2023-02-01
Fix ide typescript compatability in __tests__ dir

tonyraoul:
### Description
While having a look at https://github.com/mobxjs/mobx/pull/3590 I found out that `WeakRef` was not defined although `ESNext` was being in use.
 Apparently VSCode and most LSP servers don't recognise `tsconfig.test.json` and the directory was explicitly excluded from the project root, only `tsconfig.json` is recognised.

This PR aims at fixing Developer Experience while working within `__tests__` for both `mobx-react` and `mobx-react-light` packages.

Hope I am helping.
Cheers.


urugator:
Thanks for looking into it. Isn't there the same in issue with `mobx` package? 
I wasn't able to get rid of those "decorators are experimental" warnings in tests, so this could somehow explain it...

tonyraoul:
@urugator would you like me to close this or is it still needed ?

urugator:
No please keep it open. The PR is very much appreciated, I (or someone else) will get to it eventually. There are just other issues I would like to settle first and I don't want to accidentaly complicate their resolution due to config changes or something. Sorry for the delay.

Issue #3618, opened 2023-01-30
`requiresReaction` in options of `computed` does not throw as indicated by the documentation


upsuper:
```js
import * as mobx from 'mobx';
const a = mobx.computed(() => console.log('b'), { requiresReaction: true });
mobx.onBecomeUnobserved(a, () => console.log('c'));
a.get();
```

The document [says](https://github.com/mobxjs/mobx/blob/41c7c5c0d229c2575f5dec8fa0ab6278f137f4a0/docs/computeds.md?plain=1#L232-L234):
> ### `requiresReaction`
>
> It is recommended to set this one to true on very expensive computed values. If you try to read its value outside of the reactive context, in which case it might not be cached, __it will cause the computed to throw__ instead of doing an expensive re-evalution.

**Intended outcome:**

It should throw, because it is not called within a reactive context.

**Actual outcome:**

It does not throw.

**How to reproduce the issue:**

Just run the code above.

**Versions**

MobX 6.

upsuper:
I noticed that this behavior was changed in #2998 possibly intentionally.

But as I mentioned in https://github.com/mobxjs/mobx/issues/2417#issuecomment-1341584660 previously, we in Canva find it hard to enforce restrictions if they don't actually throw, because developers don't look at console all the time when nothing goes wrong, and it is hard to track all the requirements across a large codebase. So we would prefer there to be a way to enforce the restrictions.

If it is intended to make `requiresReaction` on par with behavior of other `requires*` to warn, it would be better to introduce a new `enforcesReaction` to actually throw.

upsuper:
There is also another potential use case of `requiresReaction`: `computed` may sometimes be used for managing resources, which may be leaking, on demand. To prevent leaking, we may use `onBecomeUnobserved` to clean up. However, it doesn't work outside observed context, so one may want to enforce reaction for that reason.

An example as below:
```ts
let _blobUrl: string | undefined;
const blobUrl = mobx.computed(() => {
  // some check
  if (_blobUrl != null) {
    URL.revokeObjectURL(_blobUrl);
  }
  // generate a blob
  _blobUrl = URL.createObjectURL(blob);
  return _blobUrl;
});
mobx.onBecomeUnobserved(blobUrl, () => {
  if (_blobUrl != null) {
    URL.revokeObjectURL(_blobUrl);
    _blobUrl = undefined;
  }
});
```

[...]

PR #3634, opened 2023-02-24
Warning severity configuration

emereum:
This PR adds a new `warningSeverity` configuration option. This option allows consumers to promote specific warnings to exceptions. This solves the issue where some consumers have unique needs for error reporting, such as #3618, while still respecting MobX's error reporting strategy in the default case.

Example usage:

```javascript
configure({
    warningSeverity: {
        computedRequiresReaction: "throw",
        enforceActionsStrict: "throw",
        enforceActionsNonStrict: "throw",
        observableRequiresReaction: "throw",
        derivationWithoutDependencies: "throw"
    }
})
```

Warnings will default to `warn` which is in line with their current behaviour in `6.8.0`.

### DIFF
--- a/.changeset/green-socks-push.md
+++ b/.changeset/green-socks-push.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Introduce warningSeverity configuration option. This allows consumers to promote specific warnings to exceptions.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -222,3 +222,19 @@ configure({
     }
 })
 ```
+
+#### `warningSeverity: object`
+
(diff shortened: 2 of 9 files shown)


emereum:
Hey @iChenLei do you think there is a scenario where this change could be introduced? On one hand, it solves #3618, on the other I see how it might make triaging user bug reports more difficult - folks might promote warnings to errors, then submit bug reports here that aren't applicable to a default warning configuration.

In this case, we could suffix exceptions with something like: "You are seeing this error because you have configured MobX to throw when this occurs. Please revert your warning severity configuration and retry before reporting issues to MobX"

emereum:
@urugator keen to get your thoughts on this PR as well as my thoughts above on the potential downsides and ways to mitigate them.

PR #3636, opened 2023-02-24
Add opt-in, fully reactive props

Matchlighter:
This one feels like I'm bringing something back from the grave (https://github.com/mobxjs/mobx-react/issues?q=observable+props+label%3Aobservable-props-or-not), but it's still an issue, that, looking at the history, seems like the main reason it was closed was due to hooks "offering better alternatives".
I'm evidently an old fogey that still likes his class components, so I ultimately wrote this change as a `patch-package` patch that I've been applying to apps. Since I'm working on another PR, I figured I'd extract it and submit it as a PR so that I can (hopefully) stop having to patch it in everywhere.

So, for your consideration, opt-in fully reactive props.

### Code change checklist


urugator:
We've removed observable props altogether:
https://github.com/mobxjs/mobx/blob/main/packages/mobx-react/README.md#note-on-using-props-and-state-in-derivations

Expect issues with any workaround that updates observables during render.

Matchlighter:
@urugator Well that bites. :+1:

Matchlighter:
> this pattern is fundamentally incompatible with StrictMode in React 18.2 and higher

@urugator Is there a place that this is elaborated a bit more? I want to understand this more, but I'm not seeing anything that seems relevant in the React changelogs.

urugator:
To support React concurrent rendering we are now maintaining global state version. Version is updated with any observable mutation. Version is checked by `useSyncExternalStore` in functional components. React won't output anything until all components have matching versions, so you don't see temporal inconsistencies in the output (tearing). If you update observable during render, you update state version - if multiple components are involved React won't ever settle and is caught in render loop. So you would need some kind of state isolation or notion of "local state", that is detached from the "global state". Even with an escape hatch that would allow updating observable without incrementing state version, you would propably have to ensure that this local state won't ever leak outside the component, otherwise you are back in tearing scenario (not 100% sure about this).
We decided we don't want to keep hacking into react, piling up workaround and risking further incompatibilities.

We also found some "bugs" we wasn't aware of. They may have popped out with recent React versions, dunno. If you pass callback that accesses `this.props`, the child component subscribes to it's parents props, which leads to extra re-renders.

There were some other things to consider: 
Performance - it affects every class component whether it's needed or not. Besides the extra observable related logic, the implementation had to perform 2 shallow checks instead of one at every update. Not ideal.
Consistency with functional components - they have to deal with the same problem and we gave up on attempts like `useAsObservableState`. The workaround is now the same for both types of components. You can still introduce an observable variant of props/state, but you have to  update them manually via `componentDidUpdate` / `useEffect`:
https://github.com/mobxjs/mobx/blob/1a693efb61e880a583393e00b4d6878ed06bb13e/packages/mobx-react/__tests__/observer.test.tsx#L1233-L1244
Obviously we don't expect one should be doing this on day to day basis.

Matchlighter:
@urugator If I'm understanding, things would work (as they have in the past) if props were to invalidate `computed`s w/o bumping the global state version, but doing such would (rightfully) be considered "piling up workarounds"? (Not a suggestion so much as trying to confirm my understanding)

urugator:
If you assume only local computeds, then maybe, but you can leak computeds to different components:
```js
class Parent extends Component {
   store = observable({
     x = 1;
     get cmp() {
       return this.props.foo + this.store.x
     }
   })
   
   render() {   	
     <Child store={this.store} />
   }
}
```
`useSyncExternalStore` gives you `callback` that you should call when state changes, which we do. However if you don't update state version, this `callback` is noop. So this would not notify `Child` component.
You could probably workaround that by using other means for updating the component, like https://github.com/mobxjs/mobx/blob/ee09a6742bcd2ac7e599b1077b54670c6973d8e1/packages/mobx-react-lite/src/useObserver.ts#L52-L53
However, when the `Child` re-renders it reads different computed value than in previous render, but by not updating state version, you're signaling to react the state is the same as before. This obviously breaks the contract. Is that an actual problem? I am honestly not sure.

Matchlighter:
Sorry, I think I'm starting to get it...! How does having a global state not cause extra renders - it looks like `getSnapshot` is returning that global state version, which changes when any MobX state is affected. Wouldn't that mean:
- Given I had some Observable that was completely unrelated to and React components
- I update that Observable global state version is different, `getSnapshot` returns a different value and triggers a re-render?

Perhaps relevant - am I understanding this right as well:
>React won't output anything until all components have matching versions

My understanding here is that React calls `getSnapshot` implicitly at multiple points (or is it only called after `onStoreChange` - ignoring the `onStoreChange` event if the snapshots are identical?). If the returned value is ever different, the previous tentative/uncommitted render of that component is invalidated.

From the React docs, it sounds like `getSnapshot` should change if and only if there was a preceding `onStoreChange`, otherwise it should _always_ return the last value.This seems to not be the case here since it will change whenever the global version is affected, but `onStoreChange` is only called when an observed changes (and triggers the `observer${adm.name}` Reaction).

PR #3637, opened 2023-02-24
Add support for WeakRefs

Matchlighter:
Implement options for making `computed`s and `reactions`/`autoruns` garbage-collectable without manually disposing them.

I've had a couple apps that I've needed to use `keepAlive` `computed`s. I'm lazy and can't make promises that future me/future my app will dispose them properly. Now that JS supports `WeakRef`s, this is a less of a problem. The use case is pretty slim, mainly around `computed(() => {}, { keepAlive: true })` stuff, but it can be used for `autoruns` too (though you'll need to keep a reference to the `autorun` in some other way less it fall out of scope and get garbage collected).


rvion:
nice, sounds handy !

PR #3642, opened 2023-03-04
Update actions.md

shum-dev:
Get rid of unnecessary arguments in the constructor functions. OR we can use those "value" arguments to initiate the value variables.


kubk:
Nice catch 👍

PR #3649, opened 2023-03-07
Fix #3648: `observableRequiresReaction`/`computedRequiresReaction` shouldn't warn with `enableStaticRendering(true)`

### LINKED ISSUE #3648: using `enableStaticRender()` on the server causes "read outside a reactive context" warnings on server
**Intended outcome:**
Trying to prevent memory leaks on SSR using the `enableStaticRendering()` function while also avoiding warnings about data being read outside a reactive context

**Actual outcome:**
Calling `enableStaticRendering()` causes the mobx warnings to be produced on my server.

I can't provided a code sandbox since it doesn't support server-side rendering

**Versions**
latest versions of `mobx` and `mobx-react`

urugator:


urugator:
Hm seems it's gonna be tricker in relation to `computedRequiresReaction`

urugator:
Why `allowStateReads` is se to `true` by default?
https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/core/globalstate.ts#L95
that doesn't seem correct...

packages/mobx-react-lite/__tests__/observer.test.tsx
```diff
@@ -1084,3 +1084,24 @@ test("Anonymous component displayName #3192", () => {
...
+        mobx.configure({ observableRequiresReaction: true, computedRequiresReaction: true })
+        const o = mobx.observable.box(0, { name: "o" })
+        const c = mobx.computed(() => o.get(), { name: "c" })
+        const TestCmp = observer(() => <div>{o.get() + c.get()}</div>)
+
+        const { unmount, container } = render(<TestCmp />)
+        expect(container).toHaveTextContent("0")
+        unmount()
+
+        expect(consoleWarnMock).toMatchSnapshot()
```
mweststrate, line 1101: Q: what does this assertion do? Should that be `expect(consoleWarnMock).not.toBeCalled()`?

packages/mobx-react-lite/src/observer.ts
```diff
@@ -79,10 +77,6 @@ export function observer<P extends object, TRef = {}>(
     }
 
     // The working of observer is explained step by step in this talk: https://www.youtube.com/watch?v=cPF4iBedoF0&feature=youtu.be&t=1307
-    if (isUsingStaticRendering()) {
-        return baseComponent
```
mweststrate, line 83: Is this change intended? Is that to make sure we can disable the flag as done below? I'm wondering if it would be simpler in that case to just wrap a sheep wrapper function around the original component definition so that all the observer setup stuff is skipped?

packages/mobx-react/__tests__/observer.test.tsx
```diff
@@ -1037,3 +1038,30 @@ test("SSR works #3448", () => {
...
+            render() {
+                return o.get() + c.get()
+            }
+        }
+
+        const { unmount, container } = render(<TestCmp />)
+        expect(container).toHaveTextContent("0")
+        unmount()
+
+        expect(consoleWarnMock).toMatchSnapshot()
```
mweststrate, line 1061: See previous comment

packages/mobx-react/src/observerClass.ts
```diff
@@ -67,7 +67,14 @@ export function makeClassComponentObserver(
     }
     target.render = function () {
         this.render = isUsingStaticRendering()
-            ? originalRender
+            ? function () {
```
mweststrate, line 70: From two comments back, like this basically :)

packages/mobx/src/core/globalstate.ts
```diff
@@ -90,9 +90,9 @@ export class MobXGlobals {
 
     /**
      * Is it allowed to read observables at this point?
-     * Used to hold the state needed for `observableRequiresReaction`
+     * Used to hold the state needed for `observableRequiresReaction`/`computedRequiresReaction`
      */
-    allowStateReads = true
+    allowStateReads = false
```
mweststrate, line 95: This might be a breaking change? By default we allow state reads outside reactive contexts?

mweststrate (changes requested):
Looking mostly ok! Left some questions

packages/mobx/src/core/globalstate.ts
urugator: No. These flags, `allowStateReads` and `allowStateChanges`, says whether you're in a context, where reads/changes are allowed - or perhaps better where they are supposed to take place. They're basically `inDerivation`/`inAction` flags - they're switched on/off when entering/leaving these, regardless of whether it should warn or not. We use these flags instead of directly checking for `action`/`derivation`, because the mapping isn't always exactly 1:1. But really the point is to demark these places, not to directly control the warning - the warning is always only raised in combination with `requiresReaction`/`enforceAction`. They should always start with `false`, because there is no running action/derivation by default.

packages/mobx-react-lite/__tests__/observer.test.tsx
urugator: The mocked function serializes to an object with an array of calls and results, eg like here:
https://github.com/mobxjs/mobx/blob/b05595be8614fa8ba2137fd4050f933bd835a044/packages/mobx-react/__tests__/__snapshots__/observer.test.tsx.snap#L41-L55
So since it's serialized only to `[MockFunction]` it means it wasn't called.

packages/mobx-react-lite/src/observer.ts
urugator: Yes it's intended. Once you wrap it in a different function, you have to do almost the same machinery - forward ref, hoist name, statics, etc etc, maybe except memo ... basically most of the code here has nothing to do with setting up the "observer", it's dealing with the fact that we have to replace the original function with a new one. I could do the check a bit earlier, like eg here (avoiding `useObserver` + extra closure):
```
let observerComponent = (props: any, ref: React.Ref<TRef>) => {
  return useObserver(() => render(props, ref), baseComponentName)
}
```
but note `useObserver` is still part of the public API and can be used standalone.
Btw notice this is another place where proxy (instead of a wrapper) would make things easier.

One more thing to note - one solution to this problem is to simply say, don't use `observableRequiresReaction: true` on the server, it's useless anyway. The point of the PR is to provide something less obstructive.

packages/mobx-react-lite/__tests__/observer.test.tsx
mweststrate: Got ya! thanks

packages/mobx-react-lite/src/observer.ts
mweststrate: Ok, sounds good!

packages/mobx/src/core/globalstate.ts
mweststrate: Check, that adds up. I haven't been working on this for too long 😅.

mweststrate (approved):
LGTM

mweststrate:
Needs rebase

heath-freenome:
@urugator I'm still waiting for this. Can you rebase and merge?

urugator:
Will try to take a look over the weekend. It will require some adaptation to the reworked observers.

packages/mobx/src/core/globalstate.ts
urugator: Turns out it's more complicated. I think it was originally designed as described, but it was maybe misunderstood and changed with the introduction of `autoAction`.
The thing is that `autoAction` is supposed to warn when you modify state inside derivation, regardless of `enforceAction` configuration. So `autoAction` assumes that if `allowStateChanges` is false, there will be a warning when setting state, no other conditions needed.
In order for this to work the `configure/resetGlobalState` was updated to synchronize `allowStateChanges` with `enforeAction` configuration.
But the actual check for the warning doesn't assume this behavior:
https://github.com/mobxjs/mobx/blob/13a222ea21bd55b7158d48a998846da608080ded/packages/mobx/src/core/derivation.ts#L140-L143
Notice there is a bug - the check doesn't respect `enforceActions: "never"` (enforceActions === false).
This bug is actually a reason why [the test for `autoAction` warning](https://github.com/mobxjs/mobx/blob/13a222ea21bd55b7158d48a998846da608080ded/packages/mobx/__tests__/v5/base/action.js#L603-L623) is passing - there is [a test above](https://github.com/mobxjs/mobx/blob/13a222ea21bd55b7158d48a998846da608080ded/packages/mobx/__tests__/v5/base/action.js#L308) that calls `mobx.configure({ enforceActions: "never" })`.
If you fix this bug, the `autoAction` warning (*Side effects like changing state are not allowed*) can never occur.

urugator:
Update: I've spent a whole day on this, I've been dealing with some [unanticipated issues](https://github.com/mobxjs/mobx/pull/3649/files#r1375306679). It might be possible to resolve your issue without bothering about these, not sure. If/when I find some more time I will try to focus on that. If you wanna give it a try, just revert the last commit (`checkpoint`) and go from that.

packages/mobx-react/__tests__/observer.test.tsx
```diff
@@ -295,27 +296,6 @@ test("issue 12", () => {
     expect(events).toEqual(["table", "row: coffee", "row: tea", "table", "row: soup"])
 })
 
-test("changing state in render should fail", () => {
```
urugator, line 298: This test is wrong on multiple levels:
I don't recall changing state in render should ever throw, with a slight exception of `autoAction`, which doesn't work ATM as mentioned elsewhere.
It doesn't `expect(...).toThrow()`, it just checks for the error (that is never thrown).
If it's about `enforceAction`, it should expect `console.warn` instead of an error.
It pollutes console with `Warning: Cannot update a component while rendering a different component` (this is a bit surprising given that only one component is involved, but not a concern atm).

urugator:
Update: Should be ready for re-review + merge and we may tackle the other issues elsewhere.

heath-freenome:
I'm curious why this MR hasn't been merged?

PR #3664, opened 2023-03-26
fix: fix mobx-react-lite release issue, fixes #3662

### LINKED ISSUE #3662: mobx-react depends on mobx-react-lite 4 which does not exists
mobx-react@8.0.0:
  version "8.0.0"
  resolved "https://registry.yarnpkg.com/mobx-react/-/mobx-react-8.0.0.tgz#ec3e13a19662f452f14d279c59168d94385c86f0"
  integrity sha512-WKfYrxIDmI/ysmugybfWBFnKC0li4VvLHhfZqn1y5hXkOJ0selFFZ4rO8KHWZljWrzseNGRRqatYX9X4PaDD8A==
  dependencies:
    mobx-react-lite "^4.0.0"

mweststrate:
Running prepublish locally on an empty check-out fails the mobx-react-lite build, as the types are nog generated when mobx itself hasn't been build:

```
(typescript) Error: /Users/mweststrate/Desktop/mobx/packages/mobx-react-lite/src/useLocalStore.ts(1,28): semantic error TS2307: Cannot find module 'mobx' or its corresponding type declarations.
Error: /Users/mweststrate/Desktop/mobx/packages/mobx-react-lite/src/useLocalStore.ts(1,28): semantic error TS2307: Cannot find module 'mobx' or its corresponding type declarations.
```

Fixed it by forcing a MobX build in the prepublish script

### DIFF
--- a/.changeset/thin-bugs-boil.md
+++ b/.changeset/thin-bugs-boil.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+(Hopefully) fixed release process for mobx-react-lite
--- a/packages/mobx-react-lite/package.json
+++ b/packages/mobx-react-lite/package.json
@@ -75,6 +75,6 @@
         "test:size": "yarn import-size --report . observer useLocalObservable",
         "test:types": "tsc --noEmit",
         "test:check": "yarn test:types",
-        "prepublish": "yarn build --target publish && yarn build:cjs && yarn build:es"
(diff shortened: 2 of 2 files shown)


urugator:
Thank you. I still quite don't undestand the problem - why it used to work or why `mobx-react` doesn't have to do that ...

mweststrate:
My suspicion is that it depends on the order in which the publish commands are executed. But that order seems to be determined by Lerna and not by us.

PR #3673, opened 2023-03-28
Fix #3671: `observer` + `StrictMode`

### LINKED ISSUE #3671: `mobx-react-lite@4.0.2` observer not triggering component re-renders
The `observer` of `mobx-react-lite` seems to not trigger the wrapped component to re-renders after upgrading from `mobx-react-lite@3.4.3` to `mobx-react-lite@4.0.2`.

**Intended outcome:**

When an `observable` value changes in the store, the `observer` component should re-render with the new value.

**Actual outcome:**

The `observer` component does not re-render when the `observable` value is updated.

Code Sandbox: https://codesandbox.io/s/new-dawn-b2shxq?file=/src/App.js

Clicking "Push new element to data" button will push a new element to the `store.data` array using a store action.

However, the component does not re-render even thought `store.data` has been updated (see console logs).

If you change the `mobx-react-lite` version from `4.0.2` to `3.4.3` in the sandbox, the component will re-render when you click the "Push new element to data" button.

Screenshot 2023-03-27 at 7 49 23 PM

**Versions**

`mobx-react-lite@4.0.2`

urugator:


mweststrate:
Put up a draft that avoids making props / state observable, so that it plays better with Reacts internals. This diff shows a) alternative patterns and b) a way to detect the otherwise now broken usage of props / state inside reactions / computed. https://github.com/mobxjs/mobx/commit/94d78da122c21e8332e0336948ca8c871e129a54

urugator:
props/state/context are no longer observable.
Accessing props/state/context in different derivation than current render throws.
Fixed #3691

TODO: 
Documentation

Shouldn't we throw instead?
```ts
if (componentClass[isMobXReactObserverSymbol]) {
        const displayName = getDisplayName(componentClass)
        console.warn(
            `The provided component class (${displayName})
                has already been declared as an observer component.`
        )
} else {
       componentClass[isMobXReactObserverSymbol] = true
}
```
Extending `observer` component doesn't seem to be supported, should we add an explicit check?

mweststrate:
> 
> Shouldn't we throw instead?
> 

Sounds good to me!

> 
> Extending `observer` component doesn't seem to be supported, should we add an explicit check?

Good idea, nails down that semantics at least. 

What can I do to help further push this change :) ?

urugator:
> What can I do to help further push this change :) ?

Documenting the non-reactive props change would be a big help :)

urugator:
Added tests demonstrating observable props/state/context workaround.

Applying `observer` multiple times now throws.

Explicit check for extending `observer` component turned out to be impossible without an access to observer base class constructor. Added note to documentation instead.

What would be nice to have, but I personally don't plan to do at the moment (help welcome):
- Test or sample app that integrates multiple functional / class components + strict mode and combines multiple techniques/approaches.

What should be done next:
- Document non-reactive props/state/context change - motivation + workarounds. When done, replace TODO here https://github.com/mobxjs/mobx/pull/3673/files#diff-43fae443fda65b95af0b56e07002a6ba8cefd6ca87a8c62933bc2ffd6a742a9cR255
- Do a proper release - the previous failed release jobs did not published release notes, so these need to be somehow revived.
- Update compatibility matrix
  https://github.com/mobxjs/mobx/blob/main/packages/mobx-react/README.md#compatibility-matrix
  https://github.com/mobxjs/mobx/blob/main/packages/mobx-react-lite/README.md#compatibility-table-major-versions

mweststrate:
* [ ] Document non-reactive props/state/context change - motivation + workarounds. When done, replace TODO here https://github.com/mobxjs/mobx/pull/3673/files#diff-43fae443fda65b95af0b56e07002a6ba8cefd6ca87a8c62933bc2ffd6a742a9cR255
* [ ] Update compatibility matrix
https://github.com/mobxjs/mobx/blob/main/packages/mobx-react/README.md#compatibility-matrix
https://github.com/mobxjs/mobx/blob/main/packages/mobx-react-lite/README.md#compatibility-table-major-versions
* [ ] Do a proper release - the previous failed release jobs did not published release notes, so these need to be somehow revived.

See https://github.com/mobxjs/mobx/pull/3718 as I don't think I can push to origin urugator.

mweststrate:
@urugator some items are still open, but how'd you feel about releasing the current state? Since I don't entirely trust the current state of releases / changelog, I'm tempted to trigger a release, and then verify is everything ended up well and roll forward if anything went south.

urugator:
go for it

mweststrate:
Merged, via #3718. Thanks for all the hard work here once again @urugator! (Feel free to invoice).

mweststrate:
@urugator just noticed no mobx-react-lite release was triggered. Should this have emitted a major there as well?

mweststrate:
Ah no, I think it worked.

PR #3676, opened 2023-04-02
Update understanding-reactivity.md

shum-dev:
.push() method actually does mutate the array


shum-dev:
sorry. this is my bad

PR #3709, opened 2023-06-20
Make trace() noop in production build

kubk:
This PR disables the `trace` functionality in the production build instead of throwing an error. Related discussion: https://github.com/mobxjs/mobx/discussions/3708


mweststrate:
Thanks!

PR #3717, opened 2023-07-06
fix(type): remove makeObservable and makeAutoObservable function proxy option

liucan233:
### Code change checklist


urugator:
Makes me wonder whether the setting shouldn't be part of annotation/enhancer, so it propagates to nested structures, but probably not worth the additional complexity.

packages/mobx/src/api/makeObservable.ts
```diff
@@ -23,10 +23,17 @@ import {
 // Fixes: https://github.com/mobxjs/mobx/issues/2325#issuecomment-691070022
 type NoInfer<T> = [T][T extends any ? 0 : never]
 
+// Options property "proxy" designs for "observable" function,
+// and makeObservable and makeAutoObservable is not Supported
+// const foo = observable({}), foo.bar = 1 will be monitored,
+// const foo = observable({}, {proxy: false}), foo.bar = 1 will be not monitored
+// The summary is that proxy is an invalid option for makeObservable and makeAutoObservable
+type NoProxyCreateObservableOptions = Omit<CreateObservableOptions, "proxy">
```
urugator, line 31: `MakeObservableOptions`

docs/observable-state.md
```diff
@@ -136,6 +136,20 @@ Inference rules:
 -   All _generator_ functions become `flow`. (Note that generator functions are not detectable in some transpiler configurations, if flow doesn't work as expected, make sure to specify `flow` explicitly.)
 -   Members marked with `false` in the `overrides` argument will not be annotated. For example, using it for read only fields such as identifiers.
 
+## Options {🚀}
```
urugator, line 139: Remove the added section and change the original:

> **`proxy: false`** forces `observable(thing)` to use non-[**proxy**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) implementation. This is a good option if the shape of the object will not change over time, as non-proxied objects are easier to debug and faster. See [avoiding proxies](#avoid-proxies).

=>

> **`proxy: false`** forces `observable(thing)` to use non-[**proxy**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) implementation. This is a good option if the shape of the object will not change over time, as non-proxied objects are easier to debug and faster. This option is not available for `make(Auto)Observable`, see [avoiding proxies](#avoid-proxies).

liucan233:
Thank you for all pointing out my errors, and All errors have been corrected. Perhaps it will take you some time to check again.

packages/mobx/src/api/makeObservable.ts
```diff
@@ -23,10 +23,17 @@ import {
 // Fixes: https://github.com/mobxjs/mobx/issues/2325#issuecomment-691070022
 type NoInfer<T> = [T][T extends any ? 0 : never]
 
+// Options property "proxy" designs for "observable" function,
+// and makeObservable and makeAutoObservable is not Supported
+// const foo = observable({}), foo.bar = 1 will be monitored,
+// const foo = observable({}, {proxy: false}), foo.bar = 1 will be not monitored
+// The summary is that proxy is an invalid option for makeObservable and makeAutoObservable
```
urugator, line 30: I am sorry, I totally missed it, but can you also remove these please?

urugator:
Thank you!

PR #3718, opened 2023-07-09
Urugator fix 3671

mweststrate:
See https://github.com/mobxjs/mobx/pull/3673


packages/mobx-react/README.md
```diff
@@ -87,6 +88,90 @@ class TodoView extends React.Component {
...
+class Alerter extends React.Component<{ counter: number }> {
+    @observable observableCounter: number
+    reactionDisposer
+
+    constructor(props) {
+        this.observableCounter = counter
+
+        // set up a reaction, by observing the observable,
+        // rather than the prop which is non-reactive:
+        this.reactionDisposer = autorun(() => {
```
urugator, line 128: Move to `componenDidMount`, otherwise this could leak (react can create instance and throw it away without calling any lifecycle methods). For inspiration see tests:
`Observable prop workaround`
`Observable props/state/context workaround`

mweststrate: Fixed

urugator:
Found a suggestions to run a microtask *before* calling the `gc()`, like:
```js
test('cleaned', async () => {
  const w = weak();
  await new Promise(resolve => setTimeout(resolve, 0));
  global.gc();
  expect(w.deref()).toEqual(null);
});
```

PR #3732, opened 2023-07-23
Fix #3728: observables initialization should not update state version

### LINKED ISSUE #3728: [Bug] Error: Maximum update depth exceeded
**Intended outcome:**
Render without any issues

**Actual outcome:**
`Error: Maximum update depth exceeded.`

Repro:
* Next.js https://stackblitz.com/edit/stackblitz-starters-wpvmwt?file=pages%2F_app.js
* Pure react https://stackblitz.com/edit/stackblitz-starters-aa5t8d?file=src%2Findex.tsx

1. `_app.jsx` - define the component as `observer`
2. `index.jsx` - use `useState` to store `mobx` store without lazy initialization (`useState(new Store())`)
4. App crashes because of `Maximum update depth exceeded`

Any component defined as `observer` higher in react tree than the defined `mobx` store is causing the issue above.

Lazy initialization (`useState(() => new Store())`), downgrading mobx-react-lite to 3.4.3  fixses issue or removing `observer` fixes issue.

**Versions**
mobx-react-lite@4.0.3

urugator:
-   fix: #3728: Observable initialization updates state version.
-   fix: Observable set initialization violates `enforceActions: "always"`.
-   fix: Changing keys of observable object does not respect `enforceActions`.
-   fix: #3734: `isolateGlobalState: true` makes observer stop to react to store changes

### DIFF
--- a/packages/mobx/__tests__/v5/base/observables.js
+++ b/packages/mobx/__tests__/v5/base/observables.js
@@ -2387,3 +2387,15 @@ test("state version updates correctly", () => {
     expect(o.x).toBe(5)
     expect(prevStateVersion).not.toBe(getGlobalState().stateVersion)
 })
+
+test.only("state version is not updated on observable creation", () => {
+    const globalState = getGlobalState()
+    let prevStateVersion = globalState.stateVersion
+    //mobx.observable({ x: 0 })
+    // mobx.observable.box(0)
+    //mobx.observable.map([["key", "val"]])
+    mobx.configure({ enforceActions: "always" })
+    mobx.makeAutoObservable({ x: "x" })
+
+    expect(prevStateVersion).toBe(globalState.stateVersion)
+})
--- a/packages/mobx/__tests__/v5/base/observables.js
+++ b/packages/mobx/__tests__/v5/base/observables.js
@@ -2388,6 +2388,15 @@ test("state version updates correctly", () => {
     expect(prevStateVersion).not.toBe(getGlobalState().stateVersion)
 })
 
+test("Atom.reportChanged does change state version when called from the batch the atom was created in", () => {
+    mobx.transaction(() => {
+        const prevStateVersion = getGlobalState().stateVersion
+        const atom = mobx.createAtom()
+        atom.reportChanged()
+        expect(prevStateVersion).toBe(getGlobalState().stateVersion)
+    })
+})
+
 test('Observables initialization does not violate `enforceActions: "always"`', () => {
     const consoleWarnSpy = jest.spyOn(console, "warn").mockImplementation(() => {})
 
@@ -2416,6 +2425,7 @@ test('Observables initialization does not violate `enforceActions: "always"`', (
         check(() => mobx.makeObservable({ x: 0 }, { x: true }))
         check(() => new MakeAutoObservable())
         check(() => mobx.makeAutoObservable({ x: 0 }))
+        check(() => mobx.extendObservable({}, { x: 0 }))
         check(() => mobx.observable(new Set([0])))
         check(() => mobx.observable(new Map([[0, 0]])))
         check(() => mobx.observable({ x: 0 }, { proxy: false }))
@@ -2470,11 +2480,23 @@ test("state version does not update on observable creation", () => {
         }
     }
 
+    class MakeObservableSubclass extends MakeObservable {
+        y = 0
+        constructor() {
+            super()
+            mobx.makeObservable(this, { y: true })
+        }
+    }
+
     check(() => mobx.observable(0))
     check(() => new MakeObservable())
+    // Batch is required by design - without batch reactions can be created/invoked inbetween individual constructors and they must see updated state version.
+    // https://github.com/mobxjs/mobx/pull/3732#discussion_r1285099080
+    check(mobx.action(() => new MakeObservableSubclass()))
     check(() => mobx.makeObservable({ x: 0 }, { x: true }))
     check(() => new MakeAutoObservable())
     check(() => mobx.makeAutoObservable({ x: 0 }))
+    check(() => mobx.extendObservable({}, { x: 0 }))
     check(() => mobx.observable(new Set([0])))
     check(() => mobx.observable(new Map([[0, 0]])))
     check(() => mobx.observable({ x: 0 }, { proxy: false }))
--- a/packages/mobx/__tests__/v5/base/observables.js
+++ b/packages/mobx/__tests__/v5/base/observables.js
@@ -2388,14 +2388,98 @@ test("state version updates correctly", () => {
     expect(prevStateVersion).not.toBe(getGlobalState().stateVersion)
 })
 
-test.only("state version is not updated on observable creation", () => {
-    const globalState = getGlobalState()
-    let prevStateVersion = globalState.stateVersion
-    //mobx.observable({ x: 0 })
-    // mobx.observable.box(0)
-    //mobx.observable.map([["key", "val"]])
-    mobx.configure({ enforceActions: "always" })
-    mobx.makeAutoObservable({ x: "x" })
+test('Observables initialization does not violate `enforceActions: "always"`', () => {
+    const consoleWarnSpy = jest.spyOn(console, "warn").mockImplementation(() => {})
 
-    expect(prevStateVersion).toBe(globalState.stateVersion)
+    const check = cb => {
+        cb()
+        expect(consoleWarnSpy).not.toBeCalled()
+    }
+
+    class MakeObservable {
+        x = 0
+        constructor() {
+            mobx.makeObservable(this, { x: true })
+        }
+    }
+    class MakeAutoObservable {
+        x = 0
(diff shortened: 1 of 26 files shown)


packages/mobx/src/api/extendobservable.ts
```diff
@@ -41,6 +44,8 @@ export function extendObservable<A extends Object, B extends Object>(
     const descriptors = getOwnPropertyDescriptors(properties)
 
     const adm: ObservableObjectAdministration = asObservableObject(target, options)[$mobx]
+    const allowStateChanges = allowStateChangesStart(true)
+    globalState.suppressReportChanged = true
```
urugator, line 48: The plan is to change these into `suppressReportChangedStart/End` and probably move it to `initObservableStart/End` (together with `unstrackedStart/End` and `allowStateChangesStart/End` as noted in comments).

urugator:
When initilizing observable structures, we should do 3 things:
1. `allowStateChanges` so we don't violate `enforceActions`.
2. `untracked` so we don't accidentaly subscribe to anything observable accessed during init in case the observable is created inside derivation.
3. `supressReportChanged` mainly to prevent version update, but also there can't be any observers yet, so there is nothing to report to, therefore we can skip the logic.

Do you agree? @mweststrate

urugator:
I have to yet investigate failing tests. From a quick first look there seems to be a reduction in render counts, which may actually be correct. I will look into it later, once you confirm that the general direction is fine.

packages/mobx/src/core/atom.ts
```diff
@@ -65,6 +65,9 @@ export class Atom implements IAtom {
      * Invoke this method _after_ this method has changed to signal mobx that all its observers should invalidate.
      */
     public reportChanged() {
+        if (globalState.suppressReportChanged) {
```
mweststrate, line 68: This feels like a potentially dangerous check, that silently might break reactivity. E.g. do the following cases behave correctly?


```javascript
const x = observable({ a: 1 })
reaction(x => x.a, a => console.log("a", a))
extendObservable(x, { a: 2, b: 1 })
```

```javascript
const x = observable({ a: 1 }, { proxy: true })
// react to new keys
reaction(x => Object.keys(x.a), a => console.log("a", a))

x.b = 1
```

I'm wondering if an alternative approach would be: only increase stateVersion if there are subscribers to the current atom?

packages/mobx/src/api/extendobservable.ts
mweststrate: yeah, some dedicated utility sounds great here to avoid copy / paste mistakes.

packages/mobx/src/core/atom.ts
urugator: > This feels like a potentially dangerous check, that silently might break reactivity

Good point. Imo the play here is to make sure it's suppressed only on creation and not on mutation like in this case.
Key changes should invoke reactions and therefore update state version, these two goes hand in hand, a reaction must always see a new state version. Therefore you shouldn't extend observable during render.
Just for clarification: `extendObservable` does notify about key changes.

> only increase stateVersion if there are subscribers to the current atom

Purely from semantic perspective, two different states should aways have different versions, doesn't matter if someone was observing some part of the state or not. That's quite easy to reason about.
Practically speaking, I don't know. I have a bad feeling about it, but I have a hard time to come up with a convincing example. My worry is, that you change some observable `x`, that nobody is observing, so version stays the same. Then something unrelated to mobx - props/state or other external sync store - changes in such a way, that the next render would read from `x`.  Now if react would check whether mobx external state changed, it doesn't know about the change of `x`, even though it's about to use it together with rest of the world (other observables, state, props, other external state) to render. I am not sure if that alone is a problem. The goal here is to avoid inconsistencies and to have an inconsistency, the `x` would probably have to be somehow correlated with the rest of the world (props/state/other store).
But then, if they're correlated, they had to change at the same moment and therefore something, somewhere had to already signal a new version? It's hard for me to contemplate about this, especially when I don't know what react exactly does under the hood and what it's assumptions are. If version always changes, I don't have to think about this...

mweststrate: > Purely from semantic perspective, two different states should aways have different versions, doesn't matter if someone was observing some part of the state or not. That's quite easy to reason about

>  If version always changes, I don't have to think about this...

Agreed, I think these principle weighs stronger than my concerns. Let's continue this direction.

urugator: There is a fundamental problem with subclasses - `new Subclass()` always counts as a state update, because the subclass's constructor calls `makeObservable` on an already observable object (`this`).
Strictly speaking it's correct behavior, because constructors can insert any arbitrary logic between individual `makeObservable` calls and there is no guarantee `new Subclass` is called from transaction, so a reaction invoked in between these `makeObservable` calls should see new state version.

What we would need is to treat all modifications of the same atom inside the same batch it was created as part of initialization. To do that we would need to introduce `globalState.batchId`, assign this `batchId` to every atom when created and suppress `reportChanged` when `atom.batchId === globalState.batchId`. Now you don't have to manually demark initialization sections - they span from atom creation until the end of the batch. If you call `new Subclass()` from action as a good citizen, the state version stays same, no matter how many times you called `makeObservable`. Thoughts?

mweststrate: Ok, that sounds smart! It might serve in the future as well to better handle any weird edge cases around reaction creations in actions, recursive updates etc etc, although I believe we handle those all consistently now.

packages/mobx/__tests__/v5/base/observables.js
```diff
@@ -2387,3 +2387,121 @@ test("state version updates correctly", () => {
     expect(o.x).toBe(5)
     expect(prevStateVersion).not.toBe(getGlobalState().stateVersion)
 })
+
+test("Atom.reportChanged does change state version when called from the batch the atom was created in", () => {
```
urugator, line 2391: missing "NOT"

packages/mobx/src/types/type-utils.ts
```diff
@@ -92,3 +98,22 @@ export function getDebugName(thing: any, property?: string): string {
...
     return named.name_
 }
+
+/**
+ * Helper function for initializing observable structures, it applies:
+ * 1. allowStateChanges so we don't violate enforceActions.
+ * 2. untracked so we don't accidentaly subscribe to anything observable accessed during init in case the observable is created inside derivation.
+ * 3. batch to avoid state version updates
+ */
+export function initObservable<T>(cb: () => T): T {
```
mweststrate, line 108: NIT: this feels so core it might fit better into atom.ts or derivation.ts? Not sure

mweststrate (approved):
Really awesome work @urugator, great fixes and clear to follow! Probably good to run the perf tests. Left two comments, but beyond that looks fine to merge by me!

packages/mobx/src/types/type-utils.ts
urugator: It's helper for initializing "complex" observable structures - things inside `types` dir, but I don't have strong opinions about it.

mweststrate: Fair enough!

packages/mobx/src/types/observablemap.ts
```diff
@@ -90,11 +90,12 @@ export type IObservableMapInitialValues<K = any, V = any> =
...
 // But: https://github.com/mobxjs/mobx/issues/1556
 export class ObservableMap<K = any, V = any>
-    implements Map<K, V>, IInterceptable<IMapWillChange<K, V>>, IListenable {
+    implements Map<K, V>, IInterceptable<IMapWillChange<K, V>>, IListenable
+{
     [$mobx] = ObservableMapMarker
-    data_: Map<K, ObservableValue<V>>
-    hasMap_: Map<K, ObservableValue<boolean>> // hasMap, not hashMap >-).
-    keysAtom_: IAtom
+    data_!: Map<K, ObservableValue<V>>
```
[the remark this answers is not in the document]
urugator: Will take a look at that. I wonder whether `"useDefineForClassFields": true` has any effect on this, since it replaces assigments with the `Object.defineProperty` calls.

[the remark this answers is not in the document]
urugator: Actually
```ts
class Foo {
 prop!;
}
// transpiles to
"use strict";
class Foo {
    constructor() {
        Object.defineProperty(this, "prop", {
            enumerable: true,
            configurable: true,
            writable: true,
            value: void 0
        });
    }
}
```
So I quess it doesn't matter?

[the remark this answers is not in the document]
mweststrate: Yes you're right, I still have 7 year old micro-optimizations against ES5 in my head when I wrote first versions 😅. Feel free to merge!

Egor-Chib:
We are using MOBX 6.10.2
and it looks like the issue persists: if I create children component, which do "new Something()" and "something" has in constructor "makeObservable" - this will cause re-render of parent.

PR #3748, opened 2023-08-31
fix #3747

mweststrate:
### Code change checklist

### DIFF
--- a/packages/mobx/src/core/atom.ts
+++ b/packages/mobx/src/core/atom.ts
@@ -68,7 +68,7 @@ export class Atom implements IAtom {
      * Invoke this method _after_ this method has changed to signal mobx that all its observers should invalidate.
      */
     public reportChanged() {
-        if (globalState.inBatch && this.batchId_ !== globalState.batchId) {
+        if (!globalState.inBatch && this.batchId_ !== globalState.batchId) {
             // We could update state version only at the end of batch,
             // but we would still have to switch some global flag here to signal a change.
             globalState.stateVersion =
--- a/packages/mobx/src/core/atom.ts
+++ b/packages/mobx/src/core/atom.ts
@@ -68,21 +68,19 @@ export class Atom implements IAtom {
      * Invoke this method _after_ this method has changed to signal mobx that all its observers should invalidate.
      */
     public reportChanged() {
-        if (globalState.inBatch && this.batchId_ === globalState.batchId) {
-            // Called from the same batch this atom was created in.
-            return
+        if (globalState.inBatch && this.batchId_ !== globalState.batchId) {
+            // We could update state version only at the end of batch,
+            // but we would still have to switch some global flag here to signal a change.
+            globalState.stateVersion =
+                globalState.stateVersion < Number.MAX_SAFE_INTEGER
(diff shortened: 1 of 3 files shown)


mweststrate:
Fails autoruns rescheduling:

Screenshot 2023-08-31 at 10 04 20

packages/mobx/src/core/atom.ts
```diff
@@ -68,21 +68,19 @@ export class Atom implements IAtom {
      * Invoke this method _after_ this method has changed to signal mobx that all its observers should invalidate.
      */
     public reportChanged() {
-        if (globalState.inBatch && this.batchId_ === globalState.batchId) {
-            // Called from the same batch this atom was created in.
-            return
+        if (globalState.inBatch && this.batchId_ !== globalState.batchId) {
```
urugator, line 71: `!globalState.inBatch || this.batchId_ !== globalState.batchId`
If not in batch mutation always updates version immediately. 
If in batch only if it's different.

urugator:
Btw the state version is only relevant for `observer`, so with the proposed fix, it should not interfere with inner workings of mobx like autoruns scheduling etc in any way.
However, no matter the implementation, we still assume the following:
If you create and mutate an observable in the same batch, then there should be no need to notify any derivations except the computeds created inside that same batch. Because there should be no way to somehow leak that observable to some different derivation (except created computeds) that isn't currently running. Eg. even if you consider nesting derivations like:
```ts
autorun(() => {
  // batch id 1
  const o = observable.box(0);
  autorun(() => {
    // This won't run until the outer `autorun` finishes
   // batch id 2
    o.get();
  });
  o.set(1); // no point in reporting change to nested autorun, it's not subscribed to `o` yet.
})
```
If there is a case, that I am not aware of, where that isn't true, then there is a problem.

mweststrate:
Heh, the fix ends in state `o.x === 3` instead of `o.x === 2`, whereas `5` is (correctly imho) expected  so still doesn't seem to fully reschedule itself correctly?

packages/mobx/src/core/atom.ts
```diff
@@ -68,7 +68,7 @@ export class Atom implements IAtom {
      * Invoke this method _after_ this method has changed to signal mobx that all its observers should invalidate.
      */
     public reportChanged() {
-        if (globalState.inBatch && this.batchId_ !== globalState.batchId) {
+        if (!globalState.inBatch && this.batchId_ !== globalState.batchId) {
```
urugator, line 71: && => ||

mweststrate: heh lol I should try to read before copy pasting next time. That fixes the test. However, the autorun doesn't initially reschedule itself (see the comment in the test), I think this used to be the case in the past, or am I mistaken here?

urugator: > I think this used to be the case in the past, or am I mistaken here?

I don't know if this was the case in the past, but it wasn't the case when I wrote this test, otherwise there wouldn't be that extra `o.x++` outside the autorun to nudge it and it would fail, because `o.x` would end up being 6. I think I was probably a bit surprised as well when writing it. However, if there is some problem here, it was there before introducing this state version business.

PR #3763, opened 2023-09-23
Fix: disable global state version usage in observer. Fixes #3728

### LINKED ISSUE #3728: [Bug] Error: Maximum update depth exceeded

mweststrate:
This diff disables the usage of global state versioning. The reasoning here is:

* global state versioning is principle more correct
* effectively it doesn't make the implementation more compatible with concurrent mode, both implementations fail layer 3 compatibility
* in neither cases, the snapshot represents what actually will be read during render; as the latest version of all data is read, instead of the state described by the snapshot version. 
* since the local versioning approach doesn't break #3728 (for which could be argued it is unidiomatic), there seems at this moment little upside to the global version approach.

This PR deliberately doesn't clean up the underlying infra; the tracking of a global state version in the first place. I will probably put up a separate PR for that, but I also want to pursue more implementation solutions that might work better, that build on that infra.

### DIFF
--- a/packages/mobx-react-lite/src/useObserver.ts
+++ b/packages/mobx-react-lite/src/useObserver.ts
@@ -25,7 +25,7 @@ type ObserverAdministration = {
 }
 
 // BC
-const globalStateVersionIsAvailable = typeof _getGlobalState().stateVersion !== "undefined"
+const globalStateVersionIsAvailable = false // See #3728; TODO: typeof _getGlobalState().stateVersion !== "undefined"
 
 function createReaction(adm: ObserverAdministration) {
     adm.reaction = new Reaction(`observer${adm.name}`, () => {
--- a/packages/mobx-react-lite/src/useObserver.ts
+++ b/packages/mobx-react-lite/src/useObserver.ts
@@ -1,4 +1,4 @@
-import { Reaction, _getGlobalState } from "mobx"
+import { Reaction } from "mobx"
 import React from "react"
 import { printDebugValue } from "./utils/printDebugValue"
 import { isUsingStaticRendering } from "./staticRendering"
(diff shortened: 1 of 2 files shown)


urugator:
I think you can remove:
```ts
// Provides ability to force component update without changing state version
const [, forceUpdate] = React.useState<Symbol>()
```
and change
```ts
               if (!adm.reaction) {
                    // We've lost our reaction and therefore all subscriptions, occurs when:
                    // 1. Timer based finalization registry disposed reaction before component mounted.
                    // 2. React "re-mounts" same component without calling render in between (typically <StrictMode>).
                    // We have to recreate reaction and schedule re-render to recreate subscriptions,
                    // even if state did not change.
                    createReaction(adm)
                    // Update version otherwise onStoreChange() would be ignored
                    adm.stateVersion = Symbol()
                    onStoreChange();
                }
```
I don't know if it's completely kosher to call `onStoreChange` during subscription, but IIRC it was working, the only problem was the call was ignored, because the version stayed the same, therefore the workaround with `forceUpdate`.

packages/mobx-react-lite/src/useObserver.ts
```diff
@@ -69,7 +60,8 @@ export function useObserver<T>(render: () => T, baseComponentName: string = "obs
                     // even if state did not change.
                     createReaction(adm)
                     // `onStoreChange` won't force update if subsequent `getSnapshot` returns same value.
-                    forceUpdate(Symbol())
+                    // So we make sure that is not the case
+                    adm.stateVersion = Symbol()
```
urugator, line 64: Ehm, but you still have to call that `onStoreChange()` ... ?

mweststrate: No I don't think so, I think due to uSES getting a different version from getSnapshot, it will schedule update already itself. I guess React checks that?

mweststrate: (at least the test that failed without the forceUpdate, now succeeds again, the exact calling order of uSES remains a little mistery)

urugator: Ah, yea, I think you're correct, but then the comment on line 62 is perhaps a bit misleading.

mweststrate: ah yeah, didn't clean that up properly, will do in the cleanup branch

PR #3764, opened 2023-09-23
State version cleanup

mweststrate:
Follow up of #3763, cleanup global stateVersion. Kept initObservable as is, as it handles several flags around warnings etc better.

### DIFF
--- a/packages/mobx-react-lite/src/useObserver.ts
+++ b/packages/mobx-react-lite/src/useObserver.ts
@@ -25,7 +25,7 @@ type ObserverAdministration = {
 }
 
 // BC
-const globalStateVersionIsAvailable = typeof _getGlobalState().stateVersion !== "undefined"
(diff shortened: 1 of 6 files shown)


packages/mobx-react-lite/src/useObserver.ts
```diff
@@ -13,10 +13,11 @@ const getServerSnapshot = () => {}
 type ObserverAdministration = {
     reaction: Reaction | null // also serves as disposed flag
     onStoreChange: Function | null // also serves as mounted flag
-    // BC: we will use local state version if global isn't available.
-    // It should behave as previous implementation - tearing is still present,
+    // stateVersion that 'ticks' for every time the reaction fires
+    // tearing is still present,
```
urugator, line 17: Is it? I think you can remove these...

urugator (approved):

PR #3776, opened 2023-10-19
Fix: observer class components going stale after 10 seconds

wbercx:
After upgrading from MobX 4 to MobX 6 in a React Native project (React 18.2, no StrictMode), I noticed almost all of my screens went stale shortly after rendering.

An observer class component is registered with `observerFinalizationRegistry` like so:
```ts
_observerFinalizationRegistry.register(this, admin, this);
```
The third argument is the `token`, which is used to store the registration in a Map. In `componentDidMount` the finalization is then unregistered so it is allowed to stay alive, but using `admin` as the token, not `this`:
```ts
_observerFinalizationRegistry.unregister(admin);
```
`admin` is not registered to begin with, so the finalization does not get cleaned up, resulting in every class component going stale after about 10 seconds of being rendered.

### Code change checklist

Happy to write tests for this - but I will need some pointers. I saw the `finalizationRegistry` test, but it is not immediately obvious to me how I should approach testing registration and deregistration.

### DIFF
--- a/.changeset/hip-bulldogs-act.md
+++ b/.changeset/hip-bulldogs-act.md
@@ -1,0 +1,5 @@
+---
+"mobx-react": patch
+---
+
+Fixed premature disposal of class component observers.
--- a/packages/mobx-react/src/observerClass.ts
+++ b/packages/mobx-react/src/observerClass.ts
@@ -133,7 +133,7 @@ export function makeClassComponentObserver(
         admin.mounted = true
 
         // Component instance committed, prevent reaction disposal.
-        observerFinalizationRegistry.unregister(admin)
+        observerFinalizationRegistry.unregister(this)
 
         // We don't set forceUpdate before mount because it requires a reference to `this`,
         // therefore `this` could NOT be garbage collected before mount,


urugator:
Damit. Thank you! LGTM

Test is not mandatory, but ppbly something like following:
```ts
test("should unregister from FinalizationRegistry once commited #3776", async () => {
    expect(typeof globalThis.FinalizationRegistry).toBe("function")

    const o = mobx.observable({ x: 0 })
    let strongRef = {};
    const weakRef = new WeakRef(strongRef);

    @observer
    class TestCmp extends React.Component<any> {
        render() {
            return o.x
        }
    }

    const { unmount, container } = render(<App />)

    expect(container).toHaveTextContent("0")
     
    // Wait for GC to kick in
    strongRef = undefined;
    await Promise.resolve()
    gc()
    await waitFor(() => expect(weakRef!.deref()).toBeUndefined(), {
        timeout: 10_000,
        interval: 150
    })
    // GC did run, component is mounted and should remain reactive
    act(() => o.x++);
    expect(container).toHaveTextContent("1")

    unmount()
})
```

urugator:
The above doesn't work, because the `strongRef` is never reclaimed by GC and even if it would, it wouldn't necessarily mean that the component instance is reclaimed as well. We can force finalization with our TimerBasedFinalizationRegistry, so I wrote a test using these: https://github.com/mobxjs/mobx/pull/3787

Issue #3767, opened 2023-09-26
The value returned from `computed(() => ...)` shouldn't have a `set(...)` method


Nokel81:
Current Behaviour:

`computed(...)` currently returns the `IComputedValue<T>` type which has the following signature:

```tsx
interface IComputedValue<T> {
  get(): T;
  set(value: T): void;
}
```

However, actually using that `.set(...)` results in an error being thrown `TypeError: cannot call function undefined`.

Expected Behaviour:
`computed(...)` should return a type that only has the `get()` field as valid.

Notes:
I tried providing a `set` impl through the `options` for `computed` but the same was still the case

versions:
```
mobx -> 6.10.9
```

mweststrate:
Please follow the issue template when filing a bug.

Nokel81:
Would you like me to recreate this filling then?

mweststrate:
yes, or copy in the template.

Nokel81:
done

mweststrate:
The above example is working as expected if no custom setter is defined. There is a .set method, and you are able to call it successfully, it just throws because the preconditions aren't met, not because the method doesn't exist.

Nokel81:
Sure, but won't it be better for `computed(() => someObservable.get())` to not even allow `.set()` to be called via typescript?

mweststrate:
Well, '.set' is a valid API of `computed`, that the runtime requirements aren't met isn't of typescripts concern. (could do some complicated juggling to make that detectable in TS in cases, but doubt that'd be worth the effort)

Nokel81:
[...]

PR #3783, opened 2023-10-26
fix!: Prevent calling set on the return value of 'computed()'

Nokel81:
BREAKING CHANGE: 'set()' method now only exists if a 'set()' is provided

### Code change checklist

resolves https://github.com/mobxjs/mobx/issues/3767

### DIFF
--- a/docs/computeds.md
+++ b/docs/computeds.md
@@ -236,3 +236,8 @@ It is recommended to set this one to `true` on very expensive computed values. I
 ### `keepAlive`
 
 This avoids suspending computed values when they are not being observed by anything (see the above explanation). Can potentially create memory leaks, similar to the ones discussed for [reactions](reactions.md#always-dispose-of-reactions).
+
+### `set`
+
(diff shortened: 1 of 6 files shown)


docs/computeds.md
```diff
@@ -236,3 +236,8 @@ It is recommended to set this one to `true` on very expensive computed values. I
 ### `keepAlive`
 
 This avoids suspending computed values when they are not being observed by anything (see the above explanation). Can potentially create memory leaks, similar to the ones discussed for [reactions](reactions.md#always-dispose-of-reactions).
+
+### `set`
+
+This optional function allows the returned `IComputedValue<T>` to also act as something that knows how to update its backing data. When not provided, however, this function that is provided will throw an error.
+To help prevent runtime errors like this, especially when refactoring code from `IObservableValue<T>` to `IComputedValue<T>` the typescript types will are set up so that trying to call a function will result in a compile time error when this option is not set.
```
mbest, line 243: Fix a couple errors in the sentence

```suggestion
To help prevent runtime errors like this, especially when refactoring code from `IObservableValue<T>` to `IComputedValue<T>` the typescript types are set up so that trying to call a function will result in a compile-time error when this option is not set.
```

PR #3803, opened 2023-11-25
Some cleanups of old stuff

mweststrate:
### Code change checklist


urugator:
IIRC there were some tests in V4 that were relevant only for `proxy: false` (don't have an analog in in V5) or had different expectations or required a bit different API to implement (`object-api`), would be nice to preserve these.

We should consolidate/rethink the behavior of `resetGlobalState` and it's usage in tests. For some time I've been using it to reset `configure(...)` at the end of a tests without realizing it actually doesn't reset everything. When I found out I went ahead and tried to change it, just to find out this would stop working:
https://github.com/mobxjs/mobx/blob/55260aa158919033e862d219e60eea601a05ac61/packages/mobx/__tests__/v4/mobx4.ts#L3-L8
I also found a test that was passing only because a previous test did not revert the configuration to default.
Ideally we should make sure tests can't have side-effects without the need to manually wrap them like so:
```
try { 
  /* test logic */
} finally { 
  configure({ foo: "You have to know the default for these set of tests" });
}
```

kubk:
The current PR had merge conflicts so I've finished the work here: https://github.com/mobxjs/mobx/pull/4666

Also addressed @urugator feedback - global state resets are now unified in one place instead of being scattered across tests

> IIRC there were some tests in V4 that were relevant only for proxy: false (don't have an analog in V5) 

We'll be dropping the non-proxy support in Mobx 7, so I didn't find anything worth transferring to V5 folder except a few tests. Please let me know if I missed anything

PR #3812, opened 2024-01-04
Prevent Reaction From Keeping a Reference to The OldValue

barroij:
HI Folks

This PR is about `reaction` and memory retainment preventing object beeing garbage collected

For some reason, the `reactionRunner` of a reaction creates a closure around an `oldValue` and so the reference to the `oldValue` is kept untill a newValue is computed.

I can see no reason for keeping such a reference "alive", `oldValue` could just be declared in the scope of the `reactionRunner` as it is the only place where it is used. Doing so, the `oldValue` can be garbage collected if needed.

No additional Unit test is needed for such a change, and there is no backward compatibility issue as it only touches internal/private stuff within a `reaction`.

### DIFF
--- a/.changeset/six-pugs-impress.md
+++ b/.changeset/six-pugs-impress.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Prevent `reaction` from heeping a Reference to the OldValue that would prevent GC.
--- a/packages/mobx/src/api/autorun.ts
+++ b/packages/mobx/src/api/autorun.ts
@@ -139,7 +139,6 @@ export function reaction<T, FireImmediately extends boolean = false>(
     let firstTime = true
     let isScheduled = false
     let value: T
-    let oldValue: T | undefined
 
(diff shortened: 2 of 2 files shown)


urugator:
Did you investigate the behavior when `expression` throws?

barroij:
@urugator
Assuming that `disableErrorBoundaries` is false or that an `onError` errorHandler is provided, if `expression` thows an exception it would be caught within `track`, but `changed` remains unchanged anyway, its value is `false`. So the `effectAction` is not called.
Except if `(firstTime && opts.fireImmediately!)` is true but in this case `oldValue` is `undefined` as it used to be, so the behavior remains the same.

urugator:
Ok, thank you. Could you please provide changeset (`yarn changeset`)?  Otherwise LGTM.

barroij:
I never made a changeset before. I ran `yarn changeset` and commited the generated .md file, is that all ?

urugator:
Yep. Github just seems to be sluggish atm, at least on mind end.

PR #3816, opened 2024-01-17
Fix IReactionDisposer and IIsObservableObject  Typing

barroij:
This is typing only PR.

Current interface definition for `IReactionDisposer` is
```
export interface IReactionDisposer {
    (): void
    $mobx: Reaction
}
```

but that's wrong because because like so typescript considers the property key `$mobx` to be a string, whereas it is a symbol. The correct indexing type is given by 
```
export interface IReactionDisposer {
    (): void
    [$mobx]: Reaction
}
```

[Edit] : same thing for `IIsObservableObject`

### DIFF
--- a/.changeset/seven-rules-add.md
+++ b/.changeset/seven-rules-add.md
@@ -2,4 +2,4 @@
 "mobx": patch
 ---
 
-Fix IReactionDisposer interface definition so that Typescript knows the property key `$mobx` is a symbol and not a string
+Fix `IReactionDisposer` and `IIsObservableObject` interface definition so that Typescript knows the property key `$mobx` is a symbol and not a string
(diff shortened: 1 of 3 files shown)


kubk:
@barroij Thank you. Do you think it makes sense to update `IIsObservableObject` as well?

https://github.com/mobxjs/mobx/blob/13a222ea21bd55b7158d48a998846da608080ded/packages/mobx/src/types/observableobject.ts#L650

barroij:
@kubk You are right, I'm gonna change it too.

barroij:
Note : ficing typing for `IIsObservableObject `  resulted in a test failure (`packages/mobx/src/api/tojs.ts:57:28 - error TS2339: Property 'forEach' does not exist on type 'ArrayLike<string | symbol>'.`).

I had to change the return type of function `ObservableObjectAdministration.ownKeys_`  from `ArrayLike<string | symbol>` to `Array<string | symbol>` to fix it

kubk (approved):

PR #3818, opened 2024-01-19
return action initializer instead of using addInitializer

### LINKED ISSUE #3817: @action field uses addInitializer from decorator context but that does not exist in the (new) spec
**Intended outcome:**

@action decorators should work with fields (lambdas).

**Actual outcome:**

It throws `addInitializer` is not a function. The reason is this code:

https://github.com/mobxjs/mobx/blob/df7e1ebd3404d4d44cd06b7d0a3b27f4c3a6aad1/packages/mobx/src/types/actionannotation.ts#L70-L78

`@action foo = () => {};`

The context passed by the decorator looks like this: `{"kind":"field","name":"addFiles","static":false,"private":false,"metadata":{},"access":{}}`
**Versions**

**Spec reference**
If you look at the class field spec, it says this: https://github.com/tc39/proposal-decorators#class-fields

> Since class fields already return an initializer, they do not receive addInitializer and cannot add additional initialization logic.

The interface of the decorator looks like this:

```ts
type ClassFieldDecorator = (value: undefined, context: {
  kind: "field";
  name: string | symbol;
  access: { get(): unknown, set(value: unknown): void };
  static: boolean;
  private: boolean;
}) => (initialValue: unknown) => unknown | void;
```

So probably instead of using addInitializer the initializer should be returned.

cc @Matchlighter 

_I think_ this issue is not visible when using babel because babel does support addInitializer (in conflict with the spec). But it does appear when compiling with `swc` (which follows the spec)

phiresky:
Fixes #3817 . 

Mobx currently fails to work with spec-compliant decorators (in swc) when using `@action x = () => {}`.

The reason this issue is hard to see is that apparently both tsc and babel do not follow the decorator spec, while swc does. I've opened an issue with typescript as well https://github.com/microsoft/TypeScript/issues/57096 because both their code as well as the types seem to conflict with the spec. 

In any case, this change works both with tsc and with swc (and I assume babel as well).

### DIFF
--- a/packages/mobx/src/types/actionannotation.ts
+++ b/packages/mobx/src/types/actionannotation.ts
@@ -75,10 +75,7 @@ function decorate_20223_(this: Annotation, mthd, context: DecoratorContext) {
 
     // Backwards/Legacy behavior, expects makeObservable(this)
     if (kind == "field") {
-        addInitializer(function () {
-            storeAnnotation(this, name, ann)
-        })
-        return
+        return _createAction
     }
 
     if (kind == "method") {


mweststrate:
Yeah the spec changed a few times in this regard through the different stages. Things is probably ok, but there is a build failure prevent the tests to run, mind updating? (seems just removing an unused variable which our config is being too sensitive about)

mweststrate (changes requested):
`storeAnnotation` is unused now, failing the build

phiresky:
This was a result of the readme of the spec text being wrong / outdated. I've fixed it in swc here: https://github.com/swc-project/swc/issues/8525 and the spec text has also been fixed, so this is no longer necessary (afaik)

Issue #3724, opened 2023-07-17
[performance] autorun with delay deriving state unnecessarily


realyze:
When `autorun` uses a `delay`, we compute derived state eagerly when dependent state is invalidated, i.e. before the autorun actually runs. It's unclear why we need to (or should) do that since computing the dependent state early means there is an opportunity for the state to be mutated yet again before autorun runs which means we potentially need to recompute one more time when autorun runs.

Steps:
1. Create an autorun with delay 1000 ms that observes a computed `c` which depends on observable `o` with value 0
2. At 1500ms set `o` to `1`, this will trigger a recomputation of `c` (but why does it need to? We're still 500ms from `autorun` reading that value)
3. At 1750ms set `o` to `2`, this won't trigger a recomputation of `c` but it means we'll need to recompute at 2000ms when `autorun` runs
4. At 2000ms, `autorun` will run and recompute (with `o === 2`)

Here's a minimal reproducible scenario: https://codesandbox.io/s/mobx-autorun-recalculation-issue-z3vggt

upsuper:
I think this is really a question of semantics, or expectation of the behavior. More specifically, do you expect the `autorun` to be triggered if the computation result is not changed even though its dependency is changed?

For example, in the testcase you provided, if we make the computed do `Math.round(o.get())` rather than returning the value directly, and set it to `0.1` at 1500ms, and `0.9` at 1600ms, would you expect the `autorun` to be triggered at 2500ms (triggered by setting `o` to `0.1` then delay by 1000ms) or at 2600ms (triggered by computation result changing to `1` then delay by 1000ms)? I would probably expect the latter that the `autorun` is triggered only when its direct dependency is actually changing. And to do that, a eager recomputation would be necessary.

upsuper:
I'd note, though, that the `autorun` should never be run at 2000ms as you indicated, because there is nothing triggering it at 1000ms. There is probably some misunderstanding on how `delay` works.

realyze:
Perhaps the issue is better visible with a `reaction` rather than an `autorun`. See e.g. [here](https://codesandbox.io/p/sandbox/gifted-rumple-r4pc68?file=%2Fsrc%2Findex.ts%3A4%2C6).

There's a reaction with a 100ms delay and being a reaction, we never ought to fire its side-effect fn if the expr function result did not change.

[...]

PR #3819, opened 2024-01-24
Derive computed states only when required for delayed Reactions

taj-p:
## Derive computed states only when required for delayed Reactions

Fixes https://github.com/mobxjs/mobx/issues/3724

## Context

We use MobX with React. Due to the above issue, we are triggering unnecessary derivations for computed values (and contexts) for `autorun`s with `delay` or custom schedulers. This means that, at times, the user presses a button on screen and a render is scheduled, but, before the browser paints, it has to compute some derivation that is not required for the render.

Sometimes our derivations are expensive, so we really rely on deferring the (re)calculation of computed values until a delayed `autorun` needs to read them. Moreover, without these changes, we trigger unnescessary (re)calculations prior to when a delayed `autorun` runs.

image

## Changes 

The core change of this PR is to lift the "scheduler" from the `autorun` and `reaction` into `Reaction`. The rationale for this change is (I haven't contributed to this library before, so please scrutinise these reasons closely 🙏 ):

It is within [`Reaction`](https://github.com/mobxjs/mobx/blob/df7e1ebd3404d4d44cd06b7d0a3b27f4c3a6aad1/packages/mobx/src/core/reaction.ts#L53) that `shouldCompute` runs. I.e., the `Reaction` itself determines whether its been invalidated and needs to run its `onInvalidate`, so, I reasoned that it should also be responsible for scheduling that invalidation.

https://github.com/mobxjs/mobx/blob/df7e1ebd3404d4d44cd06b7d0a3b27f4c3a6aad1/packages/mobx/src/core/reaction.ts#L98

## Benefits

Again, I am a MobX noob, so these are the perceived benefits from a novice's POV:

1. We only derive computed values when they're read when they're needed, both reducing the number of times some state is derived (during `delay`) and only deriving them when needed.
2. The `autorun` and `reaction` APIs appear a bit simpler since we've consolidated the scheduling logic to `Reaction`.

## Disadvantages

Are we moving logic to Reaction that shouldn't be moved there? I think it makes sense for a Reaction to schedule its own `onInvalidation`, but I don't have the full context of the library.

### Code change checklist

-   ~[ ] Updated `/docs`. For new functionality, at least `API.md` should be updated~

cc @urugator

### DIFF
--- a/packages/mobx/__tests__/v5/base/autorunAsync.js
+++ b/packages/mobx/__tests__/v5/base/autorunAsync.js
@@ -86,36 +86,50 @@ test("autorun 1", function (done) {
 })
 
 test("autorun re-calculates computed states only when required", function (done) {
-    let _result = null
-    let _cCalcs = 0
+    let autoruns = 0
+    let computedCalcs = 0
 
     const o = mobx.observable.box(0)
     const c = mobx.computed(() => {
-        _cCalcs++
+        computedCalcs++
         return o.get()
     })
 
     mobx.autorun(
         () => {
-            _result = c.get()
+            c.get()
+            autoruns++
         },
-        { delay: 100 }
+        { delay: 1000 }
     )
-
-    setTimeout(
-        mobx.action(() => o.set(1)),
-        30
-    )
-
-    expect(_cCalcs).toEqual(0)
-
-    setTimeout(() => mobx.action(() => o.set(2)), 30)
+    expect(computedCalcs).toBe(0)
 
     setTimeout(() => {
-        expect(_cCalcs).toEqual(1)
-        expect(_result).toEqual(1)
+        // We expect the first run to run the computed
+        expect(autoruns).toBe(1)
+        expect(computedCalcs).toBe(1)
+    }, 1100)
+
+    // We shouldn't eagerly calculate the state before the autorun runs.
+    // because the state can be mutated again before `delay` expires - if that
+    // happens, a re-calculation here would be wasted.
+    setTimeout(() => {
+        o.set(1)
+        expect(computedCalcs).toBe(1)
+    }, 1500)
+
+    // This will re-invalidate the state so we'll need to recalculate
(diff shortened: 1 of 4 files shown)


packages/mobx/src/core/reaction.ts
```diff
@@ -90,6 +94,16 @@ export class Reaction implements IDerivation, IReactionPublic {
...
+        if (!this.scheduledRunReaction_) {
+            this.scheduledRunReaction_ = true
+            this.scheduler(() => {
+                this.scheduledRunReaction_ = false
+                this.runReactionImpl()
+            })
+        }
+    }
+
+    private runReactionImpl() {
```
taj-p, line 106: These are the core changes - `Reaction` dictates when it should `runReactionImpl` based on the passed in scheduler.

packages/mobx/__tests__/v5/base/autorunAsync.js
```diff
@@ -85,6 +85,53 @@ test("autorun 1", function (done) {
...
+        o.set(3)
+        expect(computedCalcs).toBe(1)
+    }, 900)
+
+    setTimeout(() => {
+        expect(autoruns).toBe(2)
+        expect(computedCalcs).toBe(2)
+        done()
+    }, 1600)
+})
```
taj-p, line 133: Thanks @realyze - your [reproduction](https://codesandbox.io/p/sandbox/mobx-autorun-recalculation-issue-z3vggt?file=%2Fpackage.json%3A17%2C2) helped craft this test

taj-p: For reference to reviewers: the same test fails on the latest `main`:

image

taj-p:
As per @upsuper's [comments](https://github.com/mobxjs/mobx/issues/3724#issuecomment-1907424738), I've realised this implementation won't work exactly as we like. Will make it a draft for the time being

PR #3822, opened 2024-01-28
fix: Update the links to enabling-decorators

pranavrajs:
The link to decorators in the [installation](https://mobx.js.org/installation.html) documentation is incorrect. It generates the link as `https://mobx.js.org/enabling-decorators.md` instead of `https://mobx.js.org/enabling-decorators.html`. This PR fixes it.

Note: I've tested this locally, seems to be working fine.


kubk:
@pranavrajs Thank you

PR #3831, opened 2024-02-15
fix: ensure observer component name is only set when configurable

kitsuned:
this patch addresses issue with environments where function name property is not configurable, notably pre-ES2015.

the same condition exists in MobX: [core/action.ts](https://github.com/mobxjs/mobx/blob/main/packages/mobx/src/core/action.ts#L54).

regarding tests, I couldn't find a feasible way to seal a function, which was also discussed in #2262.

<details>

```shell
$ podman run --rm node:0.12 node -p 'Object.getOwnPropertyDescriptor(function() {}, "name")'
{ value: '',
  writable: false,
  enumerable: false,
  configurable: false }

$ podman run --rm node:0.12 node -e 'Object.defineProperty(function() {}, "name", { value: "foobar" })'
[eval]:1
Object.defineProperty(function() {}, "name", { value: "foobar" })
       ^
TypeError: Cannot redefine property: name
    at Function.defineProperty (native)
    at [eval]:1:8
    at Object.exports.runInThisContext (vm.js:74:17)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:460:26)
    at evalScript (node.js:431:25)
    at startup (node.js:90:7)
    at node.js:814:3
```

</details>

### Code change checklist

### DIFF
--- a/.changeset/wicked-cheetahs-build.md
+++ b/.changeset/wicked-cheetahs-build.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+fix: ensure observer component name is only set when configurable
--- a/packages/mobx-react-lite/src/observer.ts
+++ b/packages/mobx-react-lite/src/observer.ts
@@ -6,6 +6,9 @@ import { useObserver } from "./useObserver"
 let warnObserverOptionsDeprecated = true
 
 const hasSymbol = typeof Symbol === "function" && Symbol.for
+const isFunctionNameConfigurable =
+    Object.getOwnPropertyDescriptor(() => {}, "name")?.configurable ?? false
+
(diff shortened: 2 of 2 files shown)


shadowcat-mst:
Coming a little late to the party, but in every engine I've worked with of late one can also do:

    let namedThing = { [desiredName]: function (...) { ... } }[desiredName];
    
 and it shows up with the `.name` property already set. I don't have anything pre-ES2015 lying around to test against, so I've no idea if it successfully sets the name there, sorry; I mentioned it merely because depending on the situation one might find it preferable to applying `Object.defineProperty` after the fact.
 
 (the word 'might' in the above is why I'm leaving a comment rather than sending a PR; the code might be trivial but the decision around trade-offs is less so and I don't believe I'm the right person to make such a call)

PR #3841, opened 2024-03-07
using ts disposable

vovsemenv:
implements https://github.com/mobxjs/mobx/discussions/3840

### Code change checklist

### DIFF
--- a/.changeset/sixty-frogs-doubt.md
+++ b/.changeset/sixty-frogs-doubt.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
+add `_action` helper to use with ts disposables
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -142,6 +142,21 @@ runInAction(() => {
 })
 ```
 
+<!--using _ = _action();-->
+
+```javascript
+import { observable, runInAction } from "mobx"
+
+const state = observable({ value: 0 })
+
+{
+  using _ = _action();
+  state.value++
+  state.value++
+}
+
+```
+
 <!--END_DOCUSAURUS_CODE_TABS-->
 
 ## Wrapping functions using `action`
@@ -218,6 +233,16 @@ Usage:
 -   `runInAction(fn)`
 
 Use this utility to create a temporary action that is immediately invoked. Can be useful in asynchronous processes.
+Check out the [above code block](#examples) for an example.
+
+## `_action()`
+
+Usage:
+
+-   ` using _ = _action();`
+
+Use this utility to create a temporary action. Can be useful in asynchronous processes.
+Behaves the same as `runInAction` but ends action after leaving current block scope.
 Check out the [above code block](#examples) for an example.
 
 ## Actions and inheritance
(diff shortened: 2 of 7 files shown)


kubk:
@vovsemenv Thank you, could you add some unit tests?

kubk:
@mweststrate @urugator I think having a utility like this is a good option, as the other ways to modify code asynchronously are far from ideal:
- `﻿runInAction` with `﻿async`/`﻿await` increases code nesting.
- Mobx flow is great, but it requires a hack to have proper type support. https://github.com/mobxjs/mobx/discussions/3195

vovsemenv:
@kubk added test, please approve ci run

urugator:
I am not strictly opposed, but I think it has some negatives, while I am not sure about benefits.

1) Second way to do the same thing.
1) How will it play with eslint? I would suspect `using` becomes a subject of `no-unused-vars` or `no-unused-resource` will be introduced. This would require an extra convention and configuration.
1) It feels like bending something that was designed for something else. The code is literally saying "use this resource" and then you're not using it. Could be wrong about this, I could imagine resource that only has a side effect like spinning up a remote server. The point is, it's intended for resource disposal, not for intercepting.
1) It can't intercept error or return value. We don't need that now, but something to keep in mind.

> ﻿runInAction with ﻿async/﻿await increases code nesting.

Isn't it almost the same?
```ts
async action() {
  using _ = _action();  
  this.x = 1
  await fetch()
  {
    using _ = _action();
    this.x = 1
  } 
  await fetch()
  {
    using _ = _action();
    this.x = 1;
  }
}
```
Am I missing something?

kubk:
@urugator Valid points have been brought up 👍. Yes, the requirement of having a variable leads to issues with ESLint. 

@vovsemenv are you interested in moving to userland as a separate NPM package? It seems like the solution has its own drawbacks and can't be added to the core.

mweststrate:
It is kinda cool! But a crucial design question is still open I think; transactions in MobX are always concluded synchronously (otherwise it might interfere with concurrent transactions or 'block' updating entirely). So given something like:

```javacript
const counter = observable({count : 1, inc() { this.count++; } })

autorun(() => console.log(counter.count))

things();

async function things() {
  using _ = action();
  counter.inc()
  counter.inc()

  await sleep(100)
  console.log("slept")

  counter.inc()
  counter.inc()
}
```

I'd expect the output to be:
```
2
slept
4
```

But I think that will be the case with the current implementation, as the updates will only appear entirely at the end, and the output is just `slept` and `4`?

The reason why that is important, is that a) transactions are global in MobX and b) we don't want to block things like react components updating in an async way, as in that case the app might freeze if you are doing a async transaction that contains a network request or so.

urugator:
> as the updates will only appear entirely at the end

You're correct. I thought there is a difference between `using` and `await using`, but both are disposed at the end of the block. So the example I posted above would have to look like this:
```ts
async action() {
  {
    using _ = _action();
    this.x = 1
  } 
  await fetch()
  {
    using _ = _action();
    this.x = 1
  } 
  await fetch()
  {
    using _ = _action();
    this.x = 1;
  }
}
```

mweststrate:
Closing this PR as there seems sadly no way to make this work for async boundaries, and without it will just be another way (not particularly easier or shorter) of doing `runInAction`. Thanks for trying this cool idea though!

PR #3853, opened 2024-04-02
Supporting a new set methods

### LINKED ISSUE #3850: Supporting of new Set methods in observable Sets
**Intended outcome:**
According to a [New Set methods proposal](/tc39/proposal-set-methods) a Set class will have a bunch of new methods (more on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#instance_methods)). This proposal currently at Stage 3 therefore it means it is becoming part of ECMAScript standart

**Current browser's support**

| Method              	| Firefox (Nigthly)	| Chrome (> 122)	| Safari (> 17)	|
|---------------------	|---------	|--------	|--------	|
| intersection        	| ⏳       	| ✔️      	| ✔️      	|
| union               	| ⏳       	| ✔️      	| ✔️      	|
| difference          	| ⏳       	| ✔️      	| ✔️      	|
| symmetricDifference 	| ⏳       	| ✔️      	| ✔️      	|
| isSubsetOf          	| ⏳       	| ✔️      	| ✔️      	|
| isSupersetOf        	| ⏳       	| ✔️      	| ✔️      	|
| isDisjointFrom      	| ⏳       	| ✔️      	| ✔️      	|

I've created a [sandbox](https://codesandbox.io/p/sandbox/goofy-pond-555v98?file=%2Fsrc%2Fstore.js%3A1%2C51&layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clu86kzd90006356iqlldz1ng%2522%252C%2522sizes%2522%253A%255B100%252C0%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clu86kzd90002356itnmgf0n7%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clu86kzd90003356iqjyl1ghj%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clu86kzd90005356iou9j8qt0%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B50%252C50%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clu86kzd90002356itnmgf0n7%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clu86kzd80001356ilmmkycp0%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252Fsrc%252Findex.js%2522%252C%2522state%2522%253A%2522IDLE%2522%252C%2522initialSelections%2522%253A%255B%257B%2522startLineNumber%2522%253A9%252C%2522startColumn%2522%253A1%252C%2522endLineNumber%2522%253A9%252C%2522endColumn%2522%253A1%257D%255D%257D%252C%257B%2522id%2522%253A%2522clu872a6k0002356ig0tshaxx%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522initialSelections%2522%253A%255B%257B%2522startLineNumber%2522%253A1%252C%2522startColumn%2522%253A51%252C%2522endLineNumber%2522%253A1%252C%2522endColumn%2522%253A51%257D%255D%252C%2522filepath%2522%253A%2522%252Fsrc%252Fstore.js%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%255D%252C%2522id%2522%253A%2522clu86kzd90002356itnmgf0n7%2522%252C%2522activeTabId%2522%253A%2522clu872a6k0002356ig0tshaxx%2522%257D%252C%2522clu86kzd90005356iou9j8qt0%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clu86kzd90004356iipgkjw65%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522UNASSIGNED_PORT%2522%252C%2522port%2522%253A0%252C%2522path%2522%253A%2522%252F%2522%257D%255D%252C%2522id%2522%253A%2522clu86kzd90005356iou9j8qt0%2522%252C%2522activeTabId%2522%253A%2522clu86kzd90004356iipgkjw65%2522%257D%252C%2522clu86kzd90003356iqjyl1ghj%2522%253A%257B%2522tabs%2522%253A%255B%255D%252C%2522id%2522%253A%2522clu86kzd90003356iqjyl1ghj%2522%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Afalse%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D) with example of code that includes two Sets as store's fields. One of them is observable other one is not.

**Actual outcome:**
For example if log a `store.reactiveSet.intersection` we will see an `undefined` in the console. If log a `store.nonReactiveSet.intersection` we will see a `nonReactiveSet ƒ symmetricDifference() { [native code] }` (there I used Chrome 123)
Obviously for other Set's methods result is the same:
![image](https://github.com/mobxjs/mobx/assets/1235286/958d9714-dec7-4bd0-b64c-4bf06c921c16)


Run code from provided sandbox. Look into browser's console

**Versions**

Mobx 6.12.1

Also I propose made this as part of Mobx 7 roadmap #3796

inoyakaigor:
### Code change checklist

Fixes #3850 

These changes depends on microsoft/TypeScript/pull/57230 . According TS iteration plan (microsoft/TypeScript/issues/57475) it should be landed in TS 5.5 (in June'24)

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,5 +74,6 @@
         "hooks": {
             "pre-commit": "pretty-quick --staged"
         }
-    }
+    },
+    "packageManager": "yarn@1.22.19+sha256.732620bac8b1690d507274f025f3c6cfdc3627a84d9642e38a07452cc00e0f2e"
 }
--- a/packages/mobx/src/types/observableset.ts
+++ b/packages/mobx/src/types/observableset.ts
@@ -243,6 +243,44 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
         } as any)
     }
 
+    intersection(otherSet: Set<T>): IterableIterator<T> {
+        this.atom_.reportObserved()
+        const self = this
+        let nextIndex = 0
+        const observableValues = Array.from(this.data_.intersection(otherSet))
+        return makeIterable<T>({
+            next() {
+                return nextIndex < observableValues.length
+                    ? { value: self.dehanceValue_(observableValues[nextIndex++]), done: false }
+                    : { done: true }
+            }
+        } as any)
+    }
+
+    union(): IterableIterator<T> {
+        // TODO: implement
+    }
+
+    difference(): IterableIterator<T> {
+        // TODO: implement
+    }
+
+    symmetricDifference(): IterableIterator<T> {
+        // TODO: implement
+    }
+
+    isSubsetOf(): IterableIterator<T> {
+        // TODO: implement
+    }
+
+    isSupersetOf(): IterableIterator<T> {
+        // TODO: implement
+    }
+
+    isDisjointFrom(): IterableIterator<T> {
+        // TODO: implement
+    }
+
     replace(other: ObservableSet<T> | IObservableSetInitialValues<T>): ObservableSet<T> {
         if (isObservableSet(other)) {
             other = new Set(other)


inoyakaigor:
@mweststrate at the [issue comment](https://github.com/mobxjs/mobx/issues/3850#issuecomment-2021398319) you've mentioned that it should be enough to call an original Set.prototype method so I've copypasted a code from a `Observableset.values()` method. Is it correct? Should I add / remove / rewrite something?

inoyakaigor:
Also there is a small change in the `package.json`. This field was added by a Node.js's manager of package managers [Corepack](https://nodejs.org/api/corepack.html). It is because I've ran into a problem at initial setup the project because for developing I have to use yarn@1 but in my daily job I'm using yarn@4. It isn't a breaking change.

urugator:
> @mweststrate at the https://github.com/mobxjs/mobx/issues/3850#issuecomment-2021398319 you've mentioned that it should be enough to call an original Set.prototype method

I think the idea was something like `Set.prototype.intersect.call(observaleSetOrJustIterator, otherSet)`, but I suspect it won't be possible:

> intersection() accepts [set-like](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#set-like_objects) objects as the other parameter. It requires [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) to be an actual [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instance, because it directly retrieves the underlying data stored in this without invoking any user code.

>The set-like protocol invokes the keys() method instead of [[@@iterator]()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/@@iterator) to produce elements. This is to make maps valid set-like objects, because for maps, the iterator produces entries but the has() method takes keys.

So you will probably have to create intermediate set with dehanced values and intersect it instead.

urugator:
But in case of commutative operations (like intersection), it should be possible to call it other way around:
```ts
intersection(otherSet: Set<T>): SetLike<T> {
   return otherSet.intersect(this);
}
```

urugator:
If the argument is actual Set (instead of just being set-like) and if the operation is commutative (like intersection), it can be optimized by calling the operation on the argument:
```ts
intersection(otherSet: SetLike<T>): SetLike<T> {
   if (typeof otherSet.intersect === 'function') {
      return otherSet.intersect(this);
   } else {
      const dehancedSet = new Set(this);
      return dehancedSet.intersect(otherSet);
   }   
}
```

urugator:
> Should I keep an makeIterable code in the new functions? I mean this one:

No, it's  supposed to return new `Set`, not an iterator.
Also `const dehancedSet = new Set(this);` already invokes the iterator.

> Also am I understands correctly that code delaying possible side effects as maximum as posssible when outer code reading the ObservableSet values?

I don't understand the question.

>  Wouldn't it be better to change a typing for a private data_ field to the same generic type as in class declaration

I think it needs to be `any` because of the enhancer possibly changing the type.

> Also I just realised that there also should be added [Object.groupBy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) and [Map.groupBy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/groupBy)

Afaik we don't have to do anything to support these.

inoyakaigor:
> Afaik we don't have to do anything to support these.

I think this methods should trigger `reportObserved()` method at least

package.json
```diff
@@ -68,11 +68,12 @@
...
-        "typescript": "^5.0.2"
+        "typescript": "^5.5.0-dev.20240415"
     },
     "husky": {
         "hooks": {
             "pre-commit": "pretty-quick --staged"
         }
-    }
+    },
+    "packageManager": "yarn@1.22.19+sha256.732620bac8b1690d507274f025f3c6cfdc3627a84d9642e38a07452cc00e0f2e"
```
urugator, line 78: TODO remove before merge

package.json
```diff
@@ -68,11 +68,12 @@
         "tape": "^5.0.1",
         "ts-jest": "^29.0.5",
         "tsdx": "^0.14.1",
-        "typescript": "^5.0.2"
+        "typescript": "^5.5.0-dev.20240415"
```
urugator, line 71: TODO consolidate before merge

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,62 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
         } as any)
     }
 
+    intersection<U>(otherSet: ReadonlySetLike<U>): Set<T & U> {
+        this.atom_.reportObserved()
```
urugator, line 247: `this.atom_.reportObserved()` is only needed in the first branch:
```ts
if (typeof otherSet.intersection !== "function") {
   this.atom_.reportObserved()
   return otherSet.intersection(this.data_)
} else {
  const dehancedSet = new Set(this) // this should invoke _values(), which reports already
   return dehancedSet.intersection(otherSet)
}
```
`new Set(this)` already reports.

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,62 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+
+        if (typeof otherSet.intersection !== "function") {
```
urugator, line 249: Why `!==` => `===`

urugator:
Don't forget to provide tests please. We want to test 2 things:

1. It does what it should. Use native `Map` for `set-like` arguments. Check results of our (observable) method calls against results of native (non-observable) method calls.
2. It's reactive.

> I think this methods should trigger reportObserved() method at least

Feel free write tests for these as well and we will decide how to proceed based on the results 😉.

docs/installation.md
```diff
@@ -35,22 +35,29 @@ When using MobX with TypeScript or Babel, and you plan to use classes; make sure
...
+    For verification insert this piece of code at the beginning of your sources (eg. `index.js`)
+
 ```javascript
-if (!new class { x }().hasOwnProperty('x')) throw new Error('Transpiler is not configured correctly');
+if (
+    !new (class {
+        x
+    })().hasOwnProperty("x")
+)
+    throw new Error("Transpiler is not configured correctly")
```
kubk, line 57: @inoyakaigor Do we really need to change this file in the PR? It seems like before this script was more compact.

package.json
```diff
@@ -68,11 +68,12 @@
...
         "ts-jest": "^29.0.5",
         "tsdx": "^0.14.1",
-        "typescript": "^5.0.2"
+        "typescript": "5.5.0-dev.20240603"
     },
     "husky": {
         "hooks": {
             "pre-commit": "pretty-quick --staged"
         }
-    }
```
kubk, line 77: The 5.5.0 is finally out, so we can update it: https://github.com/microsoft/TypeScript/releases

package.json
```diff
@@ -68,11 +68,12 @@
         "tape": "^5.0.1",
         "ts-jest": "^29.0.5",
         "tsdx": "^0.14.1",
-        "typescript": "^5.0.2"
+        "typescript": "5.5.0-dev.20240603"
```
kubk, line 71: The 5.5.0 is finally out, so we can update it: https://github.com/microsoft/TypeScript/releases

docs/installation.md
inoyakaigor: @kubk seems to be automatic changes. Prettier or VSCode built-in formatter

package.json
inoyakaigor: Updated

docs/installation.md
inoyakaigor: Reverted

package.json
inoyakaigor: Removed

package.json
inoyakaigor: Removed

kubk (approved):

tsconfig.json
```diff
@@ -3,7 +3,7 @@
         "target": "es6",
         "module": "esnext",
         "moduleResolution": "node",
-        "lib": ["es6"],
+        "lib": ["es6", "esnext"],
```
urugator, line 6: Isn't `es6` already part of `esnext`?

inoyakaigor: Yes, it is. Removed the `es6`

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
         } as any)
     }
 
+    intersection<U>(otherSet: ReadonlySetLike<U> & Partial<Set<U>>): Set<T & U> {
```
urugator, line 246: Why the `& Partial<Set<U>>` ?

https://github.com/microsoft/TypeScript/blob/main/src/lib/esnext.collection.d.ts#L28-L57

urugator: I see. Change it to `otherSet: ReadonlySetLike<U> | Set<U>` and the check to:
https://github.com/mobxjs/mobx/blob/5c8066d0ff6552bd15ab069a655c872b7da3395e/packages/mobx/src/utils/utils.ts#L154

Or are there any problems with that?

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+        if (typeof otherSet.symmetricDifference === "function") {
+            this.atom_.reportObserved()
+            return otherSet.symmetricDifference(this.data_)
+        } else {
+            const dehancedSet = new Set(this)
+            return dehancedSet.symmetricDifference(otherSet)
+        }
+    }
+
+    isSubsetOf(otherSet: ReadonlySetLike<unknown> & Partial<Set<unknown>>): boolean {
```
urugator, line 280: Remove ` & Partial<Set<T>>`

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+        return new Set(this).isSubsetOf(otherSet)
+    }
+
+    isSupersetOf(otherSet: ReadonlySetLike<unknown> & Partial<Set<unknown>>): boolean {
```
urugator, line 284: Remove ` & Partial<Set<T>>`

packages/mobx/src/types/observableset.ts
inoyakaigor: Changed type to `| Set<>` and if branch check to `isES6Set`. There is no problems

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+        if (isES6Set(otherSet)) {
+            this.atom_.reportObserved()
+            return otherSet.union(this.data_)
+        } else {
+            const dehancedSet = new Set(this)
+            return dehancedSet.union(otherSet)
+        }
+    }
+
+    difference<U>(otherSet: ReadonlySetLike<U> & Set<T>): Set<T> {
```
urugator, line 266: ` | Set<U>`

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+        return new Set(this).difference(otherSet)
+    }
+
+    symmetricDifference<U>(otherSet: ReadonlySetLike<U> & Set<U>): Set<T | U> {
```
urugator, line 270: ` | Set<U>`

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+
+    isSubsetOf(otherSet: ReadonlySetLike<unknown> & Set<unknown>): boolean {
+        return new Set(this).isSubsetOf(otherSet)
+    }
+
+    isSupersetOf(otherSet: ReadonlySetLike<unknown> & Set<unknown>): boolean {
+        return new Set(this).isSupersetOf(otherSet)
+    }
+
+    isDisjointFrom(otherSet: ReadonlySetLike<unknown> & Set<unknown>): boolean {
```
urugator, line 288: ` | Set<unknown>`

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+        if (isES6Set(otherSet)) {
+            this.atom_.reportObserved()
+            return otherSet.symmetricDifference(this.data_)
+        } else {
+            const dehancedSet = new Set(this)
+            return dehancedSet.symmetricDifference(otherSet)
+        }
+    }
+
+    isSubsetOf(otherSet: ReadonlySetLike<unknown> & Set<unknown>): boolean {
```
urugator, line 280: Remove `& Set<unknown>`

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+        return new Set(this).isSubsetOf(otherSet)
+    }
+
+    isSupersetOf(otherSet: ReadonlySetLike<unknown> & Set<unknown>): boolean {
```
urugator, line 284: Remove `& Set<unknown>`

packages/mobx/src/types/observableset.ts
urugator, line 266: Remove `& Set<unknown>`

packages/mobx/src/types/observableset.ts
```diff
@@ -243,6 +243,58 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
         } as any)
     }
 
+    intersection<U>(otherSet: ReadonlySetLike<U> | Set<U>): Set<T & U> {
+        if (isES6Set(otherSet)) {
+            this.atom_.reportObserved()
+            return otherSet.intersection(this.data_)
```
urugator, line 249: This isn't correct, it won't work with custom enhancer. Was there any problem with the code I've proposed originally? That is:
```ts
if (isES6Set(otherSet)) {
   return otherSet.intersection(this)
} else {
  // ...
}
```

inoyakaigor: I've checked twice where you mentioned about deleting `reportObserved()` but didn't find. Maybe you're assumed this but I didn't get it. So anyway removing the `reportObserved()` throwing errors:

image

urugator: > where you mentioned about deleting reportObserved() but didn't find.

I didn't mention deleting `reportObserved` anywhere, because I didn't propose adding it in the first place:
https://github.com/mobxjs/mobx/pull/3853#issuecomment-2033993906

> So anyway removing the reportObserved() throwing errors:

Did you just removed `reportObserved` or did you also replaced `_data` with `this` as suggested?

inoyakaigor: Oh, I've tought you're mentioned it in your latest messages. Totally forgot about this one! In this case everything is okay

urugator:
Thank you. However we forgot the changeset... can you provide it in another PR? `yarn changeset`

urugator:
and build is not passing... https://github.com/mobxjs/mobx/actions/runs/9761029690/job/26941159292#step:6:232

inoyakaigor:
@urugator it's because of Node's version 
image
Only v22 (maybe also v21 I didn't check) supports Set methods

inoyakaigor:
@urugator I've made new PR #3898

Issue #3876, opened 2024-05-17
[eslint-plugin-mobx] `mobx/exhaustive-make-observable` autofix behavior should be inversed


kade-robertson:
**Intended outcome:**

As per the documentation of `makeObservable`, it is expected that if you do not currently have an annotation for a member, it is not affected. 

I would expect that an autofix for `mobx/exhaustive-make-observable` would put a value for all missing members that should be listed, but have the annotation set to `false` to maintain the behaviour from before. This seems especially weird because the warning message itself tells you to mark fields as false: "Missing annotation for \<members\>. To exclude a field, use `false` as annotation"

**Actual outcome:**

The `mobx/exhaustive-make-observable` autofix annotates all missing members with `true`, changing the behaviour of those members.

**How to reproduce the issue:**

You can fork this Devbox, run `pnpm lint` to see the warning with an autofix option, and `pnpm lint:fix` to see what the autofix does.

https://codesandbox.io/p/devbox/h7st47?file=%2Findex.js%3A14%2C20

**Versions**

`eslint-plugin-mobx` v0.0.9.

urugator:
The biggest problem with `makeObservable` is that it's easy to forget making something observable. That's what this eslint rule solves. The autofix is intended to provide the conviniency of `makeAutoObservable`.

>  "Missing annotation for <members>. To exclude a field, use false as annotation"

It' suspposed to say: *Hey, you probably forgot to annotate a field, but if this is actually intended and you don't want all fields observable, you don't have to disable this rule, you can just use `false` as annotation.*

If you're `makeObservable` user, it's not immediately obvious you can use `false` as annotation, because it does nothing except for silencing this rule. Therefore the hint.

kade-robertson:
I agree the eslint rule solves the problem it sets out to solve. I disagree with the autofix. You have no way of knowing what the intent of the missing field is, just that the user _probably_ didn't mean it. But they very well might have! 

[...]

PR #3877, opened 2024-05-17
fix(eslint-plugin-mobx)!: invert exhausive-make-observable autofix

kade-robertson:
To avoid changing behaviour of existing logic, the autofix for this rule should specify a `false` annotation to avoid changing code behaviour.

### Code change checklist

-   [x] Added/updated unit tests -- There were none before and this change is pretty minor.
-   [x] Updated `/docs`. For new functionality, at least `API.md` should be updated (Just updated the README.md in this case)
-   [x] Verified that there is no significant performance drop (`yarn mobx test:performance`) -- I don't believe this is relevant for this change

### DIFF
--- a/.changeset/twenty-coins-matter.md
+++ b/.changeset/twenty-coins-matter.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-mobx": minor
+---
+
+The exhaustive-make-observable autofix has been changed to specify a false annotation instead of true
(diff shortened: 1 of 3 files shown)


urugator:
Hi, please see
https://github.com/mobxjs/mobx/issues/3876#issuecomment-2119001806
Feel free to improve the wording of error message.

PR #3880, opened 2024-05-21
Shrink ComputedValue using a bitfield

peterm-canva:
### Code change checklist

-   [] N/A: Added/updated unit tests
-   [] N/A: Updated `/docs`. For new functionality, at least `API.md` should be updated

`ComputedValue` is currently 112 bytes in 64-bit chromium browsers, and it's not unreasonable to see 10s or 100s of thousands of these on very large apps. This makes this class IMO a reasonable target for memory optimization, given small changes can give big wins in absolute memory.

This change takes four boolean fields and stores them in one bitfield, adding getters and setters to keep the API to access them the same.

This shrinks 112 bytes down to 100 (verified in devtools) by removing 4x 4 byte booleans and replacing them with 1x 4 byte integer in v8. In Safari the bools take 8 bytes so this should save 3x 8 bytes (but I didn't check).

There will be an increased overhead to get/set the boolean values now given the bit-shifting required. I don't see any regressions running `yarn test:performance`, but I'm happy to investigate further if this is a concern. I don't think these booleans are in any really performance critical paths. If they are, I'm pretty confident v8 at least will optimize the bit-shifting well, and only add a few machine instructions in optimized code.

Regarding the code, using a bitfield could be done in many different ways, so take what I have here as one suggestion, I'm happy to refactor or make more general. I'd like to avoid introducing a bitfield class, because the allocation of a new object per `ComputedValue` instance will cancel out the memory savings. Better to keep the bitfield as a direct member of `ComputedValue`.

If we can reach a suitable solution here, I'd like to do the same thing to `ObservableValue`, `Atom`, `Reaction` and `ObservableObjectAdministration` which from a brief glance can all save 12-ish bytes as well.

### DIFF
--- a/packages/mobx/__tests__/v5/base/errorhandling.js
+++ b/packages/mobx/__tests__/v5/base/errorhandling.js
@@ -4,7 +4,7 @@ const utils = require("../../v5/utils/test-utils")
 
 const { observable, computed, $mobx, autorun } = mobx
 
-const voidObserver = function () { }
+const voidObserver = function () {}
 
 function checkGlobalState() {
     const gs = mobx._getGlobalState()
@@ -148,7 +148,7 @@ test("deny state changes in views", function () {
 
     m.reaction(
         () => z.get(),
-        () => { }
+        () => {}
     )
     expect(
         utils.grabConsole(() => {
@@ -194,7 +194,7 @@ test("deny array change in view", function (done) {
     }).not.toThrow()
     m.reaction(
         () => z.length,
-        () => { }
+        () => {}
     )
 
     expect(
@@ -483,7 +483,7 @@ test("peeking inside erroring computed value doesn't bork (global) state", () =>
         b.get()
     }).toThrowError(/chocolademelk/)
 
-    expect(a.isPendingUnobservation_).toBe(false)
+    expect(a.isPendingUnobservation).toBe(false)
     expect(a.observers_.size).toBe(0)
     expect(a.diffValue_).toBe(0)
     expect(a.lowestObserverState_).toBe(-1)
@@ -493,7 +493,7 @@ test("peeking inside erroring computed value doesn't bork (global) state", () =>
     expect(b.dependenciesState_).toBe(-1) // NOT_TRACKING
     expect(b.observing_.length).toBe(0)
     expect(b.newObserving_).toBe(null)
-    expect(b.isPendingUnobservation_).toBe(false)
+    expect(b.isPendingUnobservation).toBe(false)
     expect(b.observers_.size).toBe(0)
     expect(b.diffValue_).toBe(0)
     expect(b.lowestObserverState_).toBe(0)
@@ -501,7 +501,7 @@ test("peeking inside erroring computed value doesn't bork (global) state", () =>
     expect(() => {
         b.get()
     }).toThrowError(/chocolademelk/)
-    expect(b.isComputing_).toBe(false)
+    expect(b.isComputing).toBe(false)
 
     checkGlobalState()
 })
@@ -521,7 +521,7 @@ describe("peeking inside autorun doesn't bork (global) state", () => {
     expect(r).toBe(1)
 
     test("it should update correctly initially", () => {
-        expect(a.isPendingUnobservation_).toBe(false)
+        expect(a.isPendingUnobservation).toBe(false)
         expect(a.observers_.size).toBe(1)
         expect(a.diffValue_).toBe(0)
         expect(a.lowestObserverState_).toBe(-1)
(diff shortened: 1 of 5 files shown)


peterm-canva:
Not that familiar with changesets but this is an implementation detail so I think doesn't need a changeset. Otherwise I'll add one.

peterm-canva:
@urugator are you able to help with review? Many thanks

packages/mobx/src/core/computedvalue.ts
```diff
@@ -90,8 +100,13 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
     protected value_: T | undefined | CaughtException = new CaughtException(null)
     name_: string
     triggeredBy_?: string
-    isComputing_: boolean = false // to check for cycles
-    isRunningSetter_: boolean = false
+
+    private static readonly isComputingIndex = 0
+    private static readonly isRunningSetterIndex = 1
+    private static readonly isBeingObservedIndex = 2
+    private static readonly isPendingUnobservationIndex = 3
```
zhantx, line 107: if we use value 0 1 2 4 directly, does it make it slightly faster by skipping the bit shifting?

peterm-canva: Done! Yes I think that's neater thanks

peterm-canva:
Added a patchset because @taj-p mentioned I will most likely need that so we can uprev to that version.

packages/mobx/src/core/computedvalue.ts
```diff
@@ -90,8 +100,13 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
...
     name_: string
     triggeredBy_?: string
-    isComputing_: boolean = false // to check for cycles
-    isRunningSetter_: boolean = false
+
+    private static readonly isComputingMask = 0b0001
+    private static readonly isRunningSetterMask = 0b0010
+    private static readonly isBeingObservedMask = 0b0100
+    private static readonly isPendingUnobservationMask = 0b1000
+    private bitField = 0b0000
```
mweststrate, line 108: Super NIT: `flags` might read slightly nicer?

mweststrate (approved):
This looks great, thanks for adding this! If any perf regressions are reported, I expect we could loose the getter/setters, but hopefully JIT actually erases those.

mweststrate:
Thanks for this PR, this looks great! 

> If we can reach a suitable solution here, I'd like to do the same thing to ObservableValue, Atom, Reaction and ObservableObjectAdministration which from a brief glance can all save 12-ish bytes as well.

I imagine it'd be good to have the release soak for a few weeks, if it doesn't raise any issues sounds good :)

zhantx (approved):

peterm-canva:
> I imagine it'd be good to have the release soak for a few weeks, if it doesn't raise any issues sounds good :)

Sounds good, I'll get the next PR ready and we can hold off a few weeks.

peterm-canva:
There's an issue with this for the production build. The getters/setters are not renamed by minification, but the code that accesses them does get the keys renamed. This means they always come back falsey. Working on that

packages/mobx/__tests__/v5/base/errorhandling.js
```diff
@@ -4,7 +4,7 @@ const utils = require("../../v5/utils/test-utils")
 
 const { observable, computed, $mobx, autorun } = mobx
 
-const voidObserver = function () { }
```
peterm-canva, line 7: I didn't change this on purpose, I think the formatter is just getting excited because I changed other things in this file

packages/mobx/src/core/atom.ts
```diff
@@ -22,8 +22,8 @@ export interface IAtom extends IObservable {
 }
 
 export class Atom implements IAtom {
-    isPendingUnobservation_ = false // for effective unobserving. BaseAtom has true, for extra optimization, so its onBecomeUnobserved never gets called, because it's not needed
```
peterm-canva, line 25: I changed these so that they are no longer mangled (don't end with `_`). This is so that they still match the new getter/setter names when accessed. Unfortunately the code generated in the final bundle for getters/setters has their name in quotes e.g. "isPendingUnobservation" and name mangling is not smart enough to figure out these are being used as property names.

peterm-canva:
OK, good to go. I had to change names so they aren't mangled any more. See comment in atom.ts.

peterm-canva:
Can we land this one?

mweststrate:
Sorry, missed the update! Merged!

mweststrate:
Should soon be available as 6.12.4

@peterm-canva since you are validating performance, I am wondering if you are interested in checking if https://github.com/mobxjs/mobx/pull/3884 also gains these positive results for you?

PR #3881, opened 2024-05-21
feat(eslint-plugin-mobx): configurable default autofix annotation for makeObservable

### LINKED ISSUE #3876: [eslint-plugin-mobx] `mobx/exhaustive-make-observable` autofix behavior should be inversed

kade-robertson:
Adds an option for the `mobx/exhaustive-make-observable` eslint rule to configure whether fields are annotated with `true` or `false` with the autofixer.

This option defaults to `true` if not present or an invalid value is received to maintain existing behavior.

### Code change checklist

-   [x] Added/updated unit tests -- No unit testing in the eslint package, but I have tested these changes against some other projects that use this rule.

Note: It looks like some formatting rules have changed since the last time this file was touched so there's quite a lot of changes.

Closes #3876

### DIFF
--- a/packages/eslint-plugin-mobx/src/exhaustive-make-observable.js
+++ b/packages/eslint-plugin-mobx/src/exhaustive-make-observable.js
@@ -1,55 +1,61 @@
-'use strict';
+"use strict"
 
-const { findAncestor, isMobxDecorator } = require('./utils.js');
+const { findAncestor, isMobxDecorator } = require("./utils.js")
 
 // TODO support this.foo = 5; in constructor
 // TODO? report on field as well
 function create(context) {
-  const sourceCode = context.getSourceCode();
+    const sourceCode = context.getSourceCode()
 
-  function fieldToKey(field) {
-    // TODO cache on field?
-    const key = sourceCode.getText(field.key);
-    return field.computed ? `[${key}]` : key;
-  }
+    function fieldToKey(field) {
+        // TODO cache on field?
+        const key = sourceCode.getText(field.key)
+        return field.computed ? `[${key}]` : key
+    }
 
-  return {
-    'CallExpression[callee.name="makeObservable"]': makeObservable => {
-      // Only interested about makeObservable(this, ...) in constructor or makeObservable({}, ...)
-      // ClassDeclaration 
-      //   ClassBody
-      //     MethodDefinition[kind="constructor"]
-      //       FunctionExpression
-      //         BlockStatement
-      //           ExpressionStatement
-      //             CallExpression[callee.name="makeObservable"]        
-      const [firstArg, secondArg] = makeObservable.arguments;
-      if (!firstArg) return;
-      let members;
-      if (firstArg.type === 'ThisExpression') {
-        const closestFunction = findAncestor(makeObservable, node => node.type === 'FunctionExpression' || node.type === 'FunctionDeclaration')
-        if (closestFunction?.parent?.kind !== 'constructor') return;
-        members = closestFunction.parent.parent.parent.body.body;
-      } else if (firstArg.type === 'ObjectExpression') {
-        members = firstArg.properties;
-      } else {
-        return;
-      }
+    return {
+        'CallExpression[callee.name="makeObservable"]': makeObservable => {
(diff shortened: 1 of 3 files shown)


urugator:
Would you mind creating a PR that just runs prettier so we would merge it first? If not, it's completely fine, just pinpoint me to the changed lines, I assume it should be just  9, 80, 85.

packages/eslint-plugin-mobx/src/exhaustive-make-observable.js
```diff
@@ -66,46 +73,61 @@ function create(context) {
...
-          } else if (secondArg.type !== 'ObjectExpression') {
-            return fixer.replaceText(secondArg, `{ ${annotationList} }`);
-          } else {
-            const openingBracket = sourceCode.getFirstToken(secondArg)
-            return fixer.insertTextAfter(openingBracket, ` ${annotationList} `);
-          }
-        };
+                const fix = fixer => {
+                    const annotationList =
+                        keys.map(key => `${key}: ${autofixAnnotation}`).join(", ") + ","
```
urugator, line 85: Should have been `${annotation}` ?

packages/eslint-plugin-mobx/src/exhaustive-make-observable.js
```diff
@@ -66,46 +73,61 @@ function create(context) {
...
-      if (!hasAnyDecorator && nonAnnotatedMembers.length) {
-        // Set avoids reporting twice for setter+getter pair or actual duplicates
-        const keys = [...new Set(nonAnnotatedMembers.map(fieldToKey))];
-        const keyList = keys.map(key => `\`${key}\``).join(', ');
+            if (!hasAnyDecorator && nonAnnotatedMembers.length) {
+                // Set avoids reporting twice for setter+getter pair or actual duplicates
+                const keys = [...new Set(nonAnnotatedMembers.map(fieldToKey))]
+                const keyList = keys.map(key => `\`${key}\``).join(", ")
+                const annotation =
+                    typeof autofixAnnotation === "boolean" ? autofixAnnotation : "true"
```
urugator, line 81: Can we move the boolean check to the line 9, where we extract value from options? I would throw if `context.options[0]?.autofixAnnotation` is anything but boolean/undefined.

kade-robertson:
> Would you mind creating a PR that just runs prettier so we would merge it first? If not, it's completely fine, just pinpoint me to the changed lines, I assume it should be just 9, 80, 85.

I can do that -- if it needs to be applied to the whole project though it looks like about 35 files are going to be changed. For now I'll push some commits in a bit that separate out the initial prettier write step and the changes I made to make it more obvious.

packages/eslint-plugin-mobx/src/exhaustive-make-observable.js
kade-robertson: From what I understand of the eslint rule spec, the inputs are validated using the JSON schema added in the `meta.schema` section: https://eslint.org/docs/latest/extend/custom-rules#options-schemas

It should be safe to assume here already that `autofixAnnotation` is always boolean/undefined so the explicit check isn't necessary.

kade-robertson:
See https://github.com/mobxjs/mobx/pull/3881/commits/1e99618c2cb42e5238a38a11f157038e2ec56af5 for the extent of actual changes.

mweststrate:
With additional changes LGTM, thanks for adding this!

mweststrate (approved):

PR #3883, opened 2024-05-30
fix: Allow action field decorators without makeObservable (#3879)

### LINKED ISSUE #3879: @action decorators on fields do not convert the field value into action
**EDIT**: this issue happens when using modern decorators
 

**Intended outcome:**
```ts
import { action, isAction } from 'mobx';

it('action field should be wrapped into mobx.action', () => {
  class Store {
    @action
    x = () => {};
  }

  const store = new Store();

  expect(isAction(store.x)).toBeTruthy();
});
```

or

mobx `action` decorator should not implement a decorator for `field` if it's not supposed to work

**Actual outcome:**
this test above fails, the fn `x` is not an action

Link to stackblitz with replication https://stackblitz.com/edit/vitejs-vite-hdtz1l?file=src%2Fmobx.test.ts

**Versions**
mobx 6.12.3

Not sure why code in the 2022.3 decorator is marked as backwards/legacy and how it's supposed to work?
It looks like it should be returning `return _createAction` or do more stuff in `addInitializer` to actually wrap the declared fn into an action?
 
https://github.com/mobxjs/mobx/blob/bc1a7750fc24d9ad28f338cd0caf75bd504cbe94/packages/mobx/src/types/actionannotation.ts#L77-L82

Obi-Dann:
### Code change checklist

-   [ ] N/A ~Updated `/docs`. For new functionality, at least `API.md` should be updated~

fixes #3879

### DIFF
--- a/.changeset/bright-tips-invite.md
+++ b/.changeset/bright-tips-invite.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix 2022.3 @action decorators on fields no longer require makeObservable
--- a/packages/mobx/__tests__/decorators_20223/stage3-decorators.ts
+++ b/packages/mobx/__tests__/decorators_20223/stage3-decorators.ts
@@ -334,9 +334,7 @@ function normalizeSpyEvents(events: any[]) {
 
 test("action decorator (2022.3)", () => {
     class Store {
-        constructor(private multiplier: number) {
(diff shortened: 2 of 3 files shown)


zhantx:
can we merge this PR?

Matchlighter:
Closing in favor of https://github.com/mobxjs/mobx/pull/3902

PR #3891, opened 2024-06-27
docs(reactions.md): correct autorun function call count

matzar:
Updated the explanation to accurately reflect that the "Energy level" autorun function is called 11 times, not 10. This change clarifies that the initial state is logged once and then each of the 10 changes in the loop is logged, totaling 11 calls.

### DIFF
--- a/docs/reactions.md
+++ b/docs/reactions.md
@@ -107,7 +107,7 @@ Once we run the `for` loop to change the `energyLevel` with the `reduceEnergy`
 action, we see a new log entry every time an `autorun` function observes a
 change in its observable state:
 
-1.  For the _"Energy level"_ function, this is every time the `energyLevel` observable changes, 10 times in total.
+1.  For the _"Energy level"_ function, this is every time the `energyLevel` observable changes, 11 times in total.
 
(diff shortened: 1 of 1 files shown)


docs/reactions.md
```diff
@@ -107,7 +107,7 @@ Once we run the `for` loop to change the `energyLevel` with the `reduceEnergy`
 action, we see a new log entry every time an `autorun` function observes a
 change in its observable state:
 
-1.  For the _"Energy level"_ function, this is every time the `energyLevel` observable changes, 10 times in total.
+1.  For the _"Energy level"_ function, this is every time the `energyLevel` observable changes, 11 times in total.
```
urugator, line 110: The number clearly refers to the number of changes:

> **Once we run the `for` loop to change** the `energyLevel` with the `reduceEnergy`
action, we see a **new log entry every time an `autorun` function observes a
change** in its observable state:
> 1.  For the _"Energy level"_ function, this is **every time the `energyLevel` observable changes**, 10 times in total.

matzar: Ok, I got a little confused there. Closing the PR.

PR #3893, opened 2024-06-27
Fix ES6 Map/Set checks for cross-window scripts

### LINKED ISSUE #3892: Cannot initialize observable Map/Sets from another window
Hi, I'm working on a project that uses MobX to communicate between same-origin iframes and popups.

And I found some issues using observable Maps and Sets across windows (or "realms" according to the ECMAScript spec).

I should admit that this is kinda tricky use-case, but I also found several similar discussions on cross-window cases before: #1644 #2448

**Intended outcome:**

Can initialize observable maps and sets with a MobX instance from different windows.

For example,

```html
<!-- index.html -->
<iframe id="frame" src="./frame"></iframe>
<script>
  const $frame = document.getElementById('frame');
  $frame.contentWindow.mobx = mobx;
</script>
```

```html
<!-- frame.html -->
<script>
  const { autorun, makeObservable, observable } = window.mobx;
  const $size = document.getElementById('size');
  const set = observable.set(new Set());
  // ...
</script>
```

**Actual outcome:**

```
Uncaught Error: [MobX] Cannot initialize set from [object Set]
```

Here is minimal repro: [codesandbox](https://codesandbox.io/p/sandbox/mobx-cross-realm-es6-grs7k5).

**Versions**

Tested in MobX 6.

g6123:
### Related Issue

Fixes #3892 

### Code change checklist

Do I have to add unit tests for this change? I'm not sure is it even possible to make cross-realm ES6 Map/Set instance in jest/jsdom test envionrment.

### DIFF
--- a/packages/mobx/src/types/observablemap.ts
+++ b/packages/mobx/src/types/observablemap.ts
@@ -16,6 +16,7 @@ import {
     hasListeners,
     interceptChange,
     isES6Map,
+    isPlainES6Map,
     isPlainObject,
     isSpyEnabled,
     makeIterable,
@@ -351,7 +352,7 @@ export class ObservableMap<K = any, V = any>
             } else if (Array.isArray(other)) {
                 other.forEach(([key, value]) => this.set(key, value))
             } else if (isES6Map(other)) {
-                if (other.constructor.name !== "Map") {
+                if (!isPlainES6Map(other)) {
                     die(19, other)
                 }
                 other.forEach((value, key) => this.set(key, value))
--- a/packages/mobx/src/types/observablemap.ts
+++ b/packages/mobx/src/types/observablemap.ts
@@ -351,7 +351,7 @@ export class ObservableMap<K = any, V = any>
             } else if (Array.isArray(other)) {
                 other.forEach(([key, value]) => this.set(key, value))
             } else if (isES6Map(other)) {
-                if (other.constructor !== Map) {
+                if (other.constructor.name !== "Map") {
                     die(19, other)
                 }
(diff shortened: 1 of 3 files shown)


packages/mobx/src/types/observablemap.ts
```diff
@@ -351,7 +351,7 @@ export class ObservableMap<K = any, V = any>
             } else if (Array.isArray(other)) {
                 other.forEach(([key, value]) => this.set(key, value))
             } else if (isES6Map(other)) {
-                if (other.constructor !== Map) {
+                if (other.constructor.name !== "Map") {
```
urugator, line 354: Hm, I can still name my subclass `Map`, what about:
`Object.getPrototypeOf(Object.getPrototypeOf(Object.getPrototypeOf(other))) === null`
or
`Object.prototype.toString.call(Object.getPrototypeOf(Object.getPrototypeOf(other))) === '[object Object]'`

g6123: Oh, I see. I added `isPlainES6Map()` to check if a map instance is strictly plain Map object. 103d4b03

urugator:
Can you add a changeset (`yarn changeset`)? Besides that looks good. Thank you.

g6123:
Thank you. I added changeset.

g6123:
When do you think this change will be released? I've applied patch my project for now (via `pnpm patch`), but I'd be happy if I could install the released version.

urugator:
https://github.com/mobxjs/mobx/releases/tag/mobx%406.12.5

PR #3901, opened 2024-07-08
Shrink Atom and Reaction using a bitfield

peterm-canva:
Following on from https://github.com/mobxjs/mobx/pull/3880, use a bitfield to reduce the size of `Atom` and `Reaction`. Large apps can have 10s or 100s of thousands of these. Also further shrink `ComputedValue` by storing `diffValue` in the bitfield, because it's only ever `0` or `1`.

## Memory Savings
| Class | Bytes before | Bytes after |
|--------|--------|--------|
| `ComputedValue` | 100 | 96 |
| `Atom` | 48 | 40 |
| `ObservableValue` (inherits from `Atom`) | 76 | 68 |
| `Reaction` | 72 | 56 | 

## Name Changes
Note this PR changes the names of some members of the above classes to get around the problem mentioned here: https://github.com/mobxjs/mobx/pull/3880#discussion_r1623823258. If these are part of a documented, stable public API then we will need to find another solution.

## Soaking Time
We discussed here https://github.com/mobxjs/mobx/pull/3880#issuecomment-2138116645 that we should let the previous changes to `ComputedValue` soak for a bit to flush out any issues. We have been shipping those changes for a few weeks in production now and haven't seen any problems yet.

### Code change checklist

-   N/A Added/updated unit tests
-   N/A Updated `/docs`. For new functionality, at least `API.md` should be updated


mweststrate:
Looking great again, thanks a lot!

wh1sp3r:
My project stopped working after this update.
No error in output. 

Will check exactly, what's happening

urugator:
@wh1sp3r could be related to https://github.com/mobxjs/mobx/issues/3918

spion:
I'll try to make a repro case this weekend and if it succeeds send a PR over to bump the version.

PR #3902, opened 2024-07-10
fix: Allow action field decorators without makeObservable (#3879)

### LINKED ISSUE #3879: @action decorators on fields do not convert the field value into action

jzhan-canva:
### Code change checklist

Similar to #3883 (credit to @Obi-Dann ) but add extra handling for `@action.bound` on field, for example 
```
class A {
  @action.bound
  doSomething = function() {...} // `this` should be bound to instance of class A
}
```
fix #3879

### DIFF
--- a/docs/enabling-decorators.md
+++ b/docs/enabling-decorators.md
@@ -128,6 +128,9 @@ MobX' 2022.3 Decorators are very similar to the MobX 5 decorators, so usage is m
     The main cases for enumerability seem to have been around serialization and rest destructuring.
     -   Regarding serialization, implicitly serializing all properties probably isn't ideal in an OOP-world anyway, so this doesn't seem like a substantial issue (consider implementing `toJSON` or using `serializr` as possible alternatives)
     -   Addressing rest-destructuring, such is an anti-pattern in MobX - doing so would (likely unwantedly) touch all observables and make the observer overly-reactive).
+-   `@action some_field = () => {}` was and is valid usage. However, inheritance is different between legacy decorators and modern decorators.
+    -   In legacy decorators, if superclass has a field decorated by `@action`, and subclass tries to override the same field, it will throw a `TypeError: Cannot redefine property`.
+    -   In modern decorators, if superclass has a field decorated by `@action`, and subclass tries to override the same field, it's allowed to override the field. However, the field on subclass is not an action unless it's also decorated with `@action` in subclass declaration.
 
 </details>
 
--- a/docs/enabling-decorators.md
+++ b/docs/enabling-decorators.md
@@ -14,8 +14,9 @@ After years of alterations, ES decorators have finally reached Stage 3 in the TC
 With modern decorators, it is no longer needed to call `makeObservable` / `makeAutoObservable`.
 
 2022.3 Decorators are supported in:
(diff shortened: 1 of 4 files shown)


docs/enabling-decorators.md
```diff
@@ -14,8 +14,9 @@ After years of alterations, ES decorators have finally reached Stage 3 in the TC
 With modern decorators, it is no longer needed to call `makeObservable` / `makeAutoObservable`.
 
 2022.3 Decorators are supported in:
-* TypeScript (5.0 and higher, make sure that the `experimentalDecorators` flag is NOT enabled). [Example commit](https://github.com/mweststrate/currencies-demo/commit/acb9ac8c148e8beef88042c847bb395131e85d60).
-* For Babel make sure the plugin [`proposal-decorators`](https://babeljs.io/docs/babel-plugin-proposal-decorators) is enabled with the highest version (currently `2023-05`). [Example commit](https://github.com/mweststrate/currencies-demo/commit/4999d2228208f3e1e10bc00a272046eaefde8585).
+
+-   TypeScript (5.0 and higher, make sure that the `experimentalDecorators` flag is NOT enabled). [Example commit](https://github.com/mweststrate/currencies-demo/commit/acb9ac8c148e8beef88042c847bb395131e85d60).
+-   For Babel make sure the plugin [`proposal-decorators`](https://babeljs.io/docs/babel-plugin-proposal-decorators) is enabled with the highest version (currently `2023-05`). [Example commit](https://github.com/mweststrate/currencies-demo/commit/4999d2228208f3e1e10bc00a272046eaefde8585).
```
jzhan-canva, line 19: this is just auto-formatting

jzhan-canva:
Hi @Matchlighter I [updated doc](https://github.com/mobxjs/mobx/pull/3902/files#diff-ad8f23a88e79871e141c9e793c90969cb8707b9e99ec3f453dd6d93f1316b67dR131-R133) to explain the difference behaviour between legacy and modern
also added a new [unit test](https://github.com/mobxjs/mobx/pull/3902/files#diff-0fa56d5f7de3b3ebf09e12eeac1bb011f920bf5694b98835a86515f0c385f156R1097-R1117) to demonstrate the difference

PR #3906, opened 2024-07-17
Fix #3885 by adding react-dom / react-native as peerDeps

### LINKED ISSUE #3885: mobx-react-lite does not specify `react-dom` nor `react-native` as it's peer dependencies
**Intended outcome:**

`npm install` puts `react-dom` into the right directory so that `mobx-react-lite` can find it.

**Actual outcome:**

`react-dom` might be placed nested into the node_modules of another node_module so that if `mobx-react-lite` is not right next to `react-dom` it can not be found.

tl;dr

1. Clone https://github.com/clentfort/mobx-react-lite-npm-peer-deps/tree/main
2. Run `npm exec redocly`, see `Error: Cannot find module 'react-dom'` error.

I ran into this issue when trying to use a package that depends on mobx-react-lite (namely `@redocly/cli`). They have correctly added `react-dom` as a dependencies in their [package.json](https://github.com/Redocly/redocly-cli/blob/1e5505793b19f7dd239b5b82dfe100e625443993/packages/cli/package.json#L51). However, with a second dependency in play (`cdktf-cli`) running `npm install` will put `react-dom` into `node_modules/@redocly/cli/node_modules/react-dom` while `mobx-react-lite` ends up in `node_modules/mobx-react-lite`. With this `node_modules` layout `mobx-react-lite` can't find `react-dom`. I think `mobx-react-lite` needs to add `react-dom` to it's `peerDependencies`. It is already present in `peerDependenciesMeta` and flagged as optional, but that doesn't tell npm that `react-dom` is an actual peerDependency. It's just meta information about a non-exisiting peerDependency.

**Versions**

mobx-react-lite: 3.4.3
npm: 10.4.0
node: v20.10.0

mweststrate:
### Code change checklist

### DIFF
--- a/.changeset/five-trees-battle.md
+++ b/.changeset/five-trees-battle.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+Added react-dom / react-native as optional dependencies to hint package manager that one of those might be needed. Fixes #3885
--- a/packages/mobx-react-lite/package.json
+++ b/packages/mobx-react-lite/package.json
@@ -42,7 +42,8 @@
     "peerDependencies": {
         "mobx": "^6.9.0",
         "react": "^16.8.0 || ^17 || ^18",
-        "react-dom": ">=16.8.0"
+        "react-dom": ">=16.8.0",
+        "react-native": ">0.69"
     },
     "peerDependenciesMeta": {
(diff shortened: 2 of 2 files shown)


mweststrate:
Didn't help, closing.

PR #3909, opened 2024-07-24
Fix #3908: mobx/missing-observer rule false positive with forwardRef

### LINKED ISSUE #3908: eslint mobx/missing-observer rule false positive with forwardRef
**Intended outcome:**

I am using the `mobx-react` `observer` function with a component that is also wrapped with React `forwardRef`, such as the example below:

```typescript
import { observer } from 'mobx-react';
import { ForwardedRef, forwardRef } from 'react';

const MobxObserverForwardRefExample = observer(
  forwardRef(function MobxObserverForwardRefExample({ name }: { name: string }, ref: ForwardedRef<HTMLHeadingElement>) {
    return (
      <div>
        <h1 ref={ref}>Mobx Observer Forward Ref Example {name}</h1>
      </div>
    );
  }),
);

export default MobxObserverForwardRefExample;
```

I am extending `plugin:mobx/recommended` in my `.eslintrc.json`.

**Actual outcome:**

With the mobx/missing-observer eslint rule enabled, it fails to detect the `observer` wrapper on this component because of the `forwardRef` call in between:

```text
MobxObserverForwardRefExample.tsx
  5:14  warning  Component `MobxObserverForwardRefExample` is missing `observer`  mobx/missing-observer

✖ 1 problem (0 errors, 1 warning)
  0 errors and 1 warning potentially fixable with the `--fix` option.
```

I wasn't able to quickly figure out how to make ESLint work in a CodeSandbox, but this is easy to reproduce with the component above.

**Versions**

- React 18.3.1
- MobX 6.12.0
- mobx-react 9.1.0
- eslint-plugin-mobx 0.0.10

urugator:
Fixes #3908

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,5 +74,6 @@
         "hooks": {
             "pre-commit": "pretty-quick --staged"
         }
-    }
+    },
+    "packageManager": "yarn@1.22.21+sha1.1959a18351b811cdeedbd484a8f86c3cc3bbaf72"
 }
--- a/.changeset/fifty-tools-reply.md
+++ b/.changeset/fifty-tools-reply.md
@@ -1,0 +1,5 @@
+---
+"eslint-plugin-mobx": patch
+---
+
+mobx/missing-observer rule false positive with forwardRef #3908
--- a/packages/eslint-plugin-mobx/preview/missing-observer.js
+++ b/packages/eslint-plugin-mobx/preview/missing-observer.js
@@ -1,32 +1,40 @@
 /* eslint mobx/missing-observer: "error" */
 
-function Named() { }
-const foo = function Named() { }
-const Anonym = function () { };
-const Arrow = () => { };
+function Named() {}
+const named = function Named() {}
+const namedRef = forwardRef(function Named() {})
+const Anonym = function () {}
+const AnonymRef = forwardRef(function () {})
+const Arrow = () => {}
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -74,5 +74,6 @@
         "hooks": {
             "pre-commit": "pretty-quick --staged"
         }
-    }
+    },
+    "packageManager": "yarn@1.22.21+sha1.1959a18351b811cdeedbd484a8f86c3cc3bbaf72"
```
iChenLei, line 78: Why add this configuration? 🤔

urugator: `corepack`  can't live without it unless you set env variable  `COREPACK_ENABLE_AUTO_PIN` to `0`, which is quite inconvinient to do. But now when thinking about it, `yarn` maybe also checks this fields, so it may cause issues to non-corepack users, right?

urugator: Yea, there is a [`SKIP_YARN_COREPACK_CHECK` env var for `yarn`](https://github.com/yarnpkg/yarn/releases/tag/v1.22.21), but this is equally bad as the need to set `COREPACK_ENABLE_AUTO_PIN`. Feel free to revert the change if it's causing problems, unless you know of a better solution of course.

urugator: https://github.com/yarnpkg/yarn/issues/9015

PR #3930, opened 2024-10-02
Doc fix: observable fields is not specified

PredokMiF:
1. It should probably be explained in the documentation that fields marked with `flow(...)` should be excluded from wrapping in `makeObservable`/`makeAutoObservable`
2. Perhaps this place is a good place for example with `false` annotation `makeAutoObservable(this, { fetchProjects: false })`.


mweststrate:
thanks!

PR #3931, opened 2024-10-02
[Decorators] Clarify that modern decorator usage will increase TypeScript project bundle size

chadsowald:
Clarified that using modern decorators will increase the output bundle size for TypeScript projects.

I ran into this on my project after upgrading from legacy decorators.  I wanted to add this to the docs so others don't have to find out the hard way.  Once ESNext is supported by major browsers, though, it will be great!

### DIFF
--- a/docs/enabling-decorators.md
+++ b/docs/enabling-decorators.md
@@ -16,6 +16,7 @@ With modern decorators, it is no longer needed to call `makeObservable` / `makeA
 2022.3 Decorators are supported in:
 
 -   TypeScript (5.0 and higher, make sure that the `experimentalDecorators` flag is NOT enabled). [Example commit](https://github.com/mweststrate/currencies-demo/commit/acb9ac8c148e8beef88042c847bb395131e85d60).
+    -   Note: As modern decorators are not yet supported in major browsers ([implementation status](https://github.com/tc39/proposal-decorators/issues/476)), any TypeScript target other than ESNext will result in a bundle size increase from transpiling compared with legacy decorators. Consider the bundle size impact before moving to modern decorators.
 -   For Babel make sure the plugin [`proposal-decorators`](https://babeljs.io/docs/babel-plugin-proposal-decorators) is enabled with the highest version (currently `2023-05`). [Example commit](https://github.com/mweststrate/currencies-demo/commit/4999d2228208f3e1e10bc00a272046eaefde8585).
 
 ```js


mweststrate (changes requested):
Would you mind quantifying the change, and where it comes from? How it is formulated is pretty clear whether a) it is 100 bytes or 100 KB, and b) whether this is a shared resource like polyfills or something else. Mind charing the difference in transpiled output?

chadsowald:
@mweststrate definitely!  Here's some examples from the TypeScript Playground - rather than my specific project's code - that illustrates the tradeoff.  Would you like me to include these details in the doc update too, or just here?

[With experimentalDecorators on + ES2023](https://www.typescriptlang.org/play/?useDefineForClassFields=true&removeComments=true&experimentalDecorators=true&target=10&jsx=4&allowSyntheticDefaultImports=true&isolatedModules=false&ts=5.6.2#code/MYewdgzgLgBANiA5ogpgEwFwwMJwIYQQAiKoATnlCGTDALwwAUAbnnAK4oA0MA3jAGsAlmDQ8weALYoYAXwCU9AHx8AULSEAzJsNH06DAESahKOGkOLe62jDIoo7MmBib2YYFCHgmIoV7YANTZOKxtbGFBIEDgUADoEREYAAz8AuCEALxFEGAASXglpWRgAd38ACxhWDhkCtKEgkJRZZPkAbnDbe0dnGAam2s6I2WG5VRsAekmYOLnVWQnVFAAPAAdqWGB8QhgAFRRoNQBIAAFE1DRVY7X2ACMM4Bg8YGBDiGp+qBRJACEUHIAUTQ-nQWHcAjAIFKYE6Zwu6GutweQieLzehE+oL+ALAiGBoLQACZwWBIdDYRMbvdHpFwNAyOxPNRGGFjscoh9YgkkIxDN9oJZrotFkA)

[With experimentalDecorators off + ES2023](https://www.typescriptlang.org/play/?useDefineForClassFields=true&removeComments=true&target=10&jsx=4&allowSyntheticDefaultImports=true&isolatedModules=false&ts=5.6.2&experimentalDecorators=false#code/MYewdgzgLgBANiA5ogpgEwFwwMJwIYQQAiKoATnlCGTDALwwAUAbnnAK4oA0MA3jAGsAlmDQ8weALYoYAXwCU9AHx8AULSEAzJsNH06DAESahKOGkOLe62jDIoo7MmBib2YYFCHgmIoV7YANTZOKxtbGFBIEDgUADoEREYAAz8AuCEALxFEGAASXglpWRgAd38ACxhWDhkCtKEgkJRZZPkAbnDbe0dnGAam2s6I2WG5VRsAekmYOLnVWQnVFAAPAAdqWGB8QhgAFRRoNQBIAAFE1DRVY7X2ACMM4Bg8YGBDiGp+qBRJACEUHIAUTQ-nQWHcAjAIFKYE6Zwu6GutweQieLzehE+oL+ALAiGBoLQACZwWBIdDYRMbvdHpFwNAyOxPNRGGFjscoh9YgkkIxDN9oJZrotFkA)

[With experimentalDecorators off + ESNext](https://www.typescriptlang.org/play/?useDefineForClassFields=true&removeComments=true&target=99&jsx=4&allowSyntheticDefaultImports=true&isolatedModules=false&ts=5.6.2&experimentalDecorators=false#code/MYewdgzgLgBANiA5ogpgEwFwwMJwIYQQAiKoATnlCGTDALwwAUAbnnAK4oA0MA3jAGsAlmDQ8weALYoYAXwCU9AHx8AULSEAzJsNH06DAESahKOGkOLe62jDIoo7MmBib2YYFCHgmIoV7YANTZOKxtbGFBIEDgUADoEREYAAz8AuCEALxFEGAASXglpWRgAd38ACxhWDhkCtKEgkJRZZPkAbnDbe0dnGAam2s6I2WG5VRsAekmYOLnVWQnVFAAPAAdqWGB8QhgAFRRoNQBIAAFE1DRVY7X2ACMM4Bg8YGBDiGp+qBRJACEUHIAUTQ-nQWHcAjAIFKYE6Zwu6GutweQieLzehE+oL+ALAiGBoLQACZwWBIdDYRMbvdHpFwNAyOxPNRGGFjscoh9YgkkIxDN9oJZrotFkA)

You can see that web projects following the migration guide to "modern decorators" would be hit by the diff between the first 2 links above.  Then, whenever ESNext is released and enough browsers support modern decorators, then we'll all benefit from the size reduction in the 3rd link.  At least, I think that's my understanding.  

Does that seem about right?

It's hard to quantify the overall size impact as it depends on how much decorator use there is and how large the project is.

In my case, my project heavily uses mobx decorators and the move to modern decorators added ~40kb after compression!  But is has many stores, observables, and more.

PR #3932, opened 2024-10-05
Doc fix: observable fields is not created

PredokMiF:
Observable fields is not created if it placed after makeAutoObservable

### DIFF
--- a/docs/understanding-reactivity.md
+++ b/docs/understanding-reactivity.md
@@ -33,10 +33,10 @@ class Message {
     author
     likes
     constructor(title, author, likes) {
-        makeAutoObservable(this)
         this.title = title
         this.author = author
         this.likes = likes
+        makeAutoObservable(this)
     }
 
     updateTitle(title) {


mweststrate (changes requested):
If you run into this, your compilation setup is incorrect, the declarations at lines 32-34 should have already instantiated the fields in spec compliant class compilation. Please see the setup docs for the right transpiler options.

PredokMiF:
Ok, I will create sandbok for the case to try code

PredokMiF:
Yes, you are right. Some weird default TS settings for `vite` template `react-ts`. But from the learning point of view, my MR is still fair, because after `makeAutoObservable` in strict mode you can only change the referenced values in actions. So it is easier to assign first and then call `makeAutoObservable` to prevent warnings in console.

mweststrate:
That works for simple constructors, but it introduces its own set of problems. E.g. if you introduce branching in the constructor, or call local methods, they might now obtain references to objects before they are turned into observables, etc.

PredokMiF:
Understood. Thank you for your prompt response and clarification. I will close the PR

PredokMiF:
It is realy [bug](https://github.com/vitejs/vite/issues/18297) of `vite`. The workaround is contained in the issue

Issue #3927, opened 2024-09-16
Typescript 5.6.x compatibility with ObservableSet and Observable Map


ymor:
**Intended outcome:**
`ObservableMap` should implement `Map` correctly and `ObservableSet` should implement `Set` correctly
 
**Actual outcome:**

```
Type 'ObservableMap<unknown, unknown>' is not assignable to type 'Map<unknown, unknown>'.
  The types returned by 'entries()' are incompatible between these types.
    Type 'IterableIterator<IMapEntry<unknown, unknown>>' is missing the following properties from type 'MapIterator<[unknown, unknown]>': map, filter, take, drop, and 9 more.
Type 'ObservableSet<unknown>' is not assignable to type 'Set<unknown>'.
  The types returned by 'entries()' are incompatible between these types.
    Type 'IterableIterator<[unknown, unknown]>' is missing the following properties from type 'SetIterator<[unknown, unknown]>': map, filter, take, drop, and 9 more.

```

**How to reproduce the issue:**
https://www.typescriptlang.org/play/?target=99&ts=5.6.2#code/JYWwDg9gTgLgBAbzgeQEYGcCmUBuBDVAG0wFk8wAaFDbfIzAZU3gF84AzKCEOAchAioAHrwDcAKHEBjCADt08AcLJgAXHBUAeAK6yA1rIgB3WVV0HjsgHxwAvHFmYj1LLgLEVACgCUE6XIU4JSEmGHVQnX1DExt7R2c0VzpiUJ8JIA

Seems this might be relevant https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#iterator-helper-methods

No errors in Typescript 5.5.x

**Versions**
mobx@6.13.2

Weffe:
Just ran into this issue too. I think we need to update ObservableMap to match the new structure for `.entries()` with the new iterator helpers.

tonyraoul:
Started working on a potential fix [here](https://github.com/mobxjs/mobx/pull/3935)
keen for your feedback.

Issue #3903, opened 2024-07-12
Updating to 6.13.0 causing error: `Cannot find name 'ReadonlySetLike'. Did you mean 'ReadonlySet'?`


schen-fn:
**Intended outcome:**
Upgrading to the 6.13.0 minor update should not produce any errors, specifically the `Cannot find name 'ReadonlySetLike'. Did you mean 'ReadonlySet'?` error.

**Actual outcome:**
Upon upgrading to version 6.13.0, the following error occurs: `Cannot find name 'ReadonlySetLike'. Did you mean 'ReadonlySet'?`.

**How to reproduce the issue:**
1. Upgrade to version 6.13.0
2. Observe the Error: `Cannot find name 'ReadonlySetLike'. Did you mean 'ReadonlySet'?`

**Configuration Details:**

Using es2020 in tsconfig.json.
This issue mirrors the problem described in this [release job on GitHub](https://github.com/mobxjs/mobx/actions/runs/9765918034).

**Additional Information**
The error is resolved by switching the target in tsconfig.json from es2020 to esnext. However, this change is not backward-compatible, which may be problematic for a minor update.

mweststrate:
The release build was fixed in https://github.com/mobxjs/mobx/commit/32c934bea82feaddff01aad67231ee4a1d64ede9. 

In this case you'll indeed have to update the TSLib to ESNext. Like TypeScript itself, we don't follow semantic versioning closely for static types, as otherwise we'll have to fork the package for every different tsconfig / target people are using. (E.g. if we made this a major change, it means that anyone not on `esnext` would stop receiving any update to MobX! Even though there is no actual functional consequence)

It is not ideal, and generally I avoid this problem (many libs suffer from it) by slamming a `skipLibCheck: true` in my projects' tsconfig.

triemstr:
Seems there are 7 tsc errors for me, all for `ReadonlySetLike` from https://github.com/mobxjs/mobx/pull/3898/commits/7131addbf363fc94a37867194de35fbc14651efa

If @inoyakaigor has a thought otherwise that can make this work for those using defaults on vite 5+ (target: modules is the default), that would be appreciated also.

Until then, if anyone else enters this issue, yes you can use esnext (although I have not been able to get that to work despite setting it in 3 different vite config targets per stackoverflow), or give up and downgrade like me to mobx - 6.12.5.

inoyakaigor:
@trimester could you provide minimum reproducible sample?

[...]

PR #3935, opened 2024-10-08
Fix Typescript 5.6.x support by introducing esnext iterator helpers (#3927)

tonyraoul:
### Code change checklist

### Context

Fixes https://github.com/mobxjs/mobx/issues/3927

### Purpose of the PR
TypeScript 5.6.x introduces new [Iterator Helper Methods](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#iterator-helper-methods) that enhance the `IterableObject` interface with methods like `map`, `flatMap`, `take`, and `drop`. However, when the `esnext` target is enabled, MobX becomes incompatible as its current implementation lacks these methods, resulting in type errors.

This PR addresses the compatibility issues by:

- **Updating the `IterableObject` Interface**: Aligning MobX’s implementation with TypeScript’s updated type definitions to prevent type errors.
- **Integrating Support for Iterator Helper Methods**: Adding the new methods within MobX and enabling conditional polyfill support for environments that do not natively implement them.

### DIFF
--- a/packages/mobx/src/utils/iterable.ts
+++ b/packages/mobx/src/utils/iterable.ts
@@ -1,8 +1,5 @@
-export function makeIterable<T>(iterator: Iterator<T>): IterableIterator<T> {
-    iterator[Symbol.iterator] = getSelf
-    return iterator as any
-}
-
-function getSelf() {
-    return this
+export function makeIterable<T, TReturn = unknown>(
+    iterator: Iterator<T>
+): IteratorObject<T, TReturn> {
+    return Object.assign(Object.create(Iterator.prototype), iterator)
 }
--- a/.changeset/neat-ants-relate.md
+++ b/.changeset/neat-ants-relate.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Update typescript version to 5.6.2 and added support for esnext iterator helpers
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     ],
     "resolutions": {
         "jest": "^29.5.0",
-        "typescript": "^5.5.2",
+        "typescript": "^5.6.2",
         "recast": "^0.23.1"
     },
     "repository": {
@@ -68,7 +68,7 @@
         "tape": "^5.0.1",
         "ts-jest": "^29.0.5",
         "tsdx": "^0.14.1",
-        "typescript": "^5.5.2"
+        "typescript": "^5.6.2"
     },
     "husky": {
         "hooks": {
--- a/packages/mobx/__tests__/v5/base/map.js
+++ b/packages/mobx/__tests__/v5/base/map.js
@@ -915,6 +915,34 @@ test("maps.values, keys and maps.entries are iterables", () => {
     expect(Array.from(x.keys())).toEqual(["x", "y"])
 })
 
+// Test support for [iterator-helpers](https://github.com/tc39/proposal-iterator-helpers)
+test("esnext iterator helpers support", () => {
+    const map = mobx.observable(
+        new Map([
+            ["x", [1, 2]],
+            ["y", [3, 4]]
+        ])
+    )
+
+    expect(Array.from(map.keys().map(value => value))).toEqual(["x", "y"])
+    expect(Array.from(map.values().map(value => value))).toEqual([
+        [1, 2],
+        [3, 4]
+    ])
+    expect(Array.from(map.entries().map(([, value]) => value))).toEqual([
+        [1, 2],
+        [3, 4]
+    ])
+
+    expect(Array.from(map.entries().take(1))).toEqual([["x", [1, 2]]])
+    expect(Array.from(map.entries().drop(1))).toEqual([["y", [3, 4]]])
+    expect(Array.from(map.entries().filter(([key]) => key === "y"))).toEqual([["y", [3, 4]]])
+    expect(Array.from(map.entries().find(([key]) => key === "y"))).toEqual(["y", [3, 4]])
+    expect(map.entries().toArray()).toEqual(Array.from(map))
+
+    expect(map.entries().toString()).toEqual("[object MapIterator]")
+})
+
 test("toStringTag", () => {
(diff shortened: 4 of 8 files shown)


packages/mobx/src/utils/iterable.ts
```diff
@@ -1,8 +1,5 @@
...
-    iterator[Symbol.iterator] = getSelf
-    return iterator as any
-}
-
-function getSelf() {
-    return this
+export function makeIterable<T, TReturn = unknown>(
+    iterator: Iterator<T>
+): IteratorObject<T, TReturn> {
+    return Object.assign(Object.create(Iterator.prototype), iterator)
```
tonyraoul, line 4: Intentionally avoiding the usage of Iterator.from due to limited availability, using this instead to conditionally add helper functions IIF they are available within runtime either natively or via polyfill.

mweststrate (approved):
This looks great, thanks a lot!

zeldigas:
@mweststrate it seems that it is still not fixed. While typecheck is good, I got the following error with typescript 5.6.3 and mobx 6.13.4:
```
@app/app:test:  FAIL  src/components/organisms/component.store.test.ts > ComponentStore > some test
@app/app:test: ERROR: command finished with error: command (/workspace/master/packages/app) /usr/local/bin/pnpm run test exited (1)
@app/app:test: ReferenceError: Iterator is not defined
@app/app:test:  ❯ makeIterable ../../node_modules/.pnpm/mobx@6.13.4/node_modules/mobx/src/utils/iterable.ts:4:40
@app/app:test:  ❯ makeIterableForSet ../../node_modules/.pnpm/mobx@6.13.4/node_modules/mobx/src/types/observableset.ts:349:12
@app/app:test:  ❯ ObservableSet.values ../../node_modules/.pnpm/mobx@6.13.4/node_modules/mobx/src/types/observableset.ts:237:16
@app/app:test:  ❯ ObservableSet._proto.<computed> ../../node_modules/.pnpm/mobx@6.13.4/node_modules/mobx/src/types/observableset.ts:334:21
@app/app:test:  ❯ _createForOfIteratorHelperLoose ../../node_modules/.pnpm/mobx@6.13.4/node_modules/mobx/dist/mobx.cjs.development.js:308:24
@app/app:test:  ❯ ObservableSet.forEach ../../node_modules/.pnpm/mobx@6.13.4/node_modules/mobx/src/types/observableset.ts:110:9
@app/app:test:  ❯ ComponentStore.call src/components/organisms/component.store.ts:117:31

@app/app:test:     115|   get property(): Abc[] {
@app/app:test:     116|     const array: Abc[] = [];
@app/app:test:     117|     this.uiData.xyz.forEach((selected, item) => {
@app/app:test:        |                               ^
@app/app:test:     118|       if (selected) {
@app/app:test:     119|         array.push(item);
```
`xyz` property in code is initialized as `xyz: new Set<Abc>()` and is located in class that is made autoobservable

tonyraoul:
@zeldigas

As per log
`@app/app:test: ReferenceError: Iterator is not defined`

seems like the typescript version used is not 5.6.x, maybe it would be safer to add a fallback for older typescript versions support or add a warning at least.

tonyraoul:
`@app/app:test: ReferenceError: Iterator is not defined` @zeldigas 
what's your estarget ? 

Most likely it's this line 😨 
https://github.com/mobxjs/mobx/pull/3935/files#diff-22d6d812427f3e234cc1b5448baf52696bc9a71646c71128e752c61c8a789c31R4

Iterator is not available before es2015

while this issue can be fixed by updating node version I am a bit unsure about compatibility requirements for mobx.

zeldigas:
The project I'm working on have specific setup, with mixed use of es5 and es2022 in package.json files. Quick update all the places to `es2022` did not help. I'll check in more details later today. Overal of course it's good to provide some isolated example of failure, this is clear..

tonyraoul:
apologies for confusion @zeldigas believe if this is causing anyone issues a potential fix might be something like this, it's not working on safari apparently, thanks for reporting! working on a proper fix but this is what I have in mind for now.
```
    const maybeIteratorPrototype = eval("Iterator.prototype") || {}
    return Object.assign(Object.create(maybeIteratorPrototype), iterator)
 ```

my apologies for any inconveniences.

tonyraoul:
I believe I misunderstood MDN reference reporting that Iterator.prototype is widely available but that's not the case 🤔

tonyraoul:
I hope this fixes it, https://github.com/mobxjs/mobx/pull/3943

jansav:
FYI: This PR introduced breaking change by utilizing `MapIterator` which was introduced in `typescript` 5.6. It's was very annoying to debug why all types break when `mobx`-version was updated based on semver rules from `6.12.4` -> `6.13.5`.

This also means that `6.13.x` version requires at least `typescript` 5.6.

thisisanto:
> FYI: This PR introduced breaking change by utilizing `MapIterator` which was introduced in `typescript` 5.6. It's was very annoying to debug why all types break when `mobx`-version was updated based on semver rules from `6.12.4` -> `6.13.5`.
> 
> This means that `6.13.x` version requires at least `typescript` 5.6.

Plus 1 this is a breaking change that's disguised as a patch. Should probably be reverted, and re-introduced along with the appropriate `peerDependency` bump to `typescript` @tonyraoul

tonyraoul:
apologies for the inconvenience  @jansav @thisisanto
I will start working on a fix, if you don't mind I will be adding another patch rather than unpublish this one since I am not a maintainer, hopefully it gets picked up soon.

Thanks for your valuable feedback.

tonyraoul:
Working on a fix [here](https://github.com/mobxjs/mobx/pull/3960) let me know if you have any thoughts please.

mweststrate:
@thisisanto  @jansav to confirm: we (try to) always release against the latest version of TypeScript (which itself doesn't follow any versioning scheme), and strongly recommend to use `skipLibCheck` `true`. For some more details behind the rationale, see https://github.com/mobxjs/mobx/issues/3903#issuecomment-2451426412.

Issue #3856, opened 2024-04-08
[eslint-plugin-mobx] ESLint v9 Support


dartess:
ESLint has released v9: https://eslint.org/blog/2024/04/eslint-v9.0.0-released/

Could you please add compatibility with version 9?

As far as I can see, there are quite a lot of breaking changes there. And flat format is now enabled by default. I'm not sure, but this means that you either need to migrate completely to a flat config, or maintain two versions of the config.

SimenB:
As I was posting this message, the above PR showed up 😀 

---

ESLint@8 has been deprecated on npm now, would love to see official support here 🙂

PR #3942, opened 2024-10-16
add eslint@9 support and flat config

dartess:
### Code change checklist

Closed https://github.com/mobxjs/mobx/issues/3856 

- added compatibility with eslint@9;
- added support for flat config;
- added recommended config in flat format;
- added documentation for using with flat config;
- added tests to check rules in legacy (using eslint@7) and in modern (eslint@9) environment.
- everything is tested on my own project with flat config;


iChenLei:
I'm sorry, I accidentally clicked the button and removed the Draft mode

dartess:
@iChenLei moved to draft to add changeset; already done

dartess:
- added compatibility with eslint@9;
- added support for flat config;
- added recommended config in flat format;
- added documentation for using with flat config;
- added tests to check rules in legacy (using eslint@7) and in modern (eslint@9) environment.
- everything is tested on my own project with flat config;

I think that's all :raised_hands:

packages/eslint-plugin-mobx/package.json
```diff
@@ -45,7 +47,9 @@
         "mobx"
     ],
     "scripts": {
-        "test": "jest",
+        "test:7": "jest --config jest.config-eslint-7.js",
+        "test:9": "jest --config jest.config-eslint-9.js",
```
urugator, line 51: Do we need seperate configs? Can't we do something like:
```
// package.json
"test:7": "jest --config jest.config.js -- 7",
"test:9": "jest --config jest.config.js -- 9"

// jest.config.js
globals: {
        ESLINT_V: process.argv[2]
}
```

packages/eslint-plugin-mobx/rollup.config.js
```diff
@@ -13,6 +14,7 @@ export default [
         babelHelpers: "bundled",
         exclude: "**/node_modules/**",
       }),
+      json(),
```
urugator, line 17: purpose?

packages/eslint-plugin-mobx/__tests__/utils/getRuleTester.js
```diff
@@ -0,0 +1,29 @@
+const version = global.ESLINT_V;
```
urugator, line 1: kebab case for filenames: get-rule-tester.js

urugator:
There is a preview directory, that provides "live" (you need to reload window in between changes) IDE preview . This is super useful when developing a rule. Since you removed `"eslint": "^7.0.0"` from dev deps, I assume the IDE will use eslint installed at the root of the mobx repo, which is `"eslint": "^6.8.0",` . Can we make sure it uses 7 or 9?  I think we don't need preview for both at the moment.

packages/eslint-plugin-mobx/package.json
dartess: Something like that was the “Plan A”: https://github.com/mobxjs/mobx/pull/3942/commits/5c9f4bcd3d3b067fa3cb598b9b5e02be068f6625#diff-44bf0440e178d6213ede81d55d2733c534c53afb6ccdcf2451520ac2cffdf191R51

This was convenient when running "npm test" directly from the repository and required only one config (like your option).

However, CI run "npm test" from the root of the repository, and it simply runs tests for all configs of all packages, which resulted in the error: https://github.com/mobxjs/mobx/pull/3942/checks?check_run_id=31599042993 . In both my case and your the versions will be lost. Therefore, I would suggest leaving the current variant, although it requires a little duplication of code. If this is critical, we can allocate two lines to some shared config. But I think this is too much.

packages/eslint-plugin-mobx/rollup.config.js
dartess: In general for this: https://github.com/mobxjs/mobx/pull/3942/commits/5c9f4bcd3d3b067fa3cb598b9b5e02be068f6625#diff-ea6b7dd8ae5b79523bfec5004a125b001785f118b2ff73dded233e51648b1e38R3 . Eslint recommends adding name and version into `meta` to improve caching: https://eslint.org/docs/latest/extend/plugins#meta-data-in-plugins 

A common approach in other plugins is simply `require`/`import` `package.json` already on the consumer machine (in runtime) - this way the meta version will always match the package version.

However, I did not take into account that in this case the compilation process occurs, and the package is copied to the `dist` as a whole (and I don't know how you synchronize the process of increasing the version and build, so it looks unreliable). It seems that it is better to use a slightly more complex approach from the eslint docs. So, I rewrote it.

There are also options with manual control of `meta` values (that shooting yourself in the foot sooner or later) or somehow integrate the process of filling this data into your build processes, but I am not familiar with them and am not sure that I can do it well.

dartess:
> There is a preview directory, that provides "live" (you need to reload window in between changes) IDE preview . This is super useful when developing a rule

Wow! I didn't realize this was used during rules developing. This looks nice!

> if you know of better way, let me know

In fact, I see a flaw in this way. It is essentially a manual way of testing. It can be really convenient during development, but at the end of development, all preview examples (in a good way) should be copied to test cases. However, this can be forgotten, and all this is code duplication (what is bad).

Usually I've seen a simpler way - development through test cases in observation mode. I can add something like "test:watch" to the scripts. It's not as convenient, but it just works and doesn't require a "preview" dir at all.

> I assume the IDE will use eslint installed at the root of the mobx repo, which is "eslint": "^6.8.0"

Yes, that's true, at least for JetBrains IDE. I didn't take that into account either.

Option 1: remove "preview" folder for development and switch to tests and `watch` mode (to avoid the need to synchronize `preview` and `__tests__`) (I prefer this)

Option 2: configure `preview` with eslint@**7** (just return it to how it was)

Option 3: configure `preview` with eslint@**9** (all versions below 9 are officially deprecated now, so this makes sense, but there may be problems with switching between different configs for editors and IDE)

Your opinion?

urugator:
> It is essentially a manual way of testing

Let me aswer with some rhetorical questions:
Have you ever developed an application just by writing tests?
At the time of writing, eslint supported multiple suggestions/fixes, but vscode couldn't handle that, how would you know?
In the OP, you're mentioning  *"everything is tested on my own project with flat config;"*. So automated tests did not suffice?

> Your opinion?

I am fine with both 7/9 as long as it works in vscode :D Your call.

packages/eslint-plugin-mobx/package.json
urugator: I see. OK

packages/eslint-plugin-mobx/rollup.config.js
urugator: Thank you for elaborating. If I follow correctly, the current approach works and doesn't require extra dep.

dartess:
yes, these are good questions.

I rollback eslint@7 as dev deps, now it should work as before.

urugator:
Thank you for looking into it. Much appreciated.

PR #3943, opened 2024-10-16
Fix(mobx): Safer method to consume Iterator.prototype

tonyraoul:
### Code change checklist

Really sorry, I might have introduced a bug [here](https://github.com/mobxjs/mobx/pull/3935)
thanks @zeldigas for reporting! 

### Context

For some projects setups, `Observablemap.prototype.forEach`, `ObservableSet.prototype.forEach` might throw an error on older browsers and Safari, I am genuinely sorry about this one.

I wrongfully assumed that `Iterator.prototype` is widely available but that's not the case.

Screenshot 2024-10-16 at 11 01 48 pm

### DIFF
--- a/.changeset/five-news-hammer.md
+++ b/.changeset/five-news-hammer.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix browser compatability issue introduced in 6.13.4 release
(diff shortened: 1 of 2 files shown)


tonyraoul:
@mweststrate really sorry one seems a bit urgent 😞..

mweststrate:
No worries, these things happen in ever changing browser land! Thanks for the quick fix. Up to the next breakage🚀

PR #3960, opened 2024-11-15
Fix(mobx): Typescript pre-5.6 backwards compatibility

tonyraoul:
### Description
I've received some negative feedback w.r.t this patch https://github.com/mobxjs/mobx/pull/3935 `mobx@6.13.4`.
as it breaks typescript backwards compatibility with versions before 5.6.

Apologies for any inconveniences this has caused, as far as I understand typescript isn't pinned to a specific version by design hence the updates being reported as a patch within previous changes, I am not entirely sure if this changeset makes much sense to include within the mobx package or it would be preferable to start specifying peer dependencies for typing.

Alternatively you can use `mobx@6.13.4+` with older typescripts versions by adding the attached `.d.ts` file to your project if you really need to type check dependencies.

Going to quote mwestrate [here](https://github.com/mobxjs/mobx/issues/3903#issuecomment-2234093654)
>  It is not ideal, and generally I avoid this problem (many libs suffer from it) by slamming a skipLibCheck: true in my projects' tsconfig.

probably not worth it to add this patch but keen for feedback non the less.

### DIFF
--- a/packages/mobx/.changeset/odd-years-shop.md
+++ b/packages/mobx/.changeset/odd-years-shop.md
@@ -1,5 +1,0 @@
----
-"mobx": patch
----
-
-Fix typescript backwards compatability for versions pre 5.6
--- a/packages/mobx/.changeset/odd-years-shop.md
+++ b/packages/mobx/.changeset/odd-years-shop.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix typescript backwards compatability for versions pre 5.6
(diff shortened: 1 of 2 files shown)


tonyraoul:
After some testing, I believe there is no value at all in shipping this vs using `skipLibCheck: true`
Going to close it for now, sorry for the noise.

mweststrate:
Yes, in general we just stick to TS latest since TS itself doesn't have any form of semantic versioning, so no worries.

PR #3985, opened 2024-12-09
React 19 support

imjordanxd:
### Code change checklist

-   ~Added/updated unit tests~ N/A
-   ~Updated `/docs`. For new functionality, at least `API.md` should be updated~ N/A


packages/mobx-react-lite/__tests__/observer.test.tsx
```diff
@@ -997,38 +997,6 @@ it("dependencies should not become temporarily unobserved", async () => {
     expect(doubleDisposed).toBeCalledTimes(1)
 })
 
-it("Legacy context support", () => {
```
imjordanxd, line 1000: legacy context is gone in R19

mweststrate: I think it is fine to disable that that test (`it.skip("...`). In the next major we're going to drop it anyway

imjordanxd: skipping instead

packages/mobx-react-lite/__tests__/useLocalStore.deprecated.test.tsx
```diff
@@ -377,7 +377,7 @@ describe("is used to keep observable within component body", () => {
             })
             expect(container.querySelector("span")!.innerHTML).toBe("22")
             expect(counterRender).toBe(2)
-            expect(observerRender).toBe(3)
```
imjordanxd, line 380: React now renders the observer an extra time for some reason

packages/mobx-react-lite/src/ObserverComponent.ts
```diff
@@ -6,6 +6,11 @@ interface IObserverProps {
 }
 
 function ObserverComponent({ children, render }: IObserverProps) {
+    if (children && render) {
+        console.error(
+            "MobX Observer: Do not use children and render in the same time in `Observer`"
+        )
+    }
```
imjordanxd, line 13: `propTypes` removed in v19 so I've duped this inside the render body

packages/mobx-react/__tests__/hooks.test.tsx
```diff
@@ -49,7 +49,7 @@ test("computed properties react to props when using hooks", async () => {
     act(() => {
         jest.runAllTimers()
     })
-    expect(seen).toEqual(["parent", 0, "parent", 2])
+    expect(seen).toEqual(["parent", 0, "parent", 2, 2])
```
imjordanxd, line 52: ditto: https://github.com/mobxjs/mobx/pull/3985/files#r1879267971

packages/mobx-react-lite/__tests__/useLocalObservable.test.tsx
```diff
@@ -440,75 +439,108 @@ describe("enforcing actions", () => {
...
+                    inc() {
+                        this.count += 1
+                    }
+                }))
+                useEffect(() => {
+                    store.multiplier = multiplier
+                }, [multiplier])
+                useEffect(() => setMultiplier(3), [])
+            },
+            {
```
mweststrate, line 461: NIT: This pattern reoccures quite a few times in the test, might be nice to make a utility for it? (separate PR is fine if interested)

Venryx:
I must be missing something, but I am getting TypeScript errors when trying to use `mobx-react@9.2.0` in a repo with `react@19.0.0` and `@types/react@19.0.10`.

The errors relate to exports such as `React.Requireable` not being available in `@types/react@19.0.10`:
```
node_modules/mobx-react/dist/propTypes.d.ts:2:38 - error TS2694: Namespace 'React' has no exported member 'Requireable'.

2     observableArray: import("react").Requireable<any>;
```

Does anyone else see the same issue? (will keep looking into it in the meantime, but it seems to me that mobx-react is just relying on react type exports that no longer exist)

Venryx:
Okay yeah, it looks like in `@types/react@18`, there were `Requireable` and `Validator` exports: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/da70d4aee37cd531df1fc030c1e068fe51d3b5e9/types/react/v18/index.d.ts#L4149-L4157

```
    /**
     * @deprecated Use `Validator` from the ´prop-types` instead.
     */
    type Validator<T> = PropTypes.Validator<T>;

    /**
     * @deprecated Use `Requireable` from the ´prop-types` instead.
     */
    type Requireable<T> = PropTypes.Requireable<T>;
```

However, as seen, they were marked as deprecated.

And then in `@types/react@19`, they were removed (ctrl+f shows no results): https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts

So if `Requireable` and `Validator` are still needed by mobx-react, they need to import from the `prop-types` package now, not `react`.

Venryx:
@mweststrate Okay, I've created a PR for it here: https://github.com/mobxjs/mobx/pull/4534

PR #3993, opened 2025-01-17
Perf(observableset): Remove extra clone operation on .values .keys .Iterator access

tonyraoul:
### Code change checklist

### Context
within the observableset, on retriving `entries/keys/values` or during iteration, there is an extra `Array.from` operation that is causing an unwanted copy of values within memory, the significance of extra operation is amplified when considering iterator helper methods where it would read the entire range of elements within the set instead of lazy evaluation, i.e: `Set.prototype.entries().take(1)`, or any other case when the iterator is partially retrieved `Set.prototype.entries().next()`.

#### Time
Use a macbook air M1 - 8GB model - plugged in, 1000 iterations
| test |  before |  after | note |
|---|---|---|---|
|  Set: conversion to array - 1000 elements | 3.33s min - 3.41s max | 2.90s min - 2.95s max  | non significant |
|  Set: conversion to array - 10000 elements | 12.74s min - 12.87s max | 9.10s min - 9.16s max | non significant |
|  Set: iterator helpers - 1000 elements, take(1) | 0.60s min - 0.77s max | 0.31s min - 0.32s max | exponential 🎊  |
|  Set: iterator helpers - 10000 elements, take(1) | 4.19s min - 4.22s max | 0.31s min - 0.32s max | exponential 🎊   |

#### Memory
Mesuring the effect on memory when converting a large observableset to an array.

| size |  before |  after |
|---|---|---|
|  1M item |  66 Mb | 46 Mb | 
|  10M item | 617 Mb | 423 Mb | 

```
   function benchmarkMemoryUsage() {
            const beforeMemory = memoryUsage().heapUsed
            
            const obs = observable.set()
            for (let i = 0; i < 10000_000; i++) {
                obs.add(i)
            }

            const x = Array.from(obs)
            const afterMemory1 = memoryUsage().heapUsed

            console.log(`Memory diff: ${(afterMemory1 - beforeMemory) / 1000_000} Mbytes %`)
        }
```


packages/mobx/src/types/observableset.ts
```diff
@@ -211,15 +211,14 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
     }
 
     entries() {
-        let nextIndex = 0
-        const keys = Array.from(this.keys())
-        const values = Array.from(this.values())
+        const keys = this.keys()
+        const values = this.values()
```
tonyraoul, line 215: Seems like keys can be omitted.

iChenLei:
Cool 👍

mweststrate:
Nice work, thanks!

dkazakov8:
Nice work! 

Is there a specific reason to use `this.values()`

```
entries() {
  const values = this.values()
	
  return makeIterableForSet<[T, T]>({
    next() {
      const { value, done } = values.next()
	  
      return !done ? { value: [value, value], done } : { value: undefined, done }
    }
  })
}
```

instead of `this.entries()`?

```
entries() {
  const values = this.entries()
	
  return makeIterableForSet<[T, T]>({
    next() { return values.next() }
  })
}
```
	
Seems like the second approach would be more optimized (less constants, less objects, destructuring and no manual creation of arrays)

PR #4214, opened 2025-01-30
ECMA private fields with auto-accessor are in fact supported

creasty:
### Code change checklist

-   [-] Added/updated unit tests
-   [-] Verified that there is no significant performance drop (`yarn mobx test:performance`)

## Why

Needs clarification.


mweststrate:
Great catch, thanks!

PR #4528, opened 2025-02-19
Fix Set interceptors

k-g-a:
### Code change checklist

I'm sorry for the missing related issue. 
Just figured out that the `newValue` returned from the `ObservableSet`'s interceptor is ignored. The following code still uses the base value passed to `add()` method. At a quick glance, the `set()` method of the `ObservableMap` just re-defines the value variable with the `change.newValue`. So I've reproduced the logic and added coupe of tests.

### DIFF
--- a/.changeset/set-interceptors.md
+++ b/.changeset/set-interceptors.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix observable.set not respecting the new value from interceptors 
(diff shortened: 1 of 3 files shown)


k-g-a:
@urugator / @mweststrate , could you, please, have a look at this PR?

mweststrate (approved):
Looks correct, thanks @k-g-a!

PR #4534, opened 2025-03-09
Resolve TypeScript errors in React 19 (by removing the deprecated propTypes.ts)

Venryx:
Removed the deprecated `propTypes.ts`, and some remaining references to its exports, in order to make mobx-react compatible with React 19 (without TypeScript compile errors), since React 19 removes its prop-types exports of `React.Requireable` and `React.Validator`. (@mweststrate gave a go-ahead for this [here](https://github.com/mobxjs/mobx/pull/3985#issuecomment-2709007700))

Changelog:
* Removed propTypes.ts, as React 19 no longer provides the types `React.Requireable` and `React.Validator`.
* Updated docs to not mention the mobx-react specific prop-types anymore.
* Removed the `propTypes.test.ts` file. (not applicable anymore, since the mobx-react specific prop-types no longer exist)
* Removed the "component with observable propTypes" test, since it's no longer applicable. (and it was referencing typescript types that are no longer importable, so using `test.skip` is not sufficient)

### Code change checklist

-   [X] Added/updated unit tests


Venryx:
I added a basic changelog entry for the PR: (to satisfy changeset-bot)
> Removed the deprecated PropTypes export, as React 19 no longer provides the types `React.Requireable` and `React.Validator`, and this was causing TypeScript errors with `@types/react@19.0.10`.

I marked it as a major change (assuming even deprecated APIs are considered part of the semver-tracked api-surface), but I can change it to minor if needed.

mweststrate (changes requested):
Thanks for making this PR! Doing this as major sounds both as slightly overkill and conceptually correct. Doing it as minor will always trip some project up though, so let's do a major :)

Would you mind updating `peerDependencies` in package.json, and the compatibility table in the README accordingly? Otherwise looking great, thanks!

Venryx:
I've added the second change requested (compatibility table update in readme).

I'm also trying to add the first change requested (adding of react 19 as a peer dependency). But git appears to have a hook that warns me to run "yarn install" prior to pushing. When I do so, it seems to result in a massive set of changes to the lock file, which would not be feasible for maintainers to visually review. So I imagine this step would have to be performed by an existing trusted maintainer (to ensure the lock file is authentic).

kubk:
@Venryx Thanks for the original work!

The mobx 7 migration removed propTypes.ts, the public PropTypes export, related tests and documentation, and added React 19 to the peer dep range https://github.com/mobxjs/mobx/pull/4671

PR #4588, opened 2025-10-08
docs: Migration guide: Update makeObservable section with decorators support

magoni:
Update outdated claim in the 4-5 to 6 Migration guide `makeObservable` section: it's no longer true that MobX 6 does not support decorators.

It's mentioned earlier in the doc, but this is for those like myself who skim docs.


mweststrate:
Thanks!

PR #4591, opened 2025-10-17
Fix XH sponsor link in README.md

amcclain:
Fixing up our sponsor link to point to the correct domain - thanks! 🙏


kubk:
Thank you

PR #4558, opened 2025-06-04
feat: Explicit Resource Management support in reactions

vkrol:
Discussion: https://github.com/mobxjs/mobx/discussions/4559

### Code change checklist

-   [x] Added/updated unit tests
-   [x] Updated `/docs`. For new functionality, at least `API.md` should be updated
-   [x] Verified that there is no significant performance drop (`yarn mobx test:performance`)

PR #4592, opened 2025-10-18
feat: Extend `IReactionDisposer` with `Disposable`

VimHax:
To enhance #4558 this PR extends `IReactionDisposer` with the [`Disposable` interface](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management) so that TypeScript does not complain when using the disposable.

### Code change checklist

### DIFF
--- a/.changeset/gold-windows-brake.md
+++ b/.changeset/gold-windows-brake.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
+Extend IReactionDisposer with Disposable


kubk:
Thanks! Hard to think of a concrete use case, but as a follow-up to https://github.com/mobxjs/mobx/pull/4558 it should make sense

VimHax:
> Thanks! Hard to think of a concrete use case, but as a follow-up to #4558 it should make sense

Thank you so much for merging this! If I remember correctly *any* usage of reactions as a `Disposable`s gets marked as an error, so without this change the original PR's feature effectively cannot be used.

```ts
// this code is from the original PR
this.disposableStack.use(autorun(() => {
  doSomethingWith(this.price * this.amount)
})) // this would be an error because `IReactionDisposer` does not implement `Disposable`
```

PR #4609, opened 2026-01-21
fix(mobx-react-lite): Ensure onBecomeUnobserved is properly called immediately on unmount of components that did suspend

Nokel81:
Reworked https://github.com/mobxjs/mobx/pull/3777


kubk:
Thank you for the contribution. This will be superseeded with a different approach here https://github.com/mobxjs/mobx/pull/4709

PR #4612, opened 2026-02-18
fix(4610): computed keepAlive_ 'true' by default

### LINKED ISSUE #4610: Computed values should keepAlive by default
**Intended outcome:** If I create a computed value and access it twice, it should only compute its value once (even if there is no reaction that depends on its value). This is the behavior of other state management libraries and of computed values in the [signals standard proposal](https://github.com/tc39/proposal-signals), so new users would likely expect this of MobX.

**Actual outcome:** Untracked computed values recompute the value each time the value is accessed.

```ts
const computedValue = computed(() => console.log('computed'));
setInterval(() => computedValue.get(), 100);
// Should only log once
```

Passing `{ keepAlive: true }` fixes the issue, but shouldn't that be the default behavior? With #1799 I don't see any downsides to having all computed values be keepAlive.

Note that this would likely be a semver major change.

**Versions**
6.13.7

Tzahile:
### Code change checklist

This is a PR to fix #4610
keepAlive_ default value is changed to `true`.
As mentioned in the issue description:
> With https://github.com/mobxjs/mobx/pull/1799 I don't see any downsides to having all computed values be keepAlive

This PR also 
- changes the warning message and its triggers (trigger now if `keepAlive_` is false)
- updates tests for v4 and v5
- updates docs with new behaviour and examples

### DIFF
--- a/.changeset/big-dots-run.md
+++ b/.changeset/big-dots-run.md
@@ -2,10 +2,16 @@
 "mobx": minor
 ---
 
-Adds a temporary global migration flag for computed `keepAlive` default value:
+Adds a global flag for computed `keepAlive` default value:
 `configure({ globalKeepAliveState })`
 
 Current behavior is preserved by default (`globalKeepAliveState: false`),
 explicit `keepAlive` still takes precedence
+
+Computed keepAlive now resolves in this order:
+
+1. explicit `keepAlive`
+2. `configure({ globalKeepAliveState })`
+3. default `true`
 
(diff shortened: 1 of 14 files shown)


mweststrate:
Hey @Tzahile, sorry for the late follow up. No I don't think changes the default suffices, because this is an option that should truly be very rarely needed, but has a lot of implications on the global working of the package, which will make it more complicated to provide support or keep different packages that use mobx interoperable.

PR #4622, opened 2026-03-10
Use WeakRef for observers

bjb568:
This PR changes the observers of an observable to use a weak reference, which allows computed values with the keepAlive option to be garbage collected.

In the following code, without this PR, memory usage grows without bound. With this PR, buffer is able to be disposed.

```ts
import { observable, computed } from "mobx";

const box = observable.box(1);

setInterval(() => {
  const megabytes = performance.memory.usedJSHeapSize / 1e6;
  console.log(`${megabytes.toFixed(2)} MB used.`);

  const buffer = observable.box(new ArrayBuffer(100 * 1e6));
  computed(() => buffer.get().byteLength * box.get(), {
    keepAlive: true,
  }).get();
}, 1000);
```

With the following code, since it is creating a reaction, it shouldn't be disposed. This is accomplished by having a `globalState.activeReactions` set to prevent garbage collection. Both with and without this PR, memory will grow without bound (autoruns still need to be disposed of).

```ts
import { observable, computed } from "mobx";

const box = observable.box(1);

setInterval(() => {
  const megabytes = performance.memory.usedJSHeapSize / 1e6;
  console.log(`${megabytes.toFixed(2)} MB used.`);

  const buffer = observable.box(new ArrayBuffer(100 * 1e6));
  autorun(() => buffer.get().byteLength * box.get());
}, 1000);
```

Since this PR deals with garbage collection, it's difficult to write unit tests verifying this behavior, so the verification was done in a browser.

### Code change checklist

### DIFF
--- a/packages/mobx/src/core/observable.ts
+++ b/packages/mobx/src/core/observable.ts
@@ -29,7 +29,7 @@ export interface IObservable extends IDepTreeNode {
     lowestObserverState_: IDerivationState_ // Used to avoid redundant propagations
     isPendingUnobservation: boolean // Used to push itself to global.pendingUnobservations at most once per batch.
 
-    observers_: Set<IDerivation>
+    observers_: Set<WeakRef<IDerivation>>
 
     onBUO(): void
     onBO(): void
@@ -42,8 +42,22 @@ export function hasObservers(observable: IObservable): boolean {
     return observable.observers_ && observable.observers_.size > 0
 }
 
-export function getObservers(observable: IObservable): Set<IDerivation> {
+export function getObservers(observable: IObservable): IteratorObject<IDerivation> {
     return observable.observers_
+        .keys()
+        .map(ref => ref.deref())
+        .filter(observer => observer !== undefined)
+}
+
+export function forEachObserver(observable: IObservable, callback: (d: IDerivation) => void) {
+    observable.observers_.forEach(ref => {
+        const d = ref.deref()
+        if (d === undefined) {
+            observable.observers_.delete(ref)
+            return
+        }
+        callback(d)
+    })
 }
 
 // function invariantObservers(observable: IObservable) {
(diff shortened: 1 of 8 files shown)


kubk:
Thanks, but the behavior is intentional: https://github.com/mobxjs/mobx/issues/4610#issuecomment-3997404018

>  keepAlive computed holds references to it's dependencies and to the returned value indefinitely, therefore they can't be garbage collected, potentially causing memory leaks.

packages/mobx/src/core/observable.ts
```diff
@@ -29,7 +30,7 @@ export interface IObservable extends IDepTreeNode {
     lowestObserverState_: IDerivationState_ // Used to avoid redundant propagations
     isPendingUnobservation: boolean // Used to push itself to global.pendingUnobservations at most once per batch.
 
-    observers_: Set<IDerivation>
+    observers_: Set<WeakRef<IDerivation>>
```
natWetzel: consider renaming to observer refs

packages/mobx/src/core/observable.ts
```diff
@@ -42,8 +43,22 @@ export function hasObservers(observable: IObservable): boolean {
     return observable.observers_ && observable.observers_.size > 0
```
natWetzel: should this instead be `return getObservers(observable.observers_.size > 0);`? otherwise this would return true even if it only referred to refs that had been GC'ed

packages/mobx/src/core/observable.ts
```diff
@@ -190,7 +210,7 @@ export function propagateChanged(observable: IObservable) {
     observable.lowestObserverState_ = IDerivationState_.STALE_
 
     // Ideally we use for..of here, but the downcompiled version is really slow...
```
natWetzel: Do we have any other performance concerns?

Issue #4555, opened 2025-05-28
[Feature Request] Allow disabling or evicting of the descriptorCache


pcase-study:
The descriptorCache in [observableObject.ts](https://github.com/mobxjs/mobx/blob/09703d6a3dc7a6214bf2e97cf7e53c690f53d56a/packages/mobx/src/types/observableobject.ts) appears to trade a small amount of memory for performance when initializing an observable, creating one cached descriptor for every unique object key that mobx sees.

We encountered a situation where a library was using a plain java object as a Map, with UUIDs as the key. This wasn't causing any problems for end users, but on our SSR server the infinitely-many different UUIDs resulted in Mobx caching infinitely-many descriptors. This leaked memory with no way to recover it.

While we were able to patch the library to use an actual Map object to coax mobx into using ObservableMap instead of the ObservableObjectAdministration, it would be nice to have a way to either disable the descriptorCache entirely or have a utility method for clearing it.

kubk:
https://github.com/mobxjs/mobx/pull/4628#issuecomment-4095916739

PR #4628, opened 2026-03-17
Add configure option to disable descriptor cache

anchmelev:
This PR adds `useDescriptorCache` to `configure`.

By default, MobX caches property descriptors for observable object keys so repeated keys can reuse the same getter/setter pair. That behavior stays unchanged.

The issue is that in workloads with effectively unbounded dynamic keys, such as UUID-based object dictionaries on SSR servers, the module-level `descriptorCache` can grow without bound and retain descriptors for keys that will never be reused.

This change adds a configuration option to disable descriptor reuse for future observable object properties. When disabled, MobX also clears the existing descriptor cache so already accumulated entries can be released.

`observable.map` is still the preferred model for truly dynamic key-based collections. This change is meant as an escape hatch for cases where switching data structures is not immediately possible.

### Code change checklist

-   [x] Updated `/docs`. Added `useDescriptorCache` to `docs/configuration.md`

Refs: [#4555](https://github.com/mobxjs/mobx/issues/4555)

### DIFF
--- a/.changeset/curly-ladybugs-sparkle.md
+++ b/.changeset/curly-ladybugs-sparkle.md
@@ -1,0 +1,5 @@
+---
+"mobx": minor
+---
+
+Add a configure option to disable descriptor cache reuse for observable object keys.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -196,6 +196,18 @@ Note it doesn't affect existing observables, only the ones created after it's be
 configure({ safeDescriptors: false })
 ```
 
+#### `useDescriptorCache: boolean`
+
+MobX caches property descriptors for observable plain-object keys so repeated keys can reuse the same getter and setter functions. This is usually a small performance optimization, but the cache can grow without bound when observables are created with a very large number of unique property names, such as UUIDs used as object keys during SSR.
+
+`configure({ useDescriptorCache: false })` disables descriptor reuse and clears any descriptors that were cached so far. Existing observable objects keep working; only future observable properties stop using the cache. **Default: `true`**
+
+```javascript
+configure({ useDescriptorCache: false })
+```
+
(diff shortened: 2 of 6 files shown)


anchmelev:
Gentle follow-up on this PR: would appreciate a review when someone has a chance.
This adds an opt-out for descriptor cache reuse to address the unbounded dynamic-key case from #4555, while keeping the default behavior unchanged.

kubk:
@anchmelev Thank you for the PR and digging into this.

While I understand it can be a perf improvement I dont think that's a good direction. The issue here is that the observable plain objects are the wrong data structure for unbounded dynamic keys. We explicitly warn about that in the docs: https://mobx.js.org/observable-state.html

Screenshot 2026-03-20 at 13 19 21

If we need to tell users "for this workload use some extra option to avoid trouble" then i'd rather just tell them to use observable.map instead of a plain object.

`configure` is already at risk of becoming a graveyard of niche escape hatches. Each new public API option adds a maintenance burden in the future

anchmelev:
> @anchmelev Thank you for the PR and digging into this.
> 
> While I understand it can be a perf improvement I dont think that's a good direction. The issue here ins't that descriptorCache is bad, it's that observable plain objects are the wrong data structure for unbounded dynamic keys. We explicitly warn about that in the docs: https://mobx.js.org/observable-state.html
> 
> Screenshot 2026-03-20 at 13 19 21
> If we need to tell users "for this workload use some extra option to avoid trouble" then i'd rather just tell them to use observable.map instead of a plain object.
> 
> `configure` is already at risk of becoming a graveyard of niche escape hatches. Each new public API option adds a maintenance burden in the future

Thanks, that makes sense.

I agree the core issue is using observable plain objects for unbounded dynamic keys, and I understand the concern about adding another niche escape hatch to `configure`.

I’ll drop this approach.

Issue #4594, opened 2025-10-20
eslint react hooks rules do not work when a component is wrapped in observer


javiergonzalez-synth:
This is not really a but report in MobX itself (sorry), but I just wanted to point out that react-hooks eslint rules (other than dependency checking) are not working at all for components wrapped in observer (or wrapped in any function other than react built-ins like memo/forwardRef) for that matter.

There's an (ugly) workaround though, which is wrapping a named function:
```
const Component = observer(function Component() {
  ...
})
```
that perhaps could be added to the docs.

I created a fix 3 weeks ago - https://github.com/facebook/react/pull/34608 - but it seems it didn't gain much traction 😔, which is a shame because these rules are useful even in the mobx-react world (and even worse, I found they stopped working by pure chance)

javiergonzalez-synth:
Thank you for the hint, but I might be missing something... How do the settings from eslint-plugin-react (from some third party developer) affect how eslint-plugin-react-hooks (from meta/react) works?

Does the first modify the eslint AST that the second uses or something similar? 🤔

kubk:
There's also [mobx-react-observer](https://github.com/christianalfoni/mobx-react-observer) that should solve it. It's a Babel/SWC plugin that automatically wraps your components with observer, so you don't have to do it manually.

I've switched my projects to it and it has improved DX in many ways

PR #4630, opened 2026-04-02
fix: clear fieldState errors when resolver uses dot-notation string keys

tomeelog:
Adds a documentation tip to the React integration guide explaining that `eslint-plugin-react-hooks` rules (e.g. `rules-of-hooks`, `exhaustive-deps`) do not work correctly when `observer()` wraps an arrow function expression, because the linter cannot identify it as a React component.

The tip shows the workaround: pass a named function to `observer()` instead.

Addresses #4594

### Code change checklist

### DIFF
--- a/docs/react-integration.md
+++ b/docs/react-integration.md
@@ -405,6 +405,35 @@ Now you can see component names:
 
 </details>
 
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks!

PR #4639, opened 2026-04-30
feat(mobx): make 2022.3 @computed decorator lazy

### LINKED ISSUE #4616: @computed decorator should be lazy
Looking at current decorators internal implementation - it eagerly creates new `ComputedValue` for every decorated property for every new instance

```ts
function decorate_20223_(this: Annotation, get, context: ClassGetterDecoratorContext) {
    if (__DEV__) {
        assert20223DecoratorType(context, ["getter"])
    }
    const ann = this
    const { name: key, addInitializer } = context

    addInitializer(function () {
        const adm: ObservableObjectAdministration = asObservableObject(this)[$mobx]
        const options = {
            ...ann.options_,
            get,
            context: this
        }
        options.name ||= __DEV__
            ? `${adm.name_}.${key.toString()}`
            : `ObservableObject.${key.toString()}`
        adm.values_.set(key, new ComputedValue(options))
    })

    return function () {
        return this[$mobx].getObservablePropValue_(key)
    }
}
```

I think it is quite a waste of memory and performance:

- I think it is safe to say, many of computed properties of some instances will never ever be used
- say we have 1k instances, each have 5 @computed getters
- say on each instance, we only ever use 2 out of those 5 during their lifetime
- as a result - 3000 `ComputedValue` instances were created, while never being used
- those 3000 took some time to create and did occupy some memory before being disposed

What, if decorator would be implemented something like in example below.

Here, we do not add `addInitialize` and create computed value once on the first getter use

- if computed value is never used - `ComputedValue` is never created
- as there is no `addInitialize` - it should have lower performance penantly eg when creating 1000s of new instances with decorators
- AFAIK the only initial runtime cost will be similar internal prototype update during class declaration. Creating class instance will have comparable performance to creating normal, undecorated instance

```ts
import { ComputedValue } from "mobx/dist/internal";

const COMPUTED_SYMBOL = Symbol("computed");
type Key = string | symbol;
type ComputedMap = Record<Key, ComputedValue<any>>;

function initializeComputed(instance: any, key: Key, getter: () => any): ComputedValue<any> {
  // if needed - initialize our map of computed values
  const computedMap: ComputedMap = (instance[COMPUTED_SYMBOL] ??= {});

  // if needed - initialize new `ComputedValue`
  return (computedMap[key] ??= new ComputedValue({
    get: getter,
    context: instance,
  }));
}

export function $computed(get: () => any, context: ClassGetterDecoratorContext) {
  const { name: key } = context;

  return function (this: any) {
    // when used - initialize computed if needed and get its value
    return initializeComputed(this, key, get).get();
  };
}
```

**Intended outcome:**

When creating instance with `@computed` getter, but never using it - `ComputedValue` should never be created

**Actual outcome:**

`ComputedValue` is created even if we never use it

On created instance, check `$mobx` special property - computed value is there.

**Versions**

6.15.0

aschoube:
## Summary

Fixes #4616.

The 2022.3 `@computed` getter decorator currently creates a `ComputedValue` eagerly via `addInitializer` for every decorated getter on every instance — even getters that are never read. On classes with several `@computed` getters across many instances this is wasted allocation and construction work.

This change defers `ComputedValue` creation to the first read of the decorated getter:

- `decorate_20223_` in `computedannotation.ts` now stores a small factory closure on the admin's new `lazyComputedKeys_` map instead of constructing the `ComputedValue` upfront.
- `ObservableObjectAdministration.materializeLazyComputed_(key)` lazily instantiates the `ComputedValue` and inserts it into `values_` on demand. It is called from `getObservablePropValue_` / `setObservablePropValue_` (so the existing decorator getter wrapper Just Works) and from `getAtom` (so `observe(o, key, ...)` materialises before reading the underlying observable).
- `isObservableProp` checks `lazyComputedKeys_` in addition to `values_` so a decorated computed continues to report as observable before its first read.

The legacy `makeObservable`-based path (`defineComputedProperty_`) is unchanged.

## Test plan

- [x] `yarn test` (1036 passing, 14 skipped — full suite green)
- [x] `yarn test:types`
- [x] New regression tests in `stage3-decorators.ts`:
  - `4616 - @computed decorator should be lazy` — asserts `ComputedValue` is not created until first read while `isObservableProp` still returns `true`, and an unused getter never runs.
  - `4616 - observe on @computed before first read materialises it` — asserts `observe(o, key)` works for a lazy computed before any direct read.


ashishkr96:
@kubk  Can you please review this.

mweststrate:
I can take a look on Monday as well, I cleared my schedule to do OSS
maintenance that I didnt get go lately

Op vr 1 mei 2026, 12:39 schreef Ashish kumar choubey <
***@***.***>:

packages/mobx/src/types/observableobject.ts
```diff
@@ -119,10 +120,22 @@ export class ObservableObjectAdministration
     }
 
     getObservablePropValue_(key: PropertyKey): any {
+        this.materializeLazyComputed_(key)
         return this.values_.get(key)!.get()
     }
 
+    materializeLazyComputed_(key: PropertyKey): boolean {
```
mweststrate: It looks like the return value is never uses, so change return type to `void`?

mweststrate: Or better, since this is on a super hot path, would it make sense to _first_ lookup the entry in `values_` *and* return it, and otherwise run the factory, and store, and return the factory result (to avoid another `.get`). This should reduce the amount of map lookups in the common scenario where initializer did already run (or is not a computed), from 2 to 1.

mweststrate (changes requested):
Hey @ashishkr96!

Thanks for this PR. The implementation looks semantically correct, so thanks for taking this effort! I left some comments on how I think this could be faster (since map lookups are relatively expensive)

The whole premise of this PR however is "On classes with several @computed getters across many instances this is wasted allocation and construction work.". Since closures are not free either (they capture several pieces of data in this PR), and there is now an extra map lookup on every hot path of reading an observable; do you have any measurement on the savings actually achieved by this PR? E.g. eithter in memory pressure or runtime speed? For example can you notice a difference when running `yarn perf proxy` in `packages/mobx`? (Or is there a benchmark missing there that would demonstrate the difference)

Thanks!

packages/mobx/src/types/observableobject.ts
aschoube: done in a27ad5bb. `materializeLazyComputed_` now returns the freshly-built `ComputedValue` (or `undefined` when the key isn't lazy), and the callers do:

```ts
const observable = this.values_.get(key) ?? this.materializeLazyComputed_(key)
```

So the hot path (computed already materialised, plain observable, or non-lazy class) is **one** Map lookup. Same fix applied to `setObservablePropValue_` and `getAtom` in `type-utils.ts`. Boolean return is gone too.

ashishkr96:
Hey @mweststrate, thanks for the review!

Both points addressed in a27ad5bb:

### 1. Hot-path collapse (your inline suggestion)
`materializeLazyComputed_` now returns the freshly-built `ComputedValue` (or `undefined` when the key isn't lazy), and call sites do:
```ts
const observable = this.values_.get(key) ?? this.materializeLazyComputed_(key)
```
Hot path is back to **one** `Map.get` (computed already materialised, plain observable, or non-lazy class). Applied to `getObservablePropValue_`, `setObservablePropValue_`, and `getAtom` in `type-utils.ts`. The unused boolean return is gone.

### 2. Benchmarks
Existing `yarn perf proxy` doesn't exercise the `@computed` decorator path (it uses `mobx.computed(fn, { context })` directly), so I wrote a small standalone benchmark on the side to validate the lazy premise — kept it out of the PR diff since it'd add a `tsc` step to the perf workflow; happy to upstream it under `__tests__/perf/` if you'd like.

Shape: 50 000 instances of a class with 10 `@computed` getters where only one is read per instance (the realistic "wide class, sparse access" from #4616). Linux, Node 24, `--expose-gc`, 3 timed runs after one warmup:

| state | construct heap | construct time | first-read time | re-read 5× |
|---|---:|---:|---:|---:|
| `main` (no lazy) | **+304.9 MB** | 316–345 ms | 52–54 ms | 131–200 ms |
| PR original (lazy, 2 lookups) | +153.3 MB | 233–270 ms | 36–40 ms | 140–149 ms |
| PR refactored (lazy, 1 lookup) | +153.3 MB | 245–292 ms | 36–41 ms | 142–155 ms |

So the lazy decorator's actual savings on the #4616 shape:
- **−50 % heap during construction** (305 → 153 MB on 50k×10 getters), since unread `ComputedValue`s are never allocated.
- **~25 % faster construction** for the same reason.
- The 1-lookup refactor matches the 2-lookup variant within run-to-run noise on this microbenchmark — the closure + extra Map probe weren't free, but with the refactor the steady-state read path is identical to `main`'s.

I also ran `yarn perf proxy` 3× on the refactored branch — all 37 tests pass and the headline numbers (order system, array loop, computed memoization, etc.) are well within run-to-run jitter, which is expected since none of those tests touch the decorator path.

Happy to tweak the benchmark shape (different `instances × getters × reads` mix, or a `runInAction` path) — and let me know if you'd prefer the benchmark file checked in.

mweststrate:
@ashishkr96 these benchmarks look great, better than I had intuitively expect! Yes please check in the benchmark. No additional benchmarks needed I think. Also mind marking the change as feat/minor?

After that I'll give it a final review, but thanks for wondering this. 

Did you experiment with doing the same for ObservableValue as well? I suspect it'd wouldn't be as much of a delta since they have less administration, but intrigued now :)

mweststrate:
note: the CI currently failed, but that should be fixed by doing a rebase.

aschoube:
> Did you experiment with doing the same for ObservableValue as well?

 @mweststrate  Yes,  bigger delta than I expected. Quick exploration on a separate branch (`explore/lazy-observable-4616`), same wide-class/sparse-access shape (50k instances × 10 fields, 1 read per instance, 5 re-reads, Linux Node 24, `--expose-gc`, 3 timed runs after warmup):

| state | construct heap | construct time | first-read time | re-read 5× |
|---|---:|---:|---:|---:|
| eager (current `main` behaviour) | **+223.9 MB** | 615.7 ms | 23.9 ms | 82.5 ms |
| lazy `@observable accessor` (experimental) | +40.0 MB | 82.1 ms | 24.7 ms | 23.0 ms |
| **delta** | **−82.1%** | **−86.7%** | within noise | −72% |

The experimental change is one line in `observableannotation.ts` — drop the eager `initializeObservable(this, value)` call from the `init` callback, since the get/set handlers already lazy-materialise on first access (the accessor's backing slot still holds the initial value, so `desc.get.call(this)` returns it correctly).

Caveats — 3 tests in `stage3-decorators.ts` fail with this minimal change, all the same seams the `@computed` PR already closed:
- `getAtom` / `observe(o, key)` before first read — needs a `materializeLazyObservable_` call mirroring `materializeLazyComputed_` in `getAtom`.
- `isObservableProp` — needs to consult a `lazyObservableKeys_` set.
- `values_.keys()` enumeration in inheritance tests — same union with lazy keys.

So the savings are real and the gaps are the same patterns we're already shipping for `@computed`. Happy to open a follow-up PR (`feat(mobx): make @observable accessor lazy`) once #4639 lands — the wins look meaningfully bigger than the `@computed` change (which was −50% heap / −25% construct on the same shape) because the eager `ObservableValue + Atom` per field is heavier than I'd assumed.

Benchmark file used (`__tests__/perf/lazy-observable-decorator.ts`) lives on the explore branch; can include it in the follow-up PR.

aschoube:
Quick correction on the previous comment — I had only run the sparse-access shape (1 of 10 fields read per instance), which is the best case for lazy. Re-ran with **all 10 fields read per instance** to find the floor:

| scenario | construct heap | construct ms | first-read ms | re-read ms |
|---|---:|---:|---:|---:|
| SPARSE eager | 223.9 MB | 595 | 23 | 82 |
| SPARSE lazy  | 40.0 MB | 82 | 25 | 23 |
| FULL eager | 223.9 MB | 551 | 84 | 367 |
| **FULL lazy** | **40.0 MB** | **70** | **289 (+244%)** | 378 |

Lazy doesn't *eliminate* the allocation work, it **defers** it from construct to first-read. On the FULL shape, first-read regresses 3.4× — the deferred allocations land there. Total work (construct + first-read) is still less for lazy on both shapes (~107 ms vs ~618 sparse, ~360 ms vs ~635 full), likely from GC/JIT effects, but the steady-state win shrinks a lot once everything has been read.

Also worth flagging: the experimental change I measured is just the one-line deletion in `init`; the production version would still need the `getAtom`/`isObservableProp`/`values_.keys()` fixups that `@computed` got, plus a per-read branch I haven't measured.

Net read: lazy `@observable` is real but the headline number (−82% heap) is a sparse-access best case, not a steady-state win. Construction-time savings are robust, runtime savings depend on access pattern. Worth a follow-up PR, but framed as "construction-time perf" rather than "general perf win."

aschoube:
@mweststrate When you get some time do look at the benchmark and if this can ship

aschoube:
@kubk Can you help with this in getting it reviewed

mweststrate (approved):

mweststrate:
@ashishkr96 thanks for all the work here, the PR looks great! I'll merge it if the build passes. 

The lazy observable work looks pretty intruiging as well, and better results than I expect (especially on the heap wins, all though the Full lazy uses less heap than expected. Maybe it triggers GC more often which would then explain the much slower reads?). 

If you are willing to put up a separate PR for this, we can discuss more on it if we can find other opportunities to claw back some of the read regression, or conclude this is a worthwile trade-off anyway. 

Thanks again for the work!

aschoube:
Hi @mweststrate — as per your [review comment](https://github.com/mobxjs/mobx/pull/4639#issuecomment-4378567832) on this PR asking me to put up a separate PR for the lazy `@observable accessor` work I explored on `explore/lazy-observable-4616`, I've opened it as #4658.

It mirrors the lazy treatment we shipped here (`lazyObservableKeys_` map + `materializeLazyObservable_` + `getAtom` / `isObservableProp` plumbing), drops the `initializedObjects` WeakSet workaround now that `init` always registers the lazy factory, and includes a `lazy-observable-decorator.ts` perf benchmark alongside the @computed one. Full suite green, types clean.

Happy to discuss whether the FULL-shape read regression is worth clawing back — left some notes in the PR body and benchmark table.

Issue #3796, opened 2023-11-14
MobX 7 roadmap


mweststrate:
[WIP for now, needed the issue number]

* [ ] Remove legacy decorators support
* [ ] Remove deprecated APIs
   * [ ] ...
* [ ] Remove mobx-react / observer class support?
* [ ] Remove Provider/inject
* [ ] Remove deprecated mobx-react-lite hooks
* [ ] ESM support
* [ ] ....
* [ ] Remove trace
* [ ] Remove intercept / observer
* [ ] Remove legacy array support (non-proxied)
* [ ] Remove some configuration options?
* [ ] Modern build / test system
* [ ] Remove v4 unit tests

pzmosquito:
awesome news! It looks like the new version will be focused on removing deprecated stuff and updating the build system instead of introducing new features or breaking changes. It should be a breeze to upgrade!

kubk:
I have a somewhat controversial suggestion: automatically bind class methods in `makeAutoObservable`/`makeObservable`. The idea is to set `{ autoBind: true }` as the default, while still offering the option to disable it with `{ autoBind: false }`.

Rationale: People might forget to bind `this`. Here are a few examples:
- https://github.com/mobxjs/mobx/discussions/3795#discussioncomment-7567426
- https://github.com/mobxjs/mobx/discussions/2496#discussioncomment-92094

Developers these days don't usually deal with `this`, `.call`, or `.apply`, as was common before. Hence, the cause of the issue isn't immediately clear to them. They may experience loss of reactivity and attribute it to Mobx. While there might be reasons to preserve the original `this` by default, I fail to see them. All of my Mobx stores use `{ autoBind: true }`. Also, though I'm not a Vue expert, it seems they bind `this` as well: [Vue.js Computed Properties](https://v2.vuejs.org/v2/guide/computed.html).

```javascript
  computed: {
    // a computed getter
    reversedMessage: function () {
      // `this` points to the vm instance
      return this.message.split('').reverse().join('')
    }
  }
  
Perhaps we could even consider making new `@action` / `@flow` decorators bound by default, as well as introducing `@action.unbound` / `@flow.unbound`. I can work in this PR.

kiejo:
As a long-time user of MobX I would be in favor of keeping support for mobx-react / observer classes as it's still fully supported by React. I understand why new code bases might not use class based components, but for existing code bases it doesn't necessarily make sense to migrate all existing code if it still works well.

mweststrate:
[...]

PR #4646, opened 2026-05-08
Switch from Lerna and Yarn 1 to NPM

kubk:
The repo only used Lerna for very limited set of features which npm workspaces can cover directly. The workspace list is now explicit so npm could run package scripts in the required dependency order.

Yarn 1 is in maintenance mode, while npm is already bundled with the Node used in CI, so this reduces maintainer setup and avoids carrying 2 package managers + Lerna for something npm can reproduce

Related issue: https://github.com/mobxjs/mobx/issues/3796

@mweststrate Does the direction feel right for you? The goal is to simplify & modernize the tooling


mweststrate:
Yes, this looks great and been on my mental to-modernize list for a long time!

PR #4661, opened 2026-06-04
Fix Stage 3 decorator inheritance

js2me:
### Code change checklist

-   [x] Added/updated unit tests
-   [ ] Updated `/docs`. For new functionality, at least `API.md` should be updated
-   [x] Verified that there is no significant performance drop (`npm -w mobx run test:performance`)

Closes: https://github.com/mobxjs/mobx/issues/4660   

<details>
  <summary>MobX Benchmark Comparison </summary>

| | Branch A | Branch B |
|---|---|---|
| Branch | `fix/computed-lazy-inheritance-cycle` | `main` |

---

## Branch: fix/computed-lazy-inheritance-cycle
Date: 2026-06-06T12:28:31+03:00
Node: v24.16.0
Commit: 0cc041bf

### Run 1

Lazy @computed decorator benchmark — 50000 instances × 10 @computed getters, 1 read/instance, 5 re-reads.
Node v24.16.0, 3 timed runs after 1 warmup.

run | construct heap MB | construct ms | first-read ms | re-read ms
----+-------------------+--------------+---------------+-----------
  1 |    154.4          |    347.1     |     83.4      |    155.2
  2 |    154.4          |    366.4     |     81.0      |    162.7
  3 |    154.4          |    361.6     |     81.0      |    161.3
avg |    154.4          |    358.4     |     81.8      |    159.7

Lazy @observable benchmark — 50000 instances, 5 re-reads.
Node v24.16.0, 3 timed runs after 1 warmup per scenario.
Stage-3 decorators only; no makeObservable/makeAutoObservable.

DEFAULTS: sparse 1 of 10 fields read
run | construct heap MB | construct ms | first-read ms | re-read ms | write ms
----+-------------------+--------------+---------------+------------+---------
  1 |    118.2          |    181.8     |     31.9      |     47.8     |         
  2 |    118.3          |    195.4     |     32.1      |     44.4     |         
  3 |    118.2          |    171.8     |     27.3      |     45.8     |         
avg |    118.2          |    183.0     |     30.4      |     46.0     |         

DEFAULTS: 3 of 10 fields read, then write unread f9
run | construct heap MB | construct ms | first-read ms | re-read ms | write ms
----+-------------------+--------------+---------------+------------+---------
  1 |    118.2          |    218.9     |     51.4      |     98.8     |     77.1
  2 |    118.2          |    190.1     |     56.8      |     93.0     |     70.1
  3 |    118.2          |    194.2     |     56.3      |     90.7     |     73.8
[...]

PR #4662, opened 2026-06-05
fix(mobx): prevent cycle when @computed overrides a base makeObservable computed (#4660)

### LINKED ISSUE #4660: MobX 6.16.0 Cycle Detection Bug
**Intended outcome:**

When a base class uses `makeObservable` to annotate a computed property, and a child class overrides that same property with the `@computed` 2022.3 decorator, the computed property should be accessible without errors — as it was in MobX 6.15.x.

```ts
class Base {
  _value = false;
  get value() { return this._value; }

  constructor() {
    makeObservable(this, { _value: observable, value: computed });
  }
}

class Child extends Base {
  @computed
  override get value() { return super.value; }
}

new Child().value; // Should return false
```

**Actual outcome:**

Starting from MobX 6.16.0, accessing the overridden computed property throws:

```
[MobX] Cycle detected in computation
```

The `@computed` decorator replaces the prototype getter with a replacement that calls `getObservablePropValue_(key)`. When `makeObservable` runs in the base class constructor, it walks the prototype chain, finds this replacement getter on the child's prototype, and creates a `ComputedValue` using it as the derivation. That derivation calls `getObservablePropValue_(key)` on the same key — creating a self-referencing cycle.

Reproduction repo: https://github.com/js2me/mobx-computed-super-decorators-bug

1. `npm install`
2. `npm test`
3. Observe `[MobX] Cycle detected in computation` errors

The repo includes two failing scenarios:
- **Simple inheritance**: `Base` with `makeObservable({ value: computed })` + `Child` with `@computed override get value()`
- **Three-level inheritance**: `ViewModelBase` with `makeObservable({ isMounted: computed })` → `RouteViewModel` (plain getter override) → `ProductsPageVM` with `@computed override get isMounted()`

It also includes two passing scenarios that confirm the cycle only happens when both `makeObservable` and `@computed` target the same key:
- `@computed` on both parent and child (no `makeObservable`) — works fine
- `@computed` only on child, parent does NOT annotate the key as computed — works fine

**Root cause:**

In 6.15.x, the `addInitializer` from `@computed` eagerly created a new `ComputedValue` with the original getter and overwrote the cyclic one in `adm.values_`. In 6.16.0 (PR #4639), the initializer registers a lazy factory instead (`adm.lazyComputedKeys_`). But `getObservablePropValue_` checks `values_` first — where the cyclic `ComputedValue` from `makeObservable` already sits — so the lazy factory is never materialized.

Although `computedannotation.ts` line 68 does `adm.values_.delete(key)` in the initializer, the ordering is such that `makeObservable` in the base class constructor creates the cyclic entry after the initializer has already run, so the delete doesn't help.

**Workaround:**

Remove `@computed` from the child class override. The base class's `makeObservable` already walks the prototype chain and picks up the child's overridden getter automatically, so re-annotating with `@computed` is unnecessary:

```ts
// Instead of:
class Child extends Base {
  @computed
  override get value() { return super.value; }
}

// Use:
class Child extends Base {
  override get value() { return super.value; }
}
```

**Versions**

- MobX: 6.16.0 (regression from 6.15.x, introduced by PR #4639)

aschoube:
### Problem

Starting from MobX 6.16.0 (regression from #4639), accessing a `@computed` (2022.3 decorator) override of a property that a base class annotated as `computed` via `makeObservable` throws:

```
[MobX] Cycle detected in computation
```

```ts
class Base {
  _value = false
  get value() { return this._value }
  constructor() { makeObservable(this, { _value: observable, value: computed }) }
}

class Child extends Base {
  @computed
  override get value() { return super.value } // throws on access
}

new Child().value
```

### Root cause

The `@computed` decorator replaces the prototype getter with a replacement that calls `getObservablePropValue_(key)`, and (since #4639) registers a **lazy** `ComputedValue` factory in `adm.lazyComputedKeys_`.

When the base class's `makeObservable(this, { value: computed })` runs in `super()`, it walks the prototype chain, finds the child's **replacement getter**, and eagerly builds a `ComputedValue` from it — storing it directly in `adm.values_`. That derivation calls `getObservablePropValue_(key)` on the same key, so it references itself → cycle.

`getObservablePropValue_` checks `values_` before materialising the lazy factory, so the stale cyclic entry always wins. In 6.15.x the eager initializer happened to overwrite it; the lazy factory no longer does.

### Fix

In the `@computed` 2022.3 initializer, delete any stale `values_` entry for the key before registering the lazy factory, so the factory (built from the **original** getter) wins on first access. The lazy-computed optimization is preserved.

### Tests

Adds regression tests to `stage3-decorators.ts` covering:
- simple inheritance (`Base` makeObservable computed + `Child` `@computed` override)
- three-level inheritance with an intermediate plain-getter override

Both assert reactivity still works after the fix. They fail without the fix (`[MobX] Cycle detected in computation`) and pass with it.

Verified locally: full 2022.3 decorator suite (46 passed / 4 skipped) and full main suite (1000 passed / 10 skipped, 47 suites) green.

### DIFF
--- a/.changeset/fix-computed-override-cycle.md
+++ b/.changeset/fix-computed-override-cycle.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix regression from #4639 where a `@computed` override of a base-class `makeObservable` computed threw "[MobX] Cycle detected in computation" (#4660)
--- a/packages/mobx/__tests__/decorators_20223/stage3-decorators.ts
+++ b/packages/mobx/__tests__/decorators_20223/stage3-decorators.ts
@@ -1319,3 +1319,59 @@ test(`decorated field can be inherited, but doesn't inherite the effect of decor
     const subStore = new SubStore()
     expect(isAction(subStore.action)).toBe(false)
 })
+
+test("issue 4660 - @computed override of a base makeObservable computed doesn't cycle", () => {
+    class Base {
+        _value = false
+        get value() {
+            return this._value
+        }
+        constructor() {
+            makeObservable(this, { _value: observable, value: computed })
+        }
+    }
+
+    class Child extends Base {
+        @computed
+        override get value() {
+            return super.value
+        }
+    }
+
+    const child = new Child()
+    // Used to throw "[MobX] Cycle detected in computation": base makeObservable
+    // built a ComputedValue from the child's @computed replacement getter.
+    expect(child.value).toBe(false)
+    child._value = true
+    expect(child.value).toBe(true)
+})
+
+test("issue 4660 - three-level inheritance with intermediate plain override", () => {
+    class ViewModelBase {
+        _mounted = false
+        get isMounted() {
+            return this._mounted
+        }
+        constructor() {
+            makeObservable(this, { _mounted: observable, isMounted: computed })
+        }
+    }
+
+    class RouteViewModel extends ViewModelBase {
+        override get isMounted() {
+            return super.isMounted
+        }
+    }
+
+    class ProductsPageVM extends RouteViewModel {
+        @computed
+        override get isMounted() {
+            return super.isMounted
+        }
+    }
+
+    const vm = new ProductsPageVM()
+    expect(vm.isMounted).toBe(false)
+    vm._mounted = true
+    expect(vm.isMounted).toBe(true)
+})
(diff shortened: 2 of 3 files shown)


kubk:
Thanks, there's a PR already: https://github.com/mobxjs/mobx/pull/4661

PR #4670, opened 2026-06-09
Fix optional react-dom loading in mobx-react-lite

### LINKED ISSUE #4663: `mobx-react-lite` unconditionally imports `react-dom` even though it is optional in `package.json`
**Intended outcome:**

`mobx-react-lite` should not crash at module load time when `react-dom` is not installed/resolvable, because `react-dom` is marked as optional in `peerDependenciesMeta`.

At minimum, I would expect one of these to be true:
- `react-dom` is required, rather than optional, if it is needed unconditionally at runtime, or
- `mobx-react-lite` avoids importing `react-dom` eagerly and falls back gracefully when it is unavailable

---

More specifically, I was trying to use `@redocly/cli` to build REST API documentation from an `openapi.yaml` file. The specific command being used doesn't matter here. `@redocly/cli` depends on `mobx-react-lite`, but it also requires `react-dom` so this issue doesn't normally occur. However, when I added `@asyncapi/cli` to my `packages.json` (for generating documentation from an `asyncapi.yaml` file), the redocly command started throwing the error in the "Actual outcome" section. This is because npm hoisted `react`, `mobx`, and `mobx-react-lite` to the root of my `node_modules` (due to `@asyncapi/cli` depending on some of them), but it did not hoist `react-dom`, so when `mobx-react-lite` is loaded, it can't find `react-dom` (since that's still under `@redocly/cli`), resulting in the error.

**Actual outcome:**

Importing `mobx-react-lite` can throw immediately with:
```bash
Error: Cannot find module 'react-dom'
```

In the published package, `react-dom` is marked optional in `package.json`, but the CJS build appears to import it unconditionally at the top level:
- `package.json` contains `peerDependenciesMeta.react-dom.optional = true`
- `dist/mobxreactlite.cjs.development.js` contains `var reactDom = require('react-dom');`
Which means that the package can fail to load in a valid install layout where `react-dom` is not present/resolvable from `mobx-react-lite`'s location.

> Note:  I'm not making a MRE based on the exact usage that led me to encounter this bug, since there are more dependencies for using Redocly (like Java) and having Redocly isn't necessary to make the bug manifest.

In an empty directory:
```bash
npm init -y
npm install react mobx mobx-react-lite
node -e "require('mobx-react-lite')"
```

Observed result:
```bash
Error: Cannot find module 'react-dom'
```

**Versions**

- `mobx-react-lite`: 4.1.1
- `mobx`: 6.16.0
- `react`: 19.2.7
- `node`: 24.16.0
- `npm`: 11.16.0

cyphercodes:
Fixes #4663

### Summary
- Replace the static `react-dom` re-export in `mobx-react-lite` with a guarded optional load.
- Fall back to the existing no-op batch scheduler when `react-dom` is unavailable.
- Add a regression test that verifies `mobx-react-lite` can load when `react-dom` cannot be resolved.

### Code change checklist

-   [x] Updated `/docs` (not applicable: bug fix, no docs/API change)
-   [x] Verified that there is no significant performance drop (not applicable to core MobX performance; ran targeted package checks and size test)

### Verification
- `npm ci --ignore-scripts`
- `npm -w mobx run build:test`
- `npm -w mobx-react run build:test`
- `npm -w mobx-react-lite run lint`
- `npm -w mobx-react-lite run test:types`
- `npm -w mobx-react-lite run build:test`
- `npm -w mobx-react-lite test -- --runInBand`
- `npm -w mobx-react-lite run test:size`
- `npx prettier --check packages/mobx-react-lite/src/utils/reactBatchedUpdates.ts packages/mobx-react-lite/__tests__/optionalReactDom.test.ts`
- `git diff --check`
- Built CJS smoke test with `react-dom` module loading blocked

### DIFF
--- a/.changeset/optional-react-dom-batching.md
+++ b/.changeset/optional-react-dom-batching.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+Avoid requiring optional `react-dom` during module initialization and fall back to no-op batching when it is unavailable.
--- a/packages/mobx-react-lite/__tests__/optionalReactDom.test.ts
+++ b/packages/mobx-react-lite/__tests__/optionalReactDom.test.ts
@@ -1,0 +1,15 @@
+afterEach(() => {
+    jest.dontMock("react-dom")
+    jest.resetModules()
+})
+
+test("does not fail to load when react-dom is unavailable", () => {
+    jest.resetModules()
+    jest.doMock("react-dom", () => {
+        throw Object.assign(new Error("Cannot find module 'react-dom'"), {
+            code: "MODULE_NOT_FOUND"
+        })
+    })
+
+    expect(() => require("../src/index.ts")).not.toThrow()
+})
--- a/packages/mobx-react-lite/src/utils/reactBatchedUpdates.ts
+++ b/packages/mobx-react-lite/src/utils/reactBatchedUpdates.ts
@@ -1,1 +1,19 @@
-export { unstable_batchedUpdates } from "react-dom"
+import { defaultNoopBatch } from "./observerBatching"
+
+declare const require:
+    | ((moduleName: string) => { unstable_batchedUpdates?: typeof defaultNoopBatch })
+    | undefined
+
+export let unstable_batchedUpdates = defaultNoopBatch
+
+if (typeof require === "function") {
+    try {
+        const reactDom = require("react-dom")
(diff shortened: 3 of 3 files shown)


kubk:
Thank you for the PR. As described in the issue comments - the problem shouldn't be solved at all as React 18+ provides its own batching. Mobx v7 will drop React <=17 support: https://github.com/mobxjs/mobx/pull/4671

PR #4672, opened 2026-06-13
fix: ObservableSet.replace only emits events for actual changes

### LINKED ISSUE #3761: Calling "replace" on an `ObservableSet` always changes events
Current Behaviour:

`ObservableSet.replace(...)` necessarily emits a single DELETE event for every entry in the set and then a single ADD event for all the entries in the provided source.

Expected Behaviour:
`ObservableSet.replace(...)` only emits DELETE events for those items in the set that are not in the provided source and then only emits ADD events for those new entries from the source.

versions:
```
mobx -> 6.10.9
```

```tsx
import { observable, configure } from "mobx";

configure({
  enforceActions: "never"
});

const mySet = observable.set<string>(["some-initial-state"]);

mySet.observe(console.log);

// This should trigger an ADD event
mySet.add("some-second-element");

// This should trigger an ADD event
mySet.add("some-thrid-element");

// This should trigger a DELETE event
mySet.delete("some-initial-state");

// This should trigger no events, since it is in the same order and the same elements
mySet.replace(["some-second-element", "some-third-element"]);
```

chatman-media:
Closes #3761

## Problem

`ObservableSet.replace(...)` was implemented as `clear()` followed by re-adding every value of the source. As a result it always emitted a `delete` event for **every** existing element and an `add` event for **every** replacement element — even for elements that were unchanged — and unnecessarily retriggered reactions.

```ts
import { observable, configure } from "mobx"
configure({ enforceActions: "never" })

const s = observable.set(["a", "b", "c"])
s.observe(console.log)

// Before: delete a, delete b, delete c, add b, add c, add d  (6 events)
// After:  delete a, add d                                     (2 events)
s.replace(["b", "c", "d"])

// Before: delete x, delete y, add x, add y  (4 events)
// After:  (no events)
observable.set(["x", "y"]).replace(["x", "y"])
```

This mirrors the long-standing complaint that `ObservableMap.replace` already handles correctly (see #1243 / #1980 in the map tests).

## Fix

`replace` now:
- deletes only the values that are **not** part of the replacement, and
- adds only the values from the replacement (which are no-ops when already present).

`add`/`delete` already short-circuit (and emit no event / no `reportChanged`) when a value is respectively already present or already absent, so this is sufficient to make `replace` emit events only for genuine changes. The approach matches `ObservableMap.replace`.

## Tests

Added a `#3761` describe block in `set.js` covering:
- replace with array / ES6 `Set` / observable `Set` only emits events for removed & added values
- replace with identical content emits no events and does not report a change to reactions
- interceptors are still honored during `replace`

These tests fail on `main` and pass with the fix. The existing `set crud` test (which exercises `replace` with disjoint sets) still passes unchanged. Added a changeset (`patch`).


packages/mobx/src/types/observableset.ts
```diff
@@ -296,17 +293,27 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
-                die("Cannot initialize set from " + other)
-            }
-        })
+        if (Array.isArray(other) || isES6Set(other)) {
+            // Only emit `delete`/`add` events (and `reportChanged`) for values that
+            // actually change, instead of clearing and re-adding everything. `add` and
+            // `delete` are already no-ops for values that are respectively already
+            // present or already absent, so we just need to avoid deleting values that
+            // are part of the replacement. See #3761.
+            transaction(() => {
```
mweststrate, line 302: This generally looks great! Note that this seems to be making a semantic change:
- Sets to preserve insertion order (per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#description), the old behavior would preserve the insertion order of the new set (e.g. `{a,b,c}.replace({d,b,a})` would result in a set that is iterated like `d,b,a`. With this change, the iteration would become `a,b,d`. Now I think a good argument could be made that is semantically more accurate (and thus a bug fix), but it is a observable change of behavior, so we should mention it clearly in the release notes
- The new behavior now does to two iterations (first loop the old set to delete items, then loop the new set to add items), so this is likely a bit slower. I think this is fine. But it would be great to short-circuit the scenario where either the `this` or `replacement` is empty.

packages/mobx/src/types/observableset.ts
```diff
@@ -296,17 +293,27 @@ export class ObservableSet<T = any> implements Set<T>, IInterceptable<ISetWillCh
...
+                // Collect the desired values for quick lookup (dedupes the source).
+                const replacementValues = new Set<T>(other as Iterable<T>)
```
mweststrate: if `other` is already a Set, let's not allocate another one since it isn't mutated anyway.

mweststrate (changes requested):
Thanks for making this improvement! I left some notes how the implementation can be slightly improved, but overall looks solid to me!

packages/mobx/__tests__/base/set.js
```diff
@@ -512,3 +512,102 @@ describe("Observable Set interceptors", () => {
...
+    test("replace only emits delete/add for removed/added values", () => {
+        const s = set(["a", "b", "c"])
+        const events = []
+        mobx.observe(s, change => {
+            delete change.observableKind
+            delete change.debugObjectName
+            events.push(change)
+        })
+
+        s.replace(["b", "c", "d"])
```
mweststrate: if you'd reorder the items in this array it would verify the change in behavior is discussed below

packages/mobx/__tests__/base/set.js
```diff
@@ -512,3 +512,102 @@ describe("Observable Set interceptors", () => {
...
+    test("replace with an ES6 Set only emits events for actual changes", () => {
+        const s = set([1, 2, 3])
+        const events = []
+        mobx.observe(s, change => {
+            delete change.observableKind
+            delete change.debugObjectName
+            events.push(change)
+        })
+
+        s.replace(new Set([2, 3, 4]))
```
mweststrate: Same

chatman-media:
Thanks for the thorough review! Addressed the notes in 57c170d:

- **Iteration order** — agreed this is an observable change (and, as you say, arguably the more correct one). I did not try to preserve the old ordering; surviving values keep their original position and new values are appended. Documented this clearly in the changeset, with the `{a,b,c}.replace({d,b,a})` -> `a,b,d` example.
- **Two iterations / short-circuit** — added fast paths: an empty replacement is just a `clear()`, and replacing into an empty set only runs the adds, so the common trivial cases skip the second loop entirely.
- **Avoid the extra Set allocation** — when `other` is already a Set it is now reused directly for the membership checks (observable sets are snapshotted earlier, and the Set is only read, never mutated). Arrays are still wrapped so they get deduped.
- **Tests** — reordered the array and ES6-Set replacements so they assert the resulting iteration order, which exercises the documented behavior change. The existing `set crud` test still passes unchanged.

chatman-media:
Thanks for the review, @mweststrate. Addressed in 57c170d:

- documented the observable iteration-order change in the changeset
- reused existing Set inputs instead of allocating another Set
- added fast paths for empty replacement / empty current set
- updated the tests to lock in the new order semantics

Full set.js suite is green locally. Happy to adjust the order semantics if you'd prefer matching the replacement order instead.

PR #4676, opened 2026-06-22
fix: Symbol.toPrimitive of a null observable no longer throws on ES5 targets

### LINKED ISSUE #4673: ObservableValue[Symbol.toPrimitive] returns null, which throws in ES5-transpiled code (Babel _toPrimitive helper)
### Intro

When an observable holds `null`, its `Symbol.toPrimitive` implementation returns `null`. Native engines accept this (`null` is a valid primitive per spec), but **ES5-transpiled consumer code** routes the coercion through Babel's inlined `_toPrimitive` helper, which checks `typeof result !== "object"`. Because `typeof null === "object"`, the helper misclassifies `null` as an object and throws:

```
TypeError: @@toPrimitive must return a primitive value.
```

This breaks any string/primitive coercion of an observable that currently holds `null` (e.g. `` `${box}` ``, template strings, object keys) on older targets — in our case old Chrome on Smart TVs.

### How to reproduce

mobx **6.16.1**.

```js

// This is exactly what Babel inlines into consumer code when targeting ES5:
function _toPrimitive(t, r) {
  if ("object" != typeof t || !t) return t;
  var e = t[Symbol.toPrimitive];
  if (void 0 !== e) {
    var i = e.call(t, r || "default");
    if ("object" != typeof i) return i;
    throw new TypeError("@@toPrimitive must return a primitive value.");
  }
  return ("string" === r ? String : Number)(t);
}
function _toPropertyKey(t) {
  var i = _toPrimitive(t, "string");
  return "symbol" == typeof i ? i : i + "";
}

const box = observable.box(null);

String(box);            // → "null"  (native ToPrimitive, OK)
_toPropertyKey(box);    // → throws TypeError (Babel-transpiled path)
```

### Root cause (in mobx)

`ObservableValue` exposes:

```js
_proto.valueOf = function () { return toPrimitive(this.get()); };
_proto[Symbol.toPrimitive] = function () { return this.valueOf(); };
```

and mobx's internal `toPrimitive` is:

```js
function toPrimitive(value) {
  return value === null ? null : typeof value === "object" ? "" + value : value;
}
```

So for `get() === null` the `Symbol.toPrimitive` method returns `null`.

### Whose bug is it, really?

Strictly, this is a known limitation of Babel's `_toPrimitive` helper (it uses `typeof x === "object"` instead of a real "is Object" check, so it mishandles `null`). The helper is unchanged even in the latest Babel (`@babel/runtime@8.0.0`), and `null` is a valid `Symbol.toPrimitive` result per the [spec](https://tc39.es/ecma262/#sec-toprimitive), so mobx is not violating the standard.

**However**, mobx is in a unique position to make this defensive, because:

- mobx's own `toPrimitive` already special-cases values (objects → `"" + value`), so returning a non-`null` primitive for `null` would be consistent with that intent.
- Returning `null` from `Symbol.toPrimitive` is technically legal but practically hazardous: any consumer transpiling to ES5 (still common for Smart TV / legacy targets) will hit the throw. A tiny change in mobx fixes it for *all* such consumers, whereas the Babel helper realistically won't change.

### Suggested fix

Make the `Symbol.toPrimitive` result never be `null` (mirror native `String(null)` semantics), e.g.:

```js
_proto[Symbol.toPrimitive] = function (hint) {
  var v = this.valueOf();
  return v === null ? "null" : v;
};
```

(Exact replacement value is debatable — `"null"` matches native `String(null)` for the common stringification case. A hint-aware version could return `"null"` for `"string"` and `0`/`NaN` for `"number"` to better mirror native behavior, but that may be overkill.)

### Workaround we currently use

We patch the `ObservableValue` prototype at runtime (wrap `Symbol.toPrimitive` so it returns `"null"`/`"undefined"` instead of `null`/`undefined`) — version-independent, no node_modules patching.

### Environment

- mobx: 6.16.1
- Consumer build: transpiled to ES5 (Babel) for old Chrome on Smart TV
- Native (modern Chrome): works fine; only transpiled path throws

MahinAnowar:
Closes #4673.

## Problem

`ObservableValue` and `ComputedValue` implement `[Symbol.toPrimitive]` by returning `valueOf()`:

```ts
valueOf(): T { return toPrimitive(this.get()) }
[Symbol.toPrimitive]() { return this.valueOf() }
```

When the held value is `null`, `valueOf()` returns `null`. Native engines accept that (`null` is a valid `ToPrimitive` result per spec), but **ES5-transpiled consumer code** routes coercion through Babel's inlined `_toPrimitive` helper:

```js
if ("object" != typeof i) return i;
throw new TypeError("@@toPrimitive must return a primitive value.");
```

Because `typeof null === "object"`, the helper misclassifies `null` as an object and throws. So coercing a null-holding observable (template strings, object keys, etc.) throws on ES5 targets — e.g. `` `${box}` `` / `obj[box]` on old Chrome / Smart-TV builds.

```js
const box = observable.box(null)
String(box)         // "null"  (native, fine)
_toPropertyKey(box) // throws TypeError on the Babel ES5 path
```

This is technically a limitation of Babel's helper (unchanged even in `@babel/runtime@8`), but mobx is well-placed to be defensive here, and its own `toPrimitive` util already special-cases values.

## Fix

Make `[Symbol.toPrimitive]` hint-aware and never return `null`, mirroring native coercion of `null`:

```ts
[Symbol.toPrimitive](hint) {
  const value = this.valueOf()
  if (value === null) return hint === "number" ? 0 : "null"
  return value
}
```

`String(null) === "null"` and `Number(null) === 0`, so the coerced results match the native path (and `+box` of a null observable stays `0` rather than becoming `NaN`). `valueOf()` and `toJSON()` are left untouched, so numeric coercion and `JSON.stringify` are unchanged. The same fix is applied to `ComputedValue`, which shares the identical coercion path.

## Tests

Adds a regression test (`#4673 …`) that inlines Babel's exact ES5 `_toPrimitive`/`_toPropertyKey` helpers and asserts that coercing a null-holding observable **and** computed no longer throws and yields `"null"` / `0`, while non-null values still coerce as before. It throws (fails) on the current code and passes with this change. The full `observables` suite passes and `tsc` is clean. A changeset is included.

### DIFF
--- a/.changeset/toprimitive-null-es5.md
+++ b/.changeset/toprimitive-null-es5.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix `ObservableValue`/`ComputedValue` `[Symbol.toPrimitive]` returning `null` when the value is `null`. Native engines accept this, but ES5-transpiled consumer code routes coercion through Babel's inlined `_toPrimitive` helper, which treats `null` as an object (`typeof null === "object"`) and throws `TypeError: @@toPrimitive must return a primitive value.`. The method is now hint-aware and mirrors native coercion of `null` (`String(null) === "null"`, `Number(null) === 0`), so coercing a null-holding observable no longer throws on ES5 targets.
--- a/packages/mobx/__tests__/base/observables.js
+++ b/packages/mobx/__tests__/base/observables.js
@@ -1693,6 +1693,47 @@ test("#561 test toPrimitive() of observable objects", function () {
     }
 })
 
+test("#4673 Symbol.toPrimitive of a null observable does not throw in Babel's ES5 _toPrimitive helper", function () {
+    // Exactly what Babel inlines into ES5-transpiled consumer code. It uses
+    // `typeof x === "object"` to test for a primitive, so it misclassifies the
+    // spec-legal `null` result and throws.
+    function _toPrimitive(t, r) {
+        if ("object" != typeof t || !t) return t
+        var e = t[Symbol.toPrimitive]
+        if (void 0 !== e) {
+            var i = e.call(t, r || "default")
+            if ("object" != typeof i) return i
+            throw new TypeError("@@toPrimitive must return a primitive value.")
+        }
+        return ("string" === r ? String : Number)(t)
+    }
+    function _toPropertyKey(t) {
+        var i = _toPrimitive(t, "string")
+        return "symbol" == typeof i ? i : i + ""
+    }
+
+    const box = observable.box(null)
+
+    // The transpiled coercion path must not throw and must mirror native
+    // coercion of `null` (String(null) === "null", Number(null) === 0).
+    expect(() => _toPropertyKey(box)).not.toThrow()
+    expect(_toPropertyKey(box)).toBe("null")
+    expect(_toPrimitive(box, "number")).toBe(0)
+
+    // Native ToPrimitive is and stays unaffected.
+    expect(String(box)).toBe("null")
+
+    // Non-null values still coerce as before.
+    box.set(5)
+    expect(_toPropertyKey(box)).toBe("5")
+    expect(+box).toBe(5)
+
+    // computed values share the same coercion path.
+    const c = computed(() => null)
+    expect(() => _toPropertyKey(c)).not.toThrow()
+    expect(_toPropertyKey(c)).toBe("null")
+})
+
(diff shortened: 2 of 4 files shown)


kubk:
https://github.com/mobxjs/mobx/issues/4673#issuecomment-4765550601

PR #4680, opened 2026-06-28
feat(mobx-react-lite): add support passing propsAreEqual fn for obser…

js2me:
…ver HOC

### Code change checklist

-   [ ] Verified that there is no significant performance drop (`npm -w mobx run test:performance`)

### DIFF
--- a/.changeset/quiet-stars-sing.md
+++ b/.changeset/quiet-stars-sing.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+support passing `propsAreEqual` for `observer()` HOC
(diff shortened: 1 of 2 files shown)


kubk:
@js2me Thank you for the contribution. Could you share an example of when this option would be useful? I'm worried it might not be very idiomatic from the MobX perspective

mweststrate (changes requested):
Hey! 

Agreeing with @kubk  that this would require some more motivation, as this feature might easily be confusing or misleading to developers, as that function is typically used for a different mental model. Mind elaborating a bit more on why/when this would be useful?

js2me:
@kubk @mweststrate agreed with you

Issue #3919, opened 2024-09-05
`ObservableSet.intersection` triggers "Maximum call stack size exceeded" stack overflow


travismfrank:
**Intended outcome:**

Calling the newly implemented `ObservableSet.intersection` with another `ObservableSet` as the argument causes a stack overflow error.  This is because an `ObservableSet` argument passes `isES6Set`, thus calling `intersection` on _that_ set with the first set as the argument.  Repeat ad infinitum until V8 decides it's had enough.

`observableSet.intersection(otherObservableSet)` should just work.

**Actual outcome:**

"Maximum call stack size exceeded" in Chrome.  Stack overflow.

**How to reproduce the issue:**

```typescript
const foo = observable.set(['a', 'b', 'c']);
const bar = observable.set(['c', 'd', 'e']);

const shared = foo.intersection(bar);
```

[...]

PR #4681, opened 2026-07-02
fix(mobx): ObservableSet union/intersection/symmetricDifference result ordering

spokodev:
`ObservableSet` implements `Set<T>`, so its ES2024 set methods should match native `Set` result ordering: `union` appends the argument's new elements after the receiver's, and `intersection` and `symmetricDifference` keep the receiver's order. When the argument is a plain (non-observable) `Set`, `union`, `intersection` and `symmetricDifference` instead delegate to the argument (`otherSet.method(this)`), so the result comes out in the argument's order.

```js
const s = observable.set([1, 2, 3])
;[...s.union(new Set([3, 4]))]        // was [3, 4, 1, 2], native is [1, 2, 3, 4]
;[...s.intersection(new Set([3, 2]))] // argument-ordered instead of receiver-ordered
```

Values and membership are correct, so this slipped past the existing tests, which compare with `toEqual(new Set([...]))` (order-insensitive). It surfaces whenever the result is iterated (spread, `Array.from`, `forEach`, rendering a list) after moving from a plain `Set` to `observable.set`.

## Fix

Build the result from the receiver in every case: `return new Set(this).method(otherSet)`, matching the already-correct `difference`, `isSubsetOf` and `isSupersetOf`. The observable-argument case already used this form (the previous `else` branch), so the `#3919` stack-overflow behavior is unchanged; only the plain-`Set` path changes, from argument-order to receiver-order. `isDisjointFrom` is aligned to the same form as well (its boolean result was already correct, so this is only for consistency).

## Testing

Added an order-sensitive test to `__tests__/base/set.js` that compares against a native `Set` built from the same receiver. It fails before this change; the full set suite (28 tests, including the `#3919` case) stays green with it.

### DIFF
--- a/.changeset/fix-observableset-receiver-order.md
+++ b/.changeset/fix-observableset-receiver-order.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+Fix ObservableSet union, intersection and symmetricDifference to return results in receiver order, matching native Set, when the argument is a plain Set.
--- a/packages/mobx/__tests__/base/set.js
+++ b/packages/mobx/__tests__/base/set.js
@@ -339,6 +339,17 @@ describe("The Set object methods do what they are supposed to do", () => {
         expect(isDisjointFromObservableResult).toBeTruthy()
     })
 
+    test("set methods preserve receiver order (matches native Set)", () => {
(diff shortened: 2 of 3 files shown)


mweststrate:
lgtm, thanks!

PR #4682, opened 2026-07-03
perf: lazily allocate observers_ Set to reduce memory for unobserved atoms

gesposito:
## What / why

Every `Atom` and `ComputedValue` eagerly allocates an empty `Set` for `observers_`, costing ~160 bytes each. In large stores most atoms are never observed (e.g. hydrated data that no reaction reads), so this memory is wasted. This PR allocates the Set lazily on first observer instead (`observers_: Set<IDerivation> | null = null`, allocated via `??=` in `addObserver`). The field initializer is kept so the slot stays in-object and the hidden class remains stable.

`getObservers` returns a shared frozen-by-convention empty Set when unallocated; its only caller (`getObserverTree`) reads it without mutating, so there is no behavioral change.

## Benchmarks

From `npm -w mobx run test:performance` (Node 26, macOS, both sides rebuilt before measuring). Decorator benchmarks, 50k instances with 10 observable fields:

| Scenario | main heap | this PR | change |
|---|---|---|---|
| 10 `@computed` getters | 129.3 MB | 122.0 MB | −6% |
| Hydrated partial (3 of 10 fields assigned) | 152.6 MB | 123.6 MB | −19% |
| Hydrated full (all 10 assigned) | 223.9 MB | 144.2 MB | −36% |

The hydrated-full scenario also improves in time: construct 155→132 ms, re-read 60→28 ms. The proxy and legacy perf suites (37 tests each) pass with timings within run-to-run noise of main, including the "observable with many observers + dispose" case that exercises the lazy allocation path hardest.

## Notes for review

- Three existing tests asserted on the internal `observers_.size` for never-observed atoms; they were updated to `observers_?.size ?? 0` (same intent, tolerates lazy allocation).
- A changeset (patch bump for `mobx`) is included.

### Code change checklist

-   [x] Updated `/docs`. For new functionality, at least `API.md` should be updated — N/A, internal change only, no public API or documented behavior affected
-   [x] Verified that there is no significant performance drop (`npm -w mobx run test:performance`)


mweststrate:
@gesposito merging, thanks for putting this up!

PR #4683, opened 2026-07-03
perf: fast-path primitives in deepEnhancer

gesposito:
## What / why

`deepEnhancer` runs on every value that enters a deep observable — including every scalar write. For a primitive, it currently walks the full chain of type checks (`isObservable`, `Array.isArray`, `isPlainObject`, `isES6Map`, `isES6Set`, `typeof function`) before falling through to `return v`. Primitives can never be made observable, so this PR returns them immediately:

```ts
// primitives can never be made observable; skip the type checks below
if (v === null || (typeof v !== "object" && typeof v !== "function")) {
    return v
}
```

Functions are deliberately excluded from the early return so they still get wrapped into `autoAction`/`flow`. Behavior is unchanged for every input class — primitives previously fell through all checks and returned as-is.

## Benchmarks

From `npm -w mobx run test:performance` (Node 26, macOS), alternating A/B runs against `main` with a rebuild on each side. Consistent across the proxy and legacy suites and across repeated runs:

| Benchmark | main | this PR |
|---|---|---|
| Create array (1M primitives) | 39–44 ms | **10 ms** |
| Set: setting/deleting 10k props × 1000 | ~1750 ms | ~1360 ms (−22%) |
| Map: setting/deleting 100 props × 1000 | 17 ms | 12–13 ms (−25%) |

All other benchmarks are within run-to-run noise.

## Tests

Full suite passes. Added two tests pinning the fast-path boundary in `makereactive.js`:

- every primitive type (number, string, boolean, `null`, `undefined`, symbol, bigint) passes through the deep enhancer unchanged
- a function assigned to a deep observable property after creation is still wrapped into an `autoAction`

A changeset (patch bump for `mobx`) is included.

### Code change checklist

-   [x] Updated `/docs`. For new functionality, at least `API.md` should be updated — N/A, internal change only, no public API or documented behavior affected
-   [x] Verified that there is no significant performance drop (`npm -w mobx run test:performance`)

### DIFF
--- a/.changeset/deep-enhancer-primitive-fast-path.md
+++ b/.changeset/deep-enhancer-primitive-fast-path.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+perf: fast-path primitives in `deepEnhancer`. Writing a primitive into a deep observable no longer runs the observable/array/plain-object/Map/Set/function type checks; primitives can never be made observable, so they are returned immediately. Creating an observable array of primitives is ~4x faster, and observable Set/Map writes are ~20-25% faster in the perf suite.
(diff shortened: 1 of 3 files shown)


kubk:
Thank you! I've confirmed locally the perf is better with this change

PR #4687, opened 2026-07-30
[experimental][unreviewed] Mobx8 cleanup per discussion 

mweststrate:
Per discussion: https://github.com/mobxjs/mobx/pull/4671

# Remove legacy MobX APIs

**Breaking change (major).** This PR removes a set of long-deprecated / rarely-used APIs from `mobx` and cleans up all the internal machinery that existed only to support them. The result is a smaller public surface and a smaller bundle.

### Object API (top-level functions)
`get`, `set`, `has`, `remove`, `keys`, `values`, `entries`, `ownKeys` are gone.

Use the native / instance equivalents instead:

| Removed | Use instead |
|---|---|
| `keys(map)` / `values(map)` / `entries(map)` | `[...map.keys()]` / `[...map.values()]` / `[...map.entries()]` |
| `keys(set)` / `values(set)` | `[...set.keys()]` / `[...set.values()]` |
| `keys(obj)` | `Object.keys(obj)` |
| `get(map, k)` / `set(map, k, v)` / `has(map, k)` | `map.get(k)` / `map.set(k, v)` / `map.has(k)` |
| `has(obj, k)` | `"k" in obj` |
| `remove(obj, k)` | `delete obj.k` |

> The Observable Map/Set/Array **instance methods** (`.get()`, `.set()`, `.has()`, `.keys()`, `array.remove()`, …) are unchanged — only the standalone `mobx.*` functions were removed.

### Interception & observation
- **`intercept`** — removed. There is no replacement for pre-empting/altering changes; restructure so state is validated before it is written.
- **`observe`** — removed. Use **`autorun`** / **`reaction`** to react to changes.
- **`_interceptReads`** — removed (was experimental, undocumented).
- **`spy`** — removed. Use a third-party tool such as [`mobx-log`](https://github.com/kubk/mobx-log) for tracing.

### Introspection
- **`getDebugName`**, **`getDependencyTree`**, **`getObserverTree`** — removed.

### Removed types
All change-event and interception interfaces that these APIs exposed: `IValueWillChange`/`IValueDidChange`, `IArrayWillChange`/`IArrayWillSplice`/`IArrayDidChange`/`IArrayUpdate`/`IArraySplice`, `IMapWillChange`/`IMapDidChange`, `ISetWillChange`/`ISetDidChange`, `IObjectWillChange`/`IObjectDidChange`, and `IInterceptable`/`IInterceptor`/`IListenable`/`IObserverTree`/`IDependencyTree`.

## Other user-facing effects
- **MobX devtools hook removed.** The global `__MOBX_DEVTOOLS_GLOBAL_HOOK__` integration depended on `spy`/`getDebugName` and no longer exists, so the [mobx-devtools](https://github.com/mobxjs/mobx-devtools) browser extension will no longer receive data.
- **mobx-state-tree read-interception (`dehancer`) support removed.** The internal `dehancer` mechanism (and `ObservableValue.raw()`) that backed some MST integrations is gone; those MST versions are no longer supported.
- **Internal global-state version bumped** (`MOBX_GLOBALS_VERSION` 7 → 8) because the shape of MobX's internal global state changed (removed spy listeners). This only matters if multiple MobX versions are loaded in the same process.
- **`mobx-react-lite`**: no longer feeds `getDependencyTree` into React DevTools' `useDebugValue` (internal change; no API impact).

## Not affected
`observable`, `computed`, `action`, `autorun`, `reaction`, `when`, `makeObservable`/`makeAutoObservable`, `runInAction`, `toJS`, `configure`, `onBecomeObserved`/`onBecomeUnobserved`, `defineProperty`, `getAtom`, `isObservable*`, enhancers, flows — all unchanged.

## Migration
Most call sites migrate mechanically to native/instance equivalents (table above) or to `autorun`/`reaction` in place of `observe`. There is no drop-in replacement for `intercept` or `spy`.

--- 

Minimal size currently: 9.37 KiB

Not reviewed yet


andykog:
Is it necessary to remove __MOBX_DEVTOOLS_GLOBAL_HOOK__ or is there alternative API? The extension has 24.1k weekly users according to chrome web store

mweststrate:
@andykog good question. With the removal of `intercept` / `observe` it feels like most of the event system is gone, and my impression is that tools like spy / devtools become a lot less important in the age of AI. 

Since spy is mostly behind __DEV__ flag, it is bundle size quite harmless to keep it (and theoretically LLMs could even leverage it to debug, not sure if they do). Curious if there are other opinions?

andykog:
@mweststrate good point about AI. I'd like to play around with the idea of mobx MCP server that launches Chrome or attaches to the agent's browser via CDP, installs the global hook with Page.addScriptToEvaluateOnNewDocument before the app loads, and streams events back over a WebSocket. The agent could then ask things like "which actions fired since I clicked X and what changed"

gesposito:
Adding a production data point on `observe`/`intercept`, since I don't think there's one in the thread yet.

We run a realtime multiplayer app whose state-sync layer is built on them. A `ChangeTracker` walks each model tree and attaches to every `observable` in it, turning MobX change events into patches that get shipped between client and server. 

Two separable capabilities, and only one of them has an obvious substitute:
- `observe` (did-change): conceptually replaceable with reactions if we accept re-computation instead of change events, though the ergonomics of "what exactly changed, and what was the old value" get much worse.
- `intercept` (will-change): and this is the one I don't see a path off. We use it for arrays and sets specifically because we need the pre-mutation view: for a splice we need added / removedCount / index before the mutation lands to emit a correct patch. `observe` after the fact doesn't carry that. We also use the property-scoped overload `intercept(obj, key, fn)` to deep-freeze opaque (`observable.ref`) fields on write, turning a silent sync failure into a loud `TypeError` at the mutation site.

What would help is knowing the intended direction: is the expectation that `change-event` consumers move to reactions and give up `will-change` entirely, or is there appetite for something narrower surviving (e.g. a `will-change` hook for collections, without the full `intercept`/`observe`/`spy` machinery)?

PR #4689, opened 2026-08-04
fix: useObserver pinned every observer's first render result

gesposito:
### The problem

Every mounted `observer` component permanently retains the element tree returned by its **first** render, along with every fiber and DOM node reachable from it.

`useObserver` creates `adm.subscribe`/`adm.getSnapshot` inside its first invocation (`if (!admRef.current)`), while `renderResult` is declared in that same invocation and assigned inside `adm.reaction.track(() => { renderResult = render() })`. All closures created during the same function invocation share one context object, and variables captured by *any* of them (here: `render` and `renderResult`) live in that shared context. Since `useSyncExternalStore` holds `subscribe` for as long as the component is mounted, the first invocation's context (including the first render's result and the first `render` closure) can never be collected. Subsequent renders allocate fresh contexts that die normally; only the first invocation's survives.

This matters because a React element tree reaches fibers: via `_owner` in development, and in production via props that close over fibers (for example any `useState` setter passed down as a prop). A long-lived observer parent therefore pins the child elements it was first rendered with; after a route change those belong to a stale route, holding its fiber tree and detached DOM.

Found while tracing detached-DOM growth in a production app. Representative heap-snapshot retainer path (weak edges excluded):

```
Reaction --onInvalidate_--> closure --context--> Context
  --adm--> Object --subscribe--> closure      <- held by useSyncExternalStore while mounted
  --context--> Context
  --renderResult--> Object                    <- first render output, still in scope
  --props--> ... --> FiberNode
  --stateNode--> native:<div>                 «DETACHED»
```

### The fix

Create the administration object in a module-level factory (`createObserverAdministration`), so `subscribe`/`getSnapshot` close over the administration object and nothing render-related. No behavior change: the `subscribe`/`getSnapshot` bodies are untouched, they are still created once per component instance, and their identity is still stable across renders.

The file already guarded against a sibling hazard by convention ("Do not store `admRef` (even as part of a closure!)"); this is the same class of problem, reached through the shared invocation context instead. The factory makes both constraints structural, which is also why the three `// Do NOT access admRef here!` comments are removed: `admRef` is no longer in scope where those closures are created.

### The test

`useObserverRetention.test.tsx` mounts an observer, takes a `WeakRef` to the first returned element tree, re-renders a few times (React double-buffers fibers, so the `alternate` legitimately keeps the immediately-previous render alive for one extra commit), forces GC with the same `expose-gc` idiom as `strictAndConcurrentModeUsingFinalizationRegistry.test.tsx`, and asserts the first tree was collected.

Fails on current `main` (the retained element surfaces complete with its `_owner` FiberNode, matching the production heap traces), passes with this change.

There is deliberately no assertion that the first render's `props` become collectable: a control component wrapped in plain `React.memo` (no MobX involved) fails such an assertion too, so a test cannot isolate MobX's contribution. The factory removes both `renderResult` and `render` from `subscribe`'s scope regardless.

### Code change checklist

-   [ ] Updated `/docs`. For new functionality, at least `API.md` should be updated — not applicable: internal memory fix, no API or documented behavior change
-   [x] Verified that there is no significant performance drop (`npm -w mobx run test:performance`) — no `mobx` core changes; `mobx-react-lite` creates the same objects and closures once per component instance as before

### DIFF
--- a/.changeset/use-observer-first-render-retention.md
+++ b/.changeset/use-observer-first-render-retention.md
@@ -1,0 +1,5 @@
+---
+"mobx-react-lite": patch
+---
+
+fix: `useObserver` no longer retains the element tree returned by a component's first render for the component's whole mounted life. `subscribe`/`getSnapshot` were created inside `useObserver`'s first invocation and therefore shared that invocation's closure context with the `reaction.track` callback's captures (`render`, `renderResult`); since React's `useSyncExternalStore` holds `subscribe` while the component is mounted, the first render result (and every fiber and DOM node reachable from it) could never be garbage collected. The administration object is now created by a module-level factory whose scope contains nothing render-related.
--- a/packages/mobx-react-lite/__tests__/useObserverRetention.test.tsx
+++ b/packages/mobx-react-lite/__tests__/useObserverRetention.test.tsx
@@ -1,0 +1,50 @@
+import { render } from "@testing-library/react"
+import * as React from "react"
+import gc from "expose-gc/function"
+import { observer } from "../src"
+
+function nextFrame() {
+    return new Promise(accept => setTimeout(accept, 1))
+}
+
+async function gc_cycle() {
+    await nextFrame()
+    gc()
+    await nextFrame()
+}
+
+// If `subscribe` shares a closure context with the `reaction.track` callback,
+// the first render's result stays reachable for as long as the component is mounted.
+// See the comment on `createObserverAdministration` in src/useObserver.ts.
+test("a mounted observer does not retain the element tree it first returned", async () => {
+    let capturedFirstTree: object | null = null
+
+    const TestComponent = observer(function TestComponent({ label }: { label: string }) {
+        const tree = <div data-label={label} />
+        if (!capturedFirstTree) {
+            capturedFirstTree = tree
+        }
+        return tree
+    })
+
+    const rendering = render(<TestComponent label="first" />)
+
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
The shared-Context explanation was the part I wanted to actually verify rather than take as given, since "closures from the same invocation can pin each other's captures" is easy to state but not always obviously true depending on how much the engine can split scopes. Reimplemented the before/after shape standalone (subscribe created alongside a sibling `render`/`renderResult` in one case, subscribe created via a separate factory call in the other) and ran it under `--expose-gc` with `WeakRef`:

```
BEFORE fix: renderResult still reachable via retained subscribe closure? -> true
AFTER fix:  renderResult still reachable via retained subscribe closure? -> false
```

So the core claim is real on V8, not just plausible-sounding: a closure that never references a sibling variable can still keep it alive as long as both were created in the same function invocation and something holds onto the closure. Moving the object literal into a module-level factory removes the shared invocation entirely, which is the right fix rather than trying to null things out after the fact.

The retention chain in the writeup (`_owner` in dev, props closing over fiber-capturing setters in prod) matches how this would actually surface as detached DOM in a real app rather than just a synthetic test concern, and removing the `admRef` comments makes sense since the factory makes that constraint structural instead of relying on convention. Good test too, the WeakRef-plus-several-rerenders approach correctly accounts for the alternate-fiber retention window instead of asserting collection after just one re-render.

mweststrate:
This is a great fix, thanks for noticing and addressing!

PR #4694, opened 2026-08-19
Fix onBecomeObserved not firing when a computed becomes observed while serving a cached value

### LINKED ISSUE #4547: onBecomeObserved is not called
**Intended outcome:**

I want to watch for observation of some property and do network requests when some other property allows it.

**Actual outcome:**

`onBecomeObserved` is not called, but I clearly accessing property in computed property.

Code:

```ts
const events: string[] = [];

const item = makeAutoObservable({
  canLoad: false,
  value: undefined as number | undefined,
  error: undefined as Error | undefined,

  retry: () => {
    item.error = undefined;
    item.canLoad = true;
  },
});

let disposeReaction = noop;
onBecomeObserved(item, 'value', () => {
  events.push('valueBO');

  disposeReaction = reaction(
    () => item.canLoad,
    isCanLoad => {
      if (!isCanLoad) {
        return;
      }

      runInAction(() => {
        item.canLoad = false;
        item.value = 42;
      });
    },
    { fireImmediately: true },
  );
});
onBecomeUnobserved(item, 'value', () => {
  events.push('valueBU');
  disposeReaction();
});

const parentItem = makeAutoObservable({
  get hasError() {
    return Boolean(item.error);
  },
  get isLoading() {
    return item.value === undefined;
  },
});

const andAnotherItem = makeAutoObservable({
  get see() {
    if (parentItem.hasError) {
      events.push('seenError');
    } else if (parentItem.isLoading) {
      events.push('seenLoading');
    } else {
      events.push(`seenValue ${item.value}`);
    }

    return undefined;
  },
});

autorun(() => {
  void andAnotherItem.see;
});

await sleep(0);
runInAction(() => {
  item.error = new Error();
});

await sleep(0);
runInAction(() => {
  item.retry();
  void parentItem.isLoading;
  when(() => !parentItem.isLoading);
});

await sleep(50);

expect(events).toStrictEqual([
  'valueBO',
  'seenLoading',
  'seenError',
  'valueBU',
  'seenLoading',
  'valueBO',
  'seenValue 42',
]);
```

It gives following error:

```diff
- Expected  - 2
+ Received  + 0

@@ -2,8 +2,6 @@
    "valueBO",
    "seenLoading",
    "seenError",
    "valueBU",
    "seenLoading",
-   "valueBO",
-   "seenValue 42",
```

`seenLoading` indicates, that `autorun` is accessing `parentItem.isLoading`, but `valueBO` is not called.

Then cryptic magic (for me, at least) happens:

- Commenting `void parentItem.isLoading` inside action fixes issue
- Commenting `when(() => !parentItem.isLoading)` inside action fixes issue (but we are leaving `void ...`, so only one of this lines can exist)
- Commenting `disposeReaction()` inside `onBecomeUnobserved` fixes issue

**Versions**

```
"mobx": "6.13.7",
```

Thanks

mrpmohiburrahman:
Fixes #4547.

Unobservation cascades: losing the last observer runs `ComputedValue.suspend_()` → `clearObserving` → `removeObserver` → `queueForUnobservation` down the whole chain. Observation does not. `isBeingObserved` is written in exactly two places, both in `core/observable.ts`, and the observe site marks only the observable handed to `reportObserved`.

That asymmetry is usually invisible, because a newly observed computed is normally suspended (`NOT_TRACKING_`), so `shouldCompute` is true, it recomputes, and the recompute reports its dependencies inside a live `trackingContext` — the cascade happens by accident. The broken path is the one where a computed becomes observed **without** recomputing.

Reduced to 12 lines:

```js
const events = []
const o = observable.box(1)
onBecomeObserved(o, () => events.push("BO"))
const c = computed(() => o.get())

runInAction(() => {
    void c.get()                // non-reactive read inside the batch
    autorun(() => void c.get()) // runs during endBatch(), before unobservation drains
})

events // [] on main, ["BO"] expected
```

Step by step: the non-reactive read inside the batch leaves `c` `UP_TO_DATE_` and merely *queued* for unobservation. `endBatch()` runs `runReactions()` before draining `pendingUnobservations`, so the `autorun` reads `c` with `trackingContext` set and `c.isBeingObserved` flips to true — but `shouldCompute(c)` is now false, so the getter body never runs, `reportObserved(o)` is never called again, and `o.onBO()` never fires. The drain then finds `c` still has observers, so nothing is suspended and the graph stays live permanently with the atom marked unobserved.

Measured end state on `main` for the reporter's own repro:

```
valueAtom.isBeingObserved    : false          <- the lie
valueAtom.observers_         : [isLoading]    <- live edge
isLoading.isBeingObserved    : true
isLoading.dependenciesState_ : 0 (UP_TO_DATE_) <- so shouldCompute() === false
isLoading.observers_         : [see, When@6]
isLoading.observing_         : [value]
```

The atom is transitively observed by a running `autorun`, yet reports `isBeingObserved === false`. This is not a non-reactive read by the time the hook is skipped — the observer chain is live — so I think this is a genuine inconsistency rather than the documented "non-reactive reads don't count" behaviour.

### The fix

`markObserved()` in `core/observable.ts` performs the `isBeingObserved`/`onBO()` transition and recurses through `observing_`, which `IDepTreeNode` already declares. `ComputedValue.get()` snapshots `isBeingObserved` before `reportObserved(this)` and calls it on the `else` (cache-hit) branch only, when the computed just transitioned to observed.

Gated to the cache-hit branch deliberately. Cascading unconditionally from `reportObserved` would also fire for computeds that are about to recompute, and if that recompute drops a dependency the dep gets a spurious `onBO` immediately followed by `onBUO` at `endBatch`. A computed serving a cached value is by definition `UP_TO_DATE_`, so its dependencies cannot be dropped by a pending recompute.

No new instance field (`wasBeingObserved` is a local — keeping the `Atom`/`ComputedValue` bitfields intact), no new public export, and the `trackingContext` guard in `reportObserved` and the `runReactions()`-before-drain order in `endBatch` are both untouched. `markObserved`'s early return makes the recursion cycle-safe and O(newly observed), off the per-read path.

Teardown needed no change: every observable reached through `observing_` already has the computed registered in its `observers_`, so `suspend_` → `clearObserving` → `removeObserver` → `queueForUnobservation` still delivers the matching `onBUO`.

### The tests

Two, both in `__tests__/base/become-observed.ts`, both verified to fail on `main` and pass with this change:

- the 12-line case above, asserting `onBecomeObserved` fires **and** that `onBecomeUnobserved` still fires exactly once when the autorun is disposed, so the fix cannot pass by leaking a permanently-observed atom;
- a two-level chain (`atom → inner → outer`), matching the shape of the reported issue, which is the only coverage of the recursive arm — with a single-level computed `markObserved` never actually recurses.

### Code change checklist

-   [ ] Updated `/docs`. For new functionality, at least `API.md` should be updated — not applicable: bug fix, no API or documented behaviour change
-   [x] Verified that there is no significant performance drop (`npm -w mobx run test:performance`, 37/37 passing, timings within run-to-run noise of `main`). The added cost on the hot path is one bitfield read on a branch that already calls `reportObserved` and `shouldCompute`.

### DIFF
--- a/.changeset/cascade-onbecomeobserved.md
+++ b/.changeset/cascade-onbecomeobserved.md
@@ -1,0 +1,5 @@
+---
+"mobx": patch
+---
+
+fix: `onBecomeObserved` is now called for the dependencies of a computed that becomes observed while serving a cached value. Previously, observation only cascaded when the newly observed computed also happened to recompute, so an observable with a live observer chain up to a running reaction could still report itself as unobserved and never fire its hook.
--- a/packages/mobx/__tests__/base/become-observed.ts
+++ b/packages/mobx/__tests__/base/become-observed.ts
@@ -518,3 +518,48 @@ test("works with ObservableSet #3595", () => {
     expect(onSetObserved).toHaveBeenCalledTimes(1)
     expect(onSetUnobserved).toHaveBeenCalledTimes(1)
 })
+
+test("onBecomeObserved fires when a computed becomes observed while serving a cached value #4547", () => {
+    const events: string[] = []
+
+    const o = observable.box(1)
+    onBecomeObserved(o, () => events.push("BO"))
+    onBecomeUnobserved(o, () => events.push("BUO"))
+    const c = computed(() => o.get())
+
+    let disposeAutorun: () => void
+    runInAction(() => {
+        // non-reactive read inside the batch leaves `c` up-to-date but unobserved
+        void c.get()
+        // `c` becomes observed during endBatch() without recomputing, so it never
+        // re-reports `o` — the hook has to cascade instead
+        disposeAutorun = autorun(() => void c.get())
+    })
+
+    expect(events).toEqual(["BO"])
+
+    disposeAutorun!()
+    expect(events).toEqual(["BO", "BUO"])
+})
+
+test("onBecomeObserved cascades through a chain of cached computeds #4547", () => {
+    const events: string[] = []
+
+    const o = observable.box(1)
+    onBecomeObserved(o, () => events.push("BO"))
+    onBecomeUnobserved(o, () => events.push("BUO"))
+    // two levels, as in the reported issue: the cascade has to recurse
+    const inner = computed(() => o.get())
+    const outer = computed(() => inner.get())
+
+    let disposeAutorun: () => void
+    runInAction(() => {
+        void outer.get()
+        disposeAutorun = autorun(() => void outer.get())
+    })
+
+    expect(events).toEqual(["BO"])
+
+    disposeAutorun!()
+    expect(events).toEqual(["BO", "BUO"])
+})
--- a/packages/mobx/src/core/computedvalue.ts
+++ b/packages/mobx/src/core/computedvalue.ts
@@ -14,6 +14,7 @@ import {
     globalState,
     isCaughtException,
     isSpyEnabled,
+    markObserved,
     propagateChangeConfirmed,
     propagateMaybeChanged,
     reportObserved,
@@ -210,6 +211,7 @@ export class ComputedValue<T> implements IObservable, IComputedValue<T>, IDeriva
                 endBatch()
             }
         } else {
+            const wasBeingObserved = this.isBeingObserved
             reportObserved(this)
             if (shouldCompute(this)) {
                 let prevTrackingContext = globalState.trackingContext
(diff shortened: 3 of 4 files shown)


mweststrate:
Very subtle but nice catch! LGTM

Issue #4692, opened 2026-08-17
Generic ESM bundles evaluate `process.env.NODE_ENV` per `__DEV__` check: ~10× write overhead for Node ESM consumers


gesposito:
**Intended outcome:**

Importing mobx in Node (ESM) or running it through any tool that executes the shipped ESM files as-is (vitest, SSR, backend services, Metro dev builds) should perform roughly like the CJS builds: dev checks should cost what a boolean read costs.

**Actual outcome:**

The two env-agnostic bundles, `dist/mobx.esm.js` and `dist/mobx.mjs`, ship every `__DEV__` check as a live `process.env.NODE_ENV !== "production"` expression: **147 occurrences each** on 7.0.0 (135 on 6.15.0). `process.env` is an exotic object in Node; every property read goes through a host accessor that performs a real environment lookup:

|                                        | ns/op  |
| -------------------------------------- | ------ |
| `process.env.NODE_ENV !== "production"` | ~120   |
| hoisted `const`                         | ~0.9   |

These checks sit on hot paths: with change tracking active, one observable field write performs 7 of them, one array push+pop 19. End to end (Node 26, macOS arm64):

- **Dev mode, 2M observed field writes through `mobx.mjs`: 3324 ms → 319 ms (~10×)** with an otherwise identical build whose `__DEV__` is evaluated once at module scope.
- **Production mode is affected too**: `NODE_ENV=production node` still loads `mobx.mjs` (the `import` condition), which is unminified and evaluates every check per call, creating 1M boxed observables takes 258 ms vs 27 ms through `mobx.cjs.production.min.js` (~9.5×).
- Benchmarks against a real model shape (6.15.0, hand-hoisted build): construct 32,443 → 9,556 ns; scalar write 821 → 195 ns; nested write 1,069 → 189 ns; array push+pop 2,288 → 535 ns. The unmodified minified CJS build measures 192 / 177 / 510 ns on the three mutation benchmarks, so hoisting recovers essentially the whole gap.

Who is affected:

- **Node ESM** consumers (`import` condition → `dist/mobx.mjs`): SSR, backends, scripts, vitest, both dev and prod, since ESM has no equivalent of the runtime `require` switch in `dist/index.js` and Node has no automatic `development`/`production` condition.
[...]

PR #4695, opened 2026-08-19
perf: evaluate NODE_ENV once at module scope in env-agnostic esm builds

gesposito:
Implements option 1 from #4692, which got the go-ahead there.

The env-agnostic bundles (`dist/<pkg>.esm.js` and `dist/<pkg>.mjs`) ship every `__DEV__` check as a live `process.env.NODE_ENV !== "production"` expression (147 of them in mobx). `process.env` is an exotic object in Node, so every check performs a real environment lookup (~120 ns vs ~0.9 ns for a const read), and the checks sit on hot paths: 7 per observed field write. Consumers that execute these files as-is (Node ESM, vitest, SSR, Metro dev builds) pay for it on every operation.

### What changed

For env-agnostic configs only (`env === undefined`), `babel-plugin-dev-expression` is skipped so `__DEV__` survives as a free identifier, `@rollup/plugin-replace` renames it to `__MOBX_DEV__`, and the bundle gets a single `const __MOBX_DEV__ = process.env.NODE_ENV !== "production";` via `output.intro`. The identifier is renamed so consumer-side replacers that blindly define `__DEV__` (e.g. `DefinePlugin({ __DEV__: ... })`) cannot collide with the declaration. Env-set builds go through the exact same pipeline as before.

### Verification

- All env-set artifacts are byte-identical before/after this change: cjs development/production, umd development/production, `dist/index.js`, `mobx.d.ts` (mobx-react and mobx-react-lite share the config).
- In the env-agnostic bundles the only differences are the intro line, the `__DEV__` call sites now reading the const, and one rename inside an already commented-out block coming from `observableobject.ts` (the replace plugin is textual; comment-only, no code effect).
- Live `process.env.NODE_ENV` reads in `mobx.esm.js` / `mobx.mjs`: 147 → 1 (the intro).
- Bundler dead-code elimination still removes dev branches: verified with rollup+terser, webpack 4 (including a deliberate `DefinePlugin({ __DEV__: false })` collision), esbuild and bun `--define`, and metro-transform-plugins in release mode. A production rspack build of an app consuming mobx is byte-identical before/after.
- Node dev mode, 2M observed field writes through `mobx.mjs`: 3371 ms → 326 ms (~10x). Node v26.7.0, macOS arm64.

One semantic note: for unbundled ESM consumers `NODE_ENV` is now snapshotted at import time, which matches what `dist/index.js` already does at require time for CJS consumers.

### Reproducing the byte-identity check

```sh
git checkout main
npm run build -w mobx
cp -r packages/mobx/dist /tmp/mobx-dist-main

gh pr checkout 4695
npm run build -w mobx
diff -rq /tmp/mobx-dist-main packages/mobx/dist
```

Only `mobx.esm.js`, `mobx.mjs` and their two sourcemaps differ; everything else in dist (cjs development/production, umd development/production, `index.js`, the declaration tree) is byte-identical. Same procedure for the react packages with `-w mobx-react` / `-w mobx-react-lite`.

To see that the difference in the esm files is only the intended one, normalize the old expression to the new identifier and diff:

```sh
sed 's/process\.env\.NODE_ENV !== "production"/__MOBX_DEV__/g' /tmp/mobx-dist-main/mobx.mjs | diff - packages/mobx/dist/mobx.mjs
```

which leaves the intro line, `!(x)` vs `!x` parenthesization, and the one rename inside a commented-out block.

### Code change checklist

-   [x] Added/updated unit tests: no test change needed, the artifacts the test suites exercise are byte-identical; the change is covered by the artifact diffs above
-   [x] Updated `/docs`: no API or documented behavior change
-   [x] Verified that there is no significant performance drop (`npm -w mobx run test:performance`): this PR is a measured ~10x improvement for the affected bundles, everything else is byte-identical

### DIFF
--- a/.changeset/hoist-node-env-check-esm-builds.md
+++ b/.changeset/hoist-node-env-check-esm-builds.md
@@ -1,0 +1,7 @@
+---
+"mobx": patch
+"mobx-react": patch
+"mobx-react-lite": patch
+---
+
+perf: evaluate `NODE_ENV` once at module scope in the env-agnostic esm bundles (`dist/<pkg>.esm.js` and `dist/<pkg>.mjs`) instead of at every `__DEV__` call site. `process.env` is an exotic object in Node, so each check performed a real environment lookup on hot paths; consumers that execute these files as-is (Node ESM, vitest, SSR) see roughly 10x faster observable writes in dev mode. All env-set artifacts and bundler output are unchanged.
--- a/scripts/create-rollup-config.mjs
+++ b/scripts/create-rollup-config.mjs
@@ -81,7 +81,16 @@ const stripShebang = () => ({
     }
 })
 
-const babelPlugin = () =>
+// The env-agnostic esm outputs (`dist/<pkg>.esm.js` and `dist/<pkg>.mjs`) keep a
+// runtime `NODE_ENV` check, but evaluate it once at module scope instead of at
+// every `__DEV__` call site: `process.env` is an exotic object in Node, so every
+// read performs a real environment lookup, and unbundled consumers pay that on hot paths.
+// The identifier is scoped (`__MOBX_DEV__`) so consumer-side replacers that blindly define
+// `__DEV__` (e.g. `DefinePlugin({ __DEV__: ... })`) cannot collide with the declaration.
+const DEV_IDENTIFIER = "__MOBX_DEV__"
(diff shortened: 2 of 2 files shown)


mweststrate:
awesome, thanks!

PR #4699, opened 2026-08-19
refactor(eslint-plugin-mobx): rewrite in TypeScript and type flat/legacy configs

rakleed:
Rewrites `eslint-plugin-mobx` in TypeScript so `mobx.flatConfigs.recommended` and `mobx.configs.recommended` are typed when consumed from a TypeScript `eslint.config.ts`, per [@kubk's suggestion](https://github.com/mobxjs/mobx/pull/4699#issuecomment-5353041499):

- Source files renamed to `.ts`.
- Rollup now compiles `src/index.ts` with `rollup-plugin-typescript2` (already used by the other packages in this monorepo). Babel still runs afterwards (via the existing `.babelrc.js`) so the shipped bundle is downleveled the same way it was before this change.
- Declarations are generated into `dist/index.d.ts` and `"types"` points there, so the declaration can no longer drift from the implementation the way a hand-written one could. The generated `export default` is patched to `export =` after the build to match the actual `module.exports = <value>` CommonJS runtime shape regardless of the consumer's `esModuleInterop` setting, and the standalone `.d.ts` `rollup-plugin-typescript2` emits per compiled module is removed since `dist/index.d.ts` is fully self-contained.
- `@types/eslint` moved to `dependencies` (it's referenced by the generated declaration, so it needs to be installed for consumers too), and `@types/estree` is pinned to `^1.0.0` as a direct dependency, since `@types/eslint`'s own dependency on it is unconstrained.

Closes https://github.com/mobxjs/mobx/discussions/4545.

### Code change checklist

- [x] Added/updated unit tests
- [ ] Updated `/docs`. For new functionality, at least `API.md` should be updated
- [x] Verified that there is no significant performance drop (`npm -w mobx run test:performance`)


kubk:
Hi @rakleed Thank you for the PR

the issue I see here is that we need to maintain JSDoc + a handwritten declaration file. Thoughts on rewriting the eslint plugin to TS?

the eslint plugin is already bundled with rollup. Possible plan:

1. Rename the source files to .ts
2. Change the Rollup input to src/index.ts
3. Add rollup-plugin-typescript2, which the monorepo already uses
4. Generate declarations into dist
5. Set "types": "dist/index.d.ts"

kubk:
@rakleed Thank you! could you please use `git mv` so git doesn't re-create files from scratch and we can see what really changed? 

image

Same for `missing-make-observable`

kubk:
I see that some of the Eslint versions still require `module.exports = plugin` and probably because of that there was this workaround in this PR:

telegram-cloud-photo-size-2-5298492083117170528-y 

I've rewritten the eslint plugin build to use plain tsc, so the workaround is no longer needed:

image

kubk:
Thank you for the contribution!

PR #4704, opened 2026-08-26
docs(react): distinguish observer from inline Observer

MuslumYilmaz:
## Documentation gap

The React integration guide documents both the `observer` HOC and the
`<Observer>` render-prop component, but the guidance is separated across the
page. A reader can miss that they establish tracking at different rendering
boundaries.

## Proposed documentation change

Add a compact “Which boundary should be observed?” note near the callback
component example:

- Use `observer(Component)` when the component owns the render that reads the
  observable.
- Use `<Observer>{() => ...}</Observer>` when the observable is read inside a
  callback rendered by a third-party or otherwise non-observer component.
- When composing `observer` with another HOC, keep `observer` innermost so it
  can track reads performed by the wrapped component.

This would connect the existing callback example with the existing HOC
wrap-order tip. It requires no API or runtime changes.

For readers who need a general refresher on render-prop and HOC composition,
this comparison can serve as an optional exercise:
https://frontendatlas.com/react/trivia/react-render-props-vs-hocs

Disclosure: I maintain FrontendAtlas. The reference is optional and can be
removed if external learning links are outside the documentation policy.


kubk:
Thanks. IMHO the existing example is already clear: observer is the default while `<Observer>` is an escape hatch for callbacks. I am also unsure about the link to a separate project, that'd endorse a site we haven't tried

